PR #164, opened 2026-02-25
feat: add startup update check with 24h cache

### LINKED ISSUE #160: Add startup staleness check for outdated binary
## Problem

When a machine is running a stale Homeboy binary, there's no indication unless someone explicitly runs `homeboy upgrade --check`. The upgrade infrastructure is solid (install method detection, crates.io/GitHub API, auto-restart), but it's entirely opt-in — a machine can run months behind without any warning.

This is especially relevant for fleet setups where multiple machines run Homeboy and some may fall behind.

## Proposal

Add a lightweight version check on startup that warns when a newer version is available.

### Behavior

- On any command invocation, check a **local cache file** (e.g. `~/.config/homeboy/version_check.json`) for the last check timestamp
- If more than **24 hours** since last check, fetch the latest version in the background (or inline with a short timeout)
- If an update is available, print a **single-line stderr warning**:
  ```
  hint: Homeboy 0.47.1 is available (current: 0.45.3). Run `homeboy upgrade` to update.
  ```
- Cache the result so subsequent commands within 24h don't make network calls
- **Never block** — if the network call fails or times out, silently skip

### Opt-out

- `HOMEBOY_NO_UPDATE_CHECK=1` env var to disable entirely
- Or a config option: `homeboy config set update_check false`

### Implementation notes

- The `check_for_updates()` function in `src/core/upgrade.rs` already does the heavy lifting — this is mostly adding a cache layer and a startup hook in `main.rs`
- Cache file could store: `{ "latest_version": "0.47.1", "checked_at": "2026-02-25T..." }`
- The check should happen **after** arg parsing but **before** command dispatch, writing to stderr so it doesn't pollute JSON output
- Consider skipping the check for `homeboy upgrade` itself (redundant)

### Prior art

- `rustup` does this
- `npm` does this
- `gh` (GitHub CLI) does this

chubes4:
## Summary

Closes #160. Adds a startup version check that warns users when running a stale Homeboy binary.

- **Every invocation** reads `~/.config/homeboy/update_check.json` — if an update is available, prints a one-line stderr hint
- **Once per 24 hours**, fetches the latest version from crates.io/GitHub Releases and refreshes the cache
- **Skips** for `homeboy upgrade` / `homeboy update` commands (redundant)
- **Silently fails** on network errors, missing config dir, etc. — never blocks or errors

## Escape hatches

- **Env var:** `HOMEBOY_NO_UPDATE_CHECK=1` (CI pipelines, cron jobs, scripts)
- **Config:** `homeboy config set /update_check false` (permanent opt-out)

## Files changed

| File | Change |
|---|---|
| `src/core/update_check.rs` | **New** — cache read/write, freshness check, startup check logic, 5 unit tests |
| `src/core/defaults.rs` | Add `update_check: bool` to `HomeboyConfig` (default `true`) |
| `src/core/mod.rs` | Register `update_check` module |
| `src/main.rs` | Hook `run_startup_check()` after arg parsing |

## How it works

```
homeboy deploy my-site
hint: Homeboy 0.50.0 is available (current: 0.47.1). Run `homeboy upgrade` to update.
{"success": true, "data": { ... }}
```

The hint goes to **stderr** so it never pollutes JSON output on stdout.

## Cache behavior

- Cache stores: `latest_version`, `current_version`, `update_available`, `checked_at` (unix timestamp)
- Cache is invalidated if: older than 24h, or `current_version` doesn't match the running binary (user upgraded manually)
- On fresh cache with `update_available=true`: prints hint immediately, no network call
- On stale/missing cache: calls existing `upgrade::check_for_updates()`, writes cache, prints hint if needed
- Deduplicates hints (won't print twice when refreshing a stale cache that already showed an update)

## Testing

Unit tests cover: cache freshness (within 24h, stale after 24h, stale after version change), serde roundtrip, env var disable logic.

> **Note:** No Rust toolchain on the authoring machine — `cargo test` needs to be run on a machine with Rust installed.

### DIFF
--- a/src/core/defaults.rs
+++ b/src/core/defaults.rs
@@ -5,10 +5,29 @@ use crate::paths;
 use crate::utils::io;
 
 /// Root configuration structure for homeboy.json
-#[derive(Debug, Clone, Serialize, Deserialize, Default)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct HomeboyConfig {
     #[serde(default)]
     pub defaults: Defaults,
+
+    /// Enable automatic update check on startup (default: true).
+    /// Disable with `homeboy config set /update_check false`
+    /// or set HOMEBOY_NO_UPDATE_CHECK=1.
+    #[serde(default = "default_true")]
+    pub update_check: bool,
+}
+
+impl Default for HomeboyConfig {
+    fn default() -> Self {
+        Self {
+            defaults: Defaults::default(),
+            update_check: true,
+        }
+    }
+}
+
+fn default_true() -> bool {
+    true
 }
 
 /// All configurable defaults that can be overridden via homeboy.json
--- a/src/core/mod.rs
+++ b/src/core/mod.rs
@@ -25,6 +25,7 @@ pub mod release;
 
 pub mod server;
 pub mod ssh;
+pub mod update_check;
 pub mod upgrade;
 pub mod version;
 
--- a/src/core/update_check.rs
+++ b/src/core/update_check.rs
@@ -1,0 +1,198 @@
+//! Startup update check — warns users when a newer Homeboy version is available.
+//!
+//! On every command invocation, reads a local cache file. If the cache indicates
+//! an update is available, prints a one-line hint to stderr. If the cache is stale
+//! (older than 24 hours) or missing, fetches the latest version from the network
+//! and refreshes the cache.
+//!
+//! Disable via:
+//! - Environment variable: `HOMEBOY_NO_UPDATE_CHECK=1`
+//! - Config: `homeboy config set /update_check false`
+
+use crate::paths;
+use crate::upgrade;
+use serde::{Deserialize, Serialize};
+use std::time::{SystemTime, UNIX_EPOCH};
+
+const CACHE_FILENAME: &str = "update_check.json";
+const CHECK_INTERVAL_SECS: u64 = 86400; // 24 hours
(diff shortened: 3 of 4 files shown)


chubes4:
## Review: PR #164 — Startup update check

**Conflicts with main:** None (clean merge)

### Issues to address

- **Blocking network call on hot path** — `run_startup_check()` calls `upgrade::check_for_updates()` synchronously when the cache is stale. This blocks the entire command startup for the duration of the network round-trip. Consider spawning a background thread or fire-and-forget approach so the user does not wait.

- **Race condition in env var test** — `test_env_var_disable` mutates process-wide env vars (`set_var`/`remove_var`) without synchronization. If tests run in parallel (Rust default), this can flake. Use a `Mutex` or restructure to pass the env value as a parameter to a testable function.

- **Two file reads before every command** — `is_disabled_by_config()` calls `load_config()` (reads+parses homeboy.json) and `read_cache()` reads the cache file. Both happen before any command runs. Minor perf concern but worth noting.

- **Needs rebase** — main now has changes to `defaults.rs` (removed `post_version_bump_commands` field). The manual `Default` impl this PR adds should still compile after rebase, but verify.

### Looks good

- Cache design is solid (24h TTL, version-change invalidation, hint deduplication)
- stderr output preserves JSON stdout contract
- Skip for upgrade/update commands
- Env var + config escape hatches
- 5 unit tests cover the important paths

PR #446, opened 2026-03-04
fix(test): prevent homeboy flags leaking into test runners

chubes4:
## Summary
- stop `homeboy test` flags like `--analyze`, `--drift`, and `--scaffold` from leaking through to extension test runners
- make passthrough runner args explicit by requiring them after `--`, while preserving passthrough filtering safeguards
- update trailing flag normalization so all modern test flags are treated as known command flags

## Why
`homeboy test data-machine --analyze` was being interpreted partly as passthrough args, so `--analyze` reached PHPUnit and failed with `Unknown option --analyze`. This broke test analysis workflows and made new test command features unreliable in real WordPress components.

## Validation
- `cargo test`
- `homeboy test data-machine --skip-lint --analyze -- --filter='FileAbilities'` now runs and emits `analysis` JSON
- `homeboy test data-machine --skip-lint --drift --since v0.36.0`
- `homeboy test data-machine --scaffold --scaffold-file inc/Api/Chat/Tools/ApiQuery.php`

### DIFF
--- a/src/commands/test.rs
+++ b/src/commands/test.rs
@@ -46,10 +46,6 @@ pub struct TestArgs {
     #[arg(long)]
     analyze: bool,
 
-    /// Auto-fix test drift before running tests (uses generated transform rules)
-    #[arg(long)]
-    auto_fix_drift: bool,
-
     /// Detect test drift — cross-reference production changes with test files
     #[arg(long)]
     drift: bool,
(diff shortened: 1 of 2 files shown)


chubes4:
Added follow-up on this branch:

- 
- 

### New behavior
-  → report drift
-  → generate/apply drift transform fixes (dry-run)
- add  to persist fixes

### Safety hardening
- drift-generated transform rules now use identifier word boundaries for renames
- skip unsafe generic string replacements (prevents broad bad rewrites)

Validated locally with full  and real dry-run execution against data-machine.

Issue #1159, opened 2026-04-18
refactor --from all dry-run reports files_modified for edits that --write silently skips, deadlocking CI auto-refactor on cascading findings


chubes4:
## Summary

`homeboy refactor --from all` exits **1 in dry-run** whenever `sources.files_modified > 0`, but `--write` mode silently **declines to apply** several classes of those edits (returns `applied=false, fixes_applied=0` for the same input). Combined with `homeboy-action`'s `autofix-mode: on-failure` + `skipped-bot-loop` guard, this leaves the `Auto-refactor` CI stage permanently red on PRs whose changed files happen to touch baseline-accepted findings — even when the actual code is fine and there's nothing meaningful to do.

This is the rot that's been forcing `--admin` merges all week. Hit it head-on while clearing the burn-down (#1147 / #1148 / #1150 / #1151 / #1152 / #1157, all merged via `--admin`).

## The contract mismatch

`src/commands/refactor.rs:758`:

```rust
let exit_code = if sources.files_modified > 0 { 1 } else { 0 };
```

`sources.files_modified` is incremented when an edit is **collected and tentatively staged**, not when it's actually **applied to disk**. The same field counts:

1. Real edits a `--write` run will commit to disk → CI should fail and trigger autofix ✅
2. Edits the engine collects, then drops in `--write` mode because they'd cascade-break callers, fail safety guards, or be dominated by another conflicting edit → CI fails but autofix can't help ❌

The two cases need to be distinguished or the gate is unsound.

## Reproduction

Against `main` (`b26e423`):

```bash
# Dry-run reports 2 proposed edits, exits 1
homeboy refactor homeboy --from all --force --changed-since fff82fc
echo \$?  # 1
# … collected_edits: 2  intramethodduplicate × src/core/engine/edit_op_apply.rs

# Same scope with --write applies nothing, exits 0
homeboy refactor homeboy --from all --write --force --changed-since fff82fc
echo \$?  # 0
# … audit: applied=false fixes_applied=0 files_modified=0
git status  # clean
```

The dry-run says \"2 files would change.\" The write says \"nothing changed.\" Both are honest reports of the same engine state, but a CI gate built on the dry-run signal mis-fires on autofix-resistant findings.

[...]

PR #1160, opened 2026-04-19
fix(audit): suppress unreferenced_export false positives on hook callbacks and same-file references (#1149)

### LINKED ISSUE #1149: False positive: unreferenced_export flags WordPress hook callbacks registered via string literal
## Bug

The `unreferenced_export` audit check flags **WordPress hook callbacks and activation/lifecycle functions as dead code** because `is_framework_entry_point()` in `src/core/code_audit/dead_code.rs` doesn't recognize the WordPress hook system.

## Reproduction

Looking at the latest `unreferenced_export` findings filed against Extra-Chill/data-machine#684, here are real functions flagged as "not referenced by any other file":

| Function | Actually called by |
|---|---|
| `datamachine_activate_plugin` | `register_activation_hook(__FILE__, 'datamachine_activate_plugin')` |
| `datamachine_deactivate_plugin` | `register_deactivation_hook(__FILE__, 'datamachine_deactivate_plugin')` |
| `datamachine_register_capabilities` | `register_activation_hook` via `datamachine_activate_plugin`, plus direct calls in upgrade paths |
| `datamachine_activate_plugin_defaults` | `register_activation_hook` fan-out |
| `datamachine_activate_defaults_for_site` | Called from `datamachine_on_new_site` which is wired via `add_action( 'wp_initialize_site', ... )` |
| `datamachine_on_new_site` | `add_action( 'wp_initialize_site', 'datamachine_on_new_site' )` |
| `datamachine_check_requirements` | `add_action( 'admin_init', 'datamachine_check_requirements' )` |
| `datamachine_run_datamachine_plugin` | Called at the bottom of the main plugin file as the bootstrap entrypoint |
| `datamachine_scan_and_instantiate` | Called via `datamachine_load_handlers` / `datamachine_load_step_types` which are hooked |
| `datamachine_allow_json_upload` | `add_filter( 'upload_mimes', 'datamachine_allow_json_upload' )` |

Every one of these is live, critical plugin code. Deleting any of them would break the plugin.

## Root cause

`is_framework_entry_point()` (dead_code.rs:221-292) currently recognizes:

- Universal: `main, new, default, from, try_from, into, drop, clone, fmt, display, eq, hash`
- PHP magic methods: `__construct, __destruct, __get, ...`
- Generic lifecycle names: `register, init, activate, deactivate, boot, setup, render, handle, process`

What it **doesn't** recognize:

1. **Any `datamachine_*` / plugin-prefix function referenced only from `add_action` / `add_filter` / `register_activation_hook` / `register_deactivation_hook` / `register_rest_route` / `register_shortcode` / `register_block_type` / `register_widget` / `register_sidebar` / `register_nav_menu` / `register_post_type` / `register_taxonomy`** — the entire WordPress hook registration vocabulary.

2. **Functions referenced via string literal in these calls** — the reference graph only tracks direct `foo()` call sites, not `add_action( 'hook', 'foo' )` where `'foo'` is a string-literal callback.

3. **Array-callable method references** — `add_action( 'hook', array( 'ClassName', 'method' ) )` or `array( $instance, 'method' )` — these are also missed.

## Impact at scale

Looking at Extra-Chill/data-machine#684: **195 findings**. A rough scan suggests 40–60% of them are false positives caused by this one gap. That alone would drop the findings count below 100.

The signal-to-noise ratio on this issue is so low that the audit report becomes actively harmful — it would lead anyone acting on it to delete live plugin bootstrap code.

## Proposed fix

Two complementary signals:

### 1. Detect hook-callback string literals as references

Extend the fingerprint extraction to pick up callback names from known WordPress hook-registration functions as implicit references. In PHP call-site extraction (wherever `call_sites` are emitted for Php fingerprints), treat these as indirect references to the first string-literal argument (for 2-arg registration) or the second (for `add_action`/`add_filter`):

```
register_activation_hook(__FILE__, 'foo')       → reference to foo
register_deactivation_hook(__FILE__, 'foo')     → reference to foo
register_uninstall_hook(__FILE__, 'foo')        → reference to foo
add_action('hook', 'foo')                       → reference to foo
add_action('hook', 'foo', 10, 2)                → reference to foo
add_filter('hook', 'foo')                       → reference to foo
add_filter('hook', 'foo', 10, 2)                → reference to foo
add_shortcode('tag', 'foo')                     → reference to foo
register_rest_route(..., array('callback' => 'foo'))   → reference to foo
register_rest_route(..., array('permission_callback' => 'foo'))  → reference to foo
register_block_type(..., array('render_callback' => 'foo'))       → reference to foo
```

Array callables and `[$class, 'method']` would require more work and can land in a second pass.

### 2. Widen `is_framework_entry_point()` for prefix-based heuristics

For PHP fingerprints where the project has a clear function prefix convention (`datamachine_`, `wc_`, `bbp_`, `wp_`, etc. — derivable from the most common prefix across public functions in the plugin), treat functions matching that prefix as potential entry points, downgrading `unreferenced_export` from Info to a lower-confidence signal. This catches hook callbacks registered in files that weren't themselves scanned (rare but happens with optional modules loaded conditionally).

Less critical than fix #1 — I'd ship #1 first and see how many findings remain.

## Related

- Similar-shape false positives already tracked:
  - #1134 — `namespace_mismatch` on PHP reserved words
  - #1135 — `missing_import` on same-namespace references
  - #1136 — `unused_function_parameter` on required contract params
- This fits the same pattern and should land the same way (either #1147 picks it up or a follow-up).
- Upstream effect: Extra-Chill/data-machine#684 would shrink dramatically and the rest would become actionable.

## Willing to send a PR

Happy to implement (1) myself — it's localized to the PHP fingerprint extraction + `is_framework_entry_point`, which I've already read through. Let me know if the approach above aligns with the architecture you have in mind, or if there's a preferred mechanism (e.g., reference synthesis at a different phase of the audit pipeline).

chubes4:
Closes #1149

## What

Two complementary structural fixes that together drop the `unreferenced_export` false-positive rate to ~0% on WordPress plugin bootstrap files. Verified against the canonical fixture (`extra-chill/data-machine`'s `data-machine.php`): **17 of 17** public functions go from flagged to clean.

### 1. Recognize same-file hook-callback registrations as references

The check counts a public function as referenced only if some OTHER file calls or imports it. That's right for direct-call languages, but it breaks down when the function is invoked by the framework runtime (WP hooks, REST callbacks, activation hooks, block render callbacks, etc.) **from the same file that defines it** — the standard pattern in plugin bootstrap files.

The `wordpress` extension already extracts these registrations into a `hook_callbacks` set in its fingerprint output, but homeboy core was discarding the field. Now we deserialize and use it.

### 2. PHP/JS/TS: same-file internal calls count as references

For languages that don't have finer visibility than `public` for top-level functions (PHP, JavaScript, TypeScript), a public function called anywhere in its own file IS live code — there's no actionable "narrow visibility" suggestion. The original "skip self" behavior was designed for Rust where `pub(crate)` / `pub(super)` give the suggestion concrete force.

Rust still gets the original behavior. New test `rust_same_file_internal_call_still_flags_unreferenced_export` guards against overcorrection.

## Plumbing

- New `hook_callbacks: Vec<String>` on `FingerprintOutput` (`#[serde(default)]` for backward compatibility)
- Matching field on `FileFingerprint`, propagated through both fingerprint paths (extension script + grammar engine) plus `impact.rs`
- Grammar engine emits empty Vec for now; can populate when grammar-driven PHP fingerprinting matures

## Tests

3 new dead_code tests:
- `php_hook_callback_suppresses_unreferenced_export`
- `php_same_file_internal_call_suppresses_unreferenced_export`
- `rust_same_file_internal_call_still_flags_unreferenced_export`

All 16 dead_code tests pass. Two pre-existing flakes elsewhere (`write_standalone_creates_and_reads_back`, `try_load_cached_audit_reads_output_dir`) are unchanged by this PR — verified pre-existing on main.

## End-to-end verification

Ran the updated fingerprint + suppression logic against `data-machine.php` directly:

| Logic | Functions flagged |
|---|---|
| Old | 17 of 17 public functions |
| New | **0 of 17** public functions |

All 10 functions enumerated in #1149 verified suppressed:

| Function | Suppressed by |
|---|---|
| `datamachine_activate_plugin` | hook + self-call |
| `datamachine_deactivate_plugin` | hook + self-call |
| `datamachine_register_capabilities` | self-call |
| `datamachine_activate_plugin_defaults` | hook + self-call |
| `datamachine_activate_defaults_for_site` | self-call |
| `datamachine_on_new_site` | hook + self-call |
| `datamachine_check_requirements` | self-call |
| `datamachine_run_datamachine_plugin` | hook + self-call |
| `datamachine_scan_and_instantiate` | self-call |
| `datamachine_allow_json_upload` | hook + self-call |

## Companion PR

Extra-Chill/homeboy-extensions#212 extends the wordpress extension's `hook_callbacks` regex set to cover `register_rest_route`, `register_block_type` `render_callback`, `add_shortcode`, `register_uninstall_hook`, and the generic `'callback' / '*_callback'` array key family.

The two PRs compose:

- **This PR alone is safe to land:** no FP regression, just new suppressions on extensions that emit `hook_callbacks`. The wordpress extension already emits that field for the patterns its existing regex set covers.
- **The companion extension PR adds new pattern coverage** — without this core PR it has no effect, with this core PR it widens the set of suppressed FPs even further.

### DIFF
--- a/src/core/code_audit/core_fingerprint.rs
+++ b/src/core/code_audit/core_fingerprint.rs
@@ -439,6 +439,7 @@ pub fn fingerprint_from_grammar(
         internal_calls,
         call_sites: Vec::new(), // Core grammar engine doesn't extract call sites yet
         public_api,
+        hook_callbacks: Vec::new(), // Core grammar engine doesn't extract hook callbacks yet
         trait_impl_methods,
     })
 }
--- a/src/core/code_audit/dead_code.rs
+++ b/src/core/code_audit/dead_code.rs
@@ -115,7 +115,40 @@ pub(crate) fn analyze_dead_code(
         // Skip test files — test methods are invoked by the test runner via
         // reflection/convention, not by direct calls from other source files.
         if !is_test_path(&fp.relative_path) {
+            // For languages without finer-grained visibility than "public" (PHP
+            // top-level functions, for example), a public function called from
+            // anywhere in its own file IS referenced — the "make it private"
+            // suggestion doesn't apply, because PHP top-level functions are
+            // always globally accessible. Rust, by contrast, supports
+            // `pub(crate)` / `pub(super)` narrowing, so a self-only public
+            // function IS actionably dead code.
+            let language_allows_self_reference = matches!(
+                fp.language,
+                super::conventions::Language::Php
+                    | super::conventions::Language::JavaScript
+                    | super::conventions::Language::TypeScript
+            );
+
             for export in &fp.public_api {
+                // Skip if this function is registered as a hook/callback target
+                // from within this same file. Such functions ARE referenced —
+                // just by the framework runtime (WordPress hook system, REST
+                // route callbacks, activation hooks, block render callbacks,
+                // etc.) rather than by direct calls from other source files.
+                // (homeboy#1149 — WP plugin bootstrap files)
+                if fp.hook_callbacks.contains(export) {
+                    continue;
+                }
+
+                // For languages without finer visibility, same-file direct
+                // calls count as references. This catches WP plugin bootstrap
+                // patterns where a top-level helper is called from the main
+                // plugin file (e.g., the top-level WPINC guard that runs
+                // `datamachine_check_requirements()` before continuing).
+                if language_allows_self_reference && fp.internal_calls.contains(export) {
+                    continue;
+                }
+
                 // Check if any OTHER file (owned or reference) references this export.
                 let referenced_elsewhere = owned.iter().chain(reference.iter()).any(|other| {
                     // Skip self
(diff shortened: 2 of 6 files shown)


chubes4:
Pushed `7b046d4` — revert of the drive-by autofix commit `fc21651` that broke CI.

**Diagnosis:** the autofix bot ran 4 refactor fixes against `impact.rs`:
- 3× `pub fn → pub(crate) fn` narrowings on `unreferenced_export` false positives (the exact FP class this PR is trying to fix — circular!)
- 1× `intra_method_duplicate` that collapsed `make_fingerprint(7 args)` to `make_fingerprint(1 arg)`, causing E0061

This is the broken-autofix pattern documented in #1159 — cascading fixes shipping broken code that doesn't compile. Local `cargo check --tests` passes after the revert (one pre-existing dead_code warning on `docs.rs:524`, not introduced here).

Once this PR + Extra-Chill/homeboy-extensions#212 land, those three `unreferenced_export` findings on `fingerprint_from_git_ref` / `diff_changed_files` / `find_affected_files` should disappear on their own — same-file tests call them, and they're reachable through the core audit pipeline.

If the visibility narrowings are genuinely independent improvements, they should come back as a separately reviewed PR rather than as drive-by cascade. Also argues for #1161 merging sooner rather than later to stop the autofix gate from firing on FP findings.

PR #1182, opened 2026-04-22
fix(extension): treat non-git install dirs as clean on update

### LINKED ISSUE #1181: `homeboy extension update` reports "uncommitted changes" for tarball-installed (non-git) extensions
## Summary

`homeboy extension update <id>` (and therefore `homeboy upgrade`, which fans out to extension updates) fails with **"Extension has uncommitted changes; update may overwrite them. Use --force to proceed."** on extensions that were installed as plain directory snapshots (no `.git/`), even when the directory has not been modified.

The directory isn't actually dirty — it just isn't a git repo at all. The clean-check can't tell those two cases apart.

## Environment

- `homeboy 0.89.1` (Homebrew, `extra-chill/tap`)
- macOS (darwin)
- Extension: `wordpress` v2.10.8 (`source_revision: 312fa0a`)
- Installed at `~/.config/homeboy/extensions/wordpress/` — regular directory, **not** a symlink, **not** a git repo (no `.git/`)
- `homeboy extension show wordpress` reports `"linked": false`

## Reproduction

```
$ homeboy extension update wordpress
{
  "success": false,
  "error": {
    "code": "validation.invalid_argument",
    "message": "Invalid argument 'extension_id': Extension has uncommitted changes; update may overwrite them. Use --force to proceed.",
    "details": {
      "field": "extension_id",
      "id": "wordpress",
      "problem": "Extension has uncommitted changes; update may overwrite them. Use --force to proceed."
    }
  }
}
```

Confirming the directory is not a git repo:

```
$ cd ~/.config/homeboy/extensions/wordpress && git status
fatal: not a git repository (or any of the parent directories): .git
```

And has not been modified — it's byte-for-byte the install snapshot, with `.source-revision` pinning `312fa0a`.

## Root cause

In `src/core/extension/lifecycle.rs`:

```rust
fn is_workdir_clean(path: &Path) -> bool {
    let output = Command::new("git")
        .args(["status", "--porcelain"])
        .current_dir(path)
        .output();

    match output {
        Ok(output) => output.status.success() && output.stdout.is_empty(),
        Err(_) => false,
    }
}
```

When the extension dir isn't a git repo, `git status --porcelain` returns a non-zero exit (`fatal: not a git repository`). `output.status.success()` is `false`, so `is_workdir_clean` returns `false` — i.e. "dirty." The update path then bails with the misleading "uncommitted changes" error.

That branch runs for non-linked extensions at `update()` (see `lifecycle.rs:389-396`, current main `bb73f446`). Linked extensions take the `update_linked_extension` path and are checked against their actual source git root, which is correct. The bug is specific to tarball / directory installs.

## Impact

- `homeboy upgrade` breaks for any user whose extensions were installed as plain directory snapshots rather than cloned as git repos. Users see a false "uncommitted changes" warning and assume their extension has local edits they need to recover, when in fact the directory is untouched.
- Workaround today is `--force`, or `homeboy upgrade --skip-extensions` followed by a manual `--force` update. Both are safe in practice but the warning is scary.

## Proposed fix

Have `is_workdir_clean` (or the caller) treat "not a git repo" as **clean**, since there's no working tree to be dirty in the first place. Detect via `git rev-parse --is-inside-work-tree` before running `git status`, or check for `.git` in the extension dir. If the dir isn't a git repo:

- If the extension has a `source_url`, proceed with update (reinstall from source), since there are no local git changes to overwrite — the pristine snapshot is defined by `.source-revision`.
- Keep the current dirty-check behavior only when the dir actually is a git repo.

Happy to send a PR if a direction is preferred. Two reasonable options:

1. Narrow fix inside `is_workdir_clean`: return `true` if `.git` is absent.
2. Split the check: `is_git_repo(path) && !is_workdir_clean(path)` — clearer intent, same result.

## Related code

- `src/core/extension/lifecycle.rs:55` — `is_workdir_clean`
- `src/core/extension/lifecycle.rs:389` — update() bail-out
- `src/core/extension/lifecycle.rs:459` — linked-extension path (not affected)

chubes4:
Fixes #1181.

## Summary

- `is_workdir_clean` now probes with `git rev-parse --is-inside-work-tree` before running `git status --porcelain`. Non-git directories short-circuit to **clean**, since there is no working tree that can be dirty.
- Dirty detection for real git working trees is unchanged: clean repos return `true`, repos with uncommitted/untracked changes return `false`.
- Adds three unit tests covering the non-git / clean-repo / dirty-repo cases.

## Why

Before this change, `homeboy extension update <id>` aborted with

> Extension has uncommitted changes; update may overwrite them. Use --force to proceed.

for any extension installed as a plain directory snapshot (tarball install — no `.git/`). On such directories, `git status` exits non-zero with `fatal: not a git repository`, and the previous match arm conflated that with "dirty tree." This broke `homeboy upgrade` for users whose extensions were installed via the tarball / monorepo snapshot path, and made it look like they had uncommitted local edits when the directory was byte-for-byte the install snapshot.

This hit me on a Homebrew-installed `homeboy 0.89.1` with a tarball-installed `wordpress` extension at `~/.config/homeboy/extensions/wordpress` (no `.git/`, just a `.source-revision` file pinning the commit).

## Manual verification

Built and ran a standalone Rust binary mirroring the new logic against four scenarios:

| Scenario | Before fix | After fix |
| --- | --- | --- |
| non-git dir with a file | `false` (buggy) | **`true`** ✓ |
| fresh empty git repo | `true` | `true` ✓ |
| git repo with untracked file | `false` | `false` ✓ |
| real `~/.config/homeboy/extensions/wordpress` install | `false` (buggy) | **`true`** ✓ |

## Tests

Added a `#[cfg(test)] mod tests` block at the end of `src/core/extension/lifecycle.rs` with:

- `is_workdir_clean_non_git_dir_returns_true` — regression test for #1181
- `is_workdir_clean_clean_git_repo_returns_true`
- `is_workdir_clean_dirty_git_repo_returns_false`

Uses the existing `tempfile` dev-dependency (already in `Cargo.toml`).

## Heads-up on CI

`cargo build` currently fails on a fresh `main` checkout (independent of this PR) with:

```
error[E0599]: no method named `test_mapping` found for struct `ExtensionManifest`
error[E0599]: no method named `test_script` found for struct `ExtensionManifest`
  --> src/core/extension/mod.rs:108, 200, 261
```

`test_mapping` exists as a **field** on `ExtensionManifest` (not a method), and there is no `test_script` method — only `lint_script`, `build_script`, `fingerprint_script`, `refactor_script`, etc. The last several CI runs on `main` are red, so this pre-existing breakage will surface on this PR's CI too. It is unrelated to the fix here and should be addressed in its own PR.

## Post-merge follow-up

Users on `< 0.89.2` (once released) should be able to run `homeboy upgrade` cleanly without the false "uncommitted changes" warning. No user action required to clear the false positive — the next `extension update` pass just works.

### DIFF
--- a/src/core/extension/lifecycle.rs
+++ b/src/core/extension/lifecycle.rs
@@ -51,15 +51,41 @@ pub fn is_git_url(source: &str) -> bool {
         || source.ends_with(".git")
 }
 
-/// Check if a git working directory is clean (no uncommitted changes).
+/// Check if an extension directory is safe to overwrite on update.
+///
+/// Treats the directory as "clean" when:
+/// - It is not a git working tree (tarball / plain-directory installs have no
+///   `.git`, so there is no working tree to be dirty in the first place), or
+/// - It is a git working tree with no uncommitted changes.
+///
+/// Returns `false` only when the directory is a git working tree **and** has
+/// uncommitted changes. This avoids the historical false positive where
+/// `git status` returning a non-zero exit on a non-repo directory was treated
+/// as "dirty" (see Extra-Chill/homeboy#1181).
 fn is_workdir_clean(path: &Path) -> bool {
-    let output = Command::new("git")
+    // If the directory is not a git working tree, there is nothing that can
+    // be "uncommitted". Short-circuit to clean before running `git status`.
+    let inside_tree = Command::new("git")
+        .args(["rev-parse", "--is-inside-work-tree"])
+        .current_dir(path)
+        .output();
+
+    match inside_tree {
+        Ok(output) if output.status.success() => {
+            // Fall through: this is a git working tree; check for changes.
+        }
+        _ => return true,
+    }
+
+    let status = Command::new("git")
         .args(["status", "--porcelain"])
         .current_dir(path)
         .output();
 
-    match output {
+    match status {
         Ok(output) => output.status.success() && output.stdout.is_empty(),
+        // If `status` unexpectedly fails after `rev-parse` succeeded, err on
+        // the side of blocking an overwrite so the user can investigate.
         Err(_) => false,
     }
 }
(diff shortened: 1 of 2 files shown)


chubes4:
Retargeted base to #1183 and rebased on top so CI passes. #1183 is a pure restoration of two accessor methods (`test_script()` and `test_mapping()`) that were accidentally removed in 9c13c5 (#1176) and have been breaking `cargo build` on main. Once #1183 merges, GitHub will auto-retarget this PR back to main.

With the rebase applied:

- `cargo test --lib core::extension::lifecycle::tests` — **3 passed, 0 failed** ✓
  - `is_workdir_clean_non_git_dir_returns_true`
  - `is_workdir_clean_clean_git_repo_returns_true`
  - `is_workdir_clean_dirty_git_repo_returns_false`

PR #1295, opened 2026-04-24
fix(validate): invoke extension validate scripts with bash, not sh (#1276)

### LINKED ISSUE #1276: validate-syntax.sh fails under sh(1): uses bash-only process substitution but is invoked with sh
## Summary

`homeboy validate` crashes immediately on macOS when the `wordpress` extension is active, because the bundled `validate-syntax.sh` uses bash-only process substitution (`< <(...)`) but homeboy invokes it with `sh` instead of `bash`. Every `validate_write` gate after `refactor transform --write` currently degrades to "couldn't verify — assume OK" or short-circuits the rollback path, depending on exit code handling.

## Reproduce

On macOS (any recent — this is `/bin/sh` = bash 3.2.57 in sh-compat mode):

```bash
cd /path/to/any/data-machine/checkout
homeboy validate --path .
```

Output:

```json
{
  "success": false,
  "data": {
    "command": "sh '/Users/.../.config/homeboy/extensions/wordpress/scripts/validation/validate-syntax.sh'",
    "files_checked": 1,
    "output": "/Users/.../validate-syntax.sh: line 27: syntax error near unexpected token `<'\n/Users/.../validate-syntax.sh: line 27: `done < <(find \"$ROOT\" \\'",
    "rolled_back": false,
    "success": false
  }
}
```

## Root cause

`~/.config/homeboy/extensions/wordpress/scripts/validation/validate-syntax.sh` has:

```sh
#!/usr/bin/env bash
...
while IFS= read -r -d '' php_file; do
    ...
done < <(find "$ROOT" ...)
```

The `< <(...)` process-substitution form is a bash/zsh extension — **not** POSIX sh. On macOS, `/bin/sh` is bash in sh-compat mode, which explicitly disables process substitution. So invoking the script as `sh validate-syntax.sh` bypasses the shebang and picks POSIX mode, where `< <(...)` is a syntax error.

The command homeboy emits (visible in the JSON `command` field above) is literally `sh '/path/to/validate-syntax.sh'` — that's the bug. The shebang is ignored when `sh` is the explicit interpreter.

## Expected

Either:

1. **Respect the shebang.** Invoke the script directly (`/path/to/validate-syntax.sh`, not `sh /path/to/validate-syntax.sh`), so the `#!/usr/bin/env bash` line does its job. This is the normal pattern.
2. **Invoke with `bash` explicitly.** Change the spawner from `sh` to `bash`. Requires bash to be available, but the script already requires bash.
3. **Rewrite the loop in POSIX.** Replace `done < <(find ... -print0)` with a `find ... | while read` pipe (loses the nul-separator safety for weird filenames, but works in pure sh).

Option 1 or 2 is the right fix. Option 3 is a regression in file-name safety.

## Impact

- Any `refactor transform --write` or `audit --fix --write` run on macOS with the wordpress extension enabled silently skips the syntax-validation gate. If a transform produces a broken-PHP file, homeboy's rollback-on-validation-failure protection never triggers, and the broken file is persisted to disk.
- Hit in practice today during an architectural cleanup pass on Extra-Chill/data-machine: `refactor transform drop_wp_ability_dead_guards --write` emitted `\\\\WP_Abilities_Registry` (double-backslash) instead of `\\WP_Abilities_Registry` in 4 files, and validate-syntax should have caught it but never ran. (Separate issue: the double-backslash template-escape bug itself — will file that as its own issue with a reproducer.)
- Probably affects every Linux distro too (where `/bin/sh` = dash, also no process substitution) — not macOS-specific.

## Environment

- `homeboy 0.89.1+e8dab371`
- macOS 25, arm64
- `/bin/sh` = GNU bash 3.2.57(1)-release (sh-compat mode)
- Wordpress extension active via `~/.config/homeboy/extensions/wordpress/`

## AI assistance

- **AI assistance:** Yes
- **Tool(s):** Claude Code (Opus 4.7)
- **Used for:** Hit during a data-machine architectural cleanup session; diagnosed the shebang-vs-interpreter mismatch, drafted the issue body. Chris reviewed and requested filing.

SAY-5:
## Summary

`resolve_validate_command` built the validation command as `sh <script>`, and `validate_write` then wrapped that in another `sh -c`, so bundled scripts like `validate-syntax.sh` were ultimately executed by `/bin/sh`. On macOS `/bin/sh` is bash in sh-compat mode, which disables bash-only constructs such as `< <(...)` process substitution. The scripts already declare `#!/usr/bin/env bash`, but that shebang is ignored when an explicit interpreter is passed.

The failure reported in #1276:

```
/Users/.../validate-syntax.sh: line 27: syntax error near unexpected token `<`
/Users/.../validate-syntax.sh: line 27: `done < <(find "$ROOT" \'
```

## Fix

Swap the prefix from `sh` to `bash`. All bundled extension scripts require bash anyway; this keeps the call via `Command::new("sh").args(["-c", ...])` shape (so quoting semantics are unchanged) while honouring the bash contract the scripts actually depend on.

Fixes #1276

## Test plan

- [x] `cargo build` — clean
- [x] `cargo test --lib engine::validate_write` — all 7 existing tests pass (only the one-line format! change is in scope; no existing test asserts the exact "sh" prefix, and adding an extension-fixture test would require a full installed extension, which is out of scope for a trivial interpreter swap)

### DIFF
--- a/src/core/engine/validate_write.rs
+++ b/src/core/engine/validate_write.rs
@@ -243,9 +243,15 @@ fn resolve_validate_command(root: &Path, changed_files: &[PathBuf]) -> Option<St
             let script_path = std::path::Path::new(ext_path).join(script_rel);
 
             if script_path.exists() {
-                // Build the command: pass root and changed files as JSON on stdin
+                // Invoke via `bash` rather than `sh`: the bundled
+                // extension scripts (e.g. validate-syntax.sh) declare
+                // `#!/usr/bin/env bash` and use bash-only constructs
+                // like `< <(...)` process substitution. On macOS
+                // /bin/sh is bash in sh-compat mode, which disables
+                // those constructs and fails with a syntax error.
+                // See https://github.com/Extra-Chill/homeboy/issues/1276.
                 return Some(format!(
-                    "sh {}",
+                    "bash {}",
                     crate::engine::shell::quote_path(&script_path.to_string_lossy())
                 ));
             }


chubes4:
Thanks for the PR, @SAY-5! You beat me to a fix by about 2 minutes 😅 — I'd filed the issue and was drafting a fix in parallel, landed #1296 before seeing yours.

Your diagnosis is spot-on and the `sh → bash` swap would have worked. I went with the alternative approach from the issue body (option 1 — invoke the script directly and let the `#!/usr/bin/env bash` shebang do its job) to avoid assuming `bash` is on PATH. Same net result for the reported failure.

Closing as superseded by #1296. Appreciate you taking a look!

Issue #1492, opened 2026-04-24
audit: share codebase scan, fingerprints, and symbol indexes across audit/fixability/refactor


chubes4:
## Problem

`homeboy audit --changed-since ... --json-summary` can finish the audit scan and baseline comparison, then spend a long time in post-processing. The observed clue was the command printing:

```text
[audit] Drift reduced: 912 finding(s) resolved since baseline
```

and then continuing for minutes before output completed.

The architecture is doing repeated whole-repo work rather than sharing primitive analysis state.

## Current flow

```text
audit --changed-since
  -> audit_internal()
     -> discovery::auto_discover_groups()
        -> code_audit::walker::walk_source_files()
           -> engine::codebase_scan::walk_files()
        -> fingerprint_file() for each source file
     -> detectors use all_fingerprints
     -> baseline::compare()
  -> report::compute_fixability()
     -> generate_audit_fixes()
        -> ModuleSurfaceIndex::build(root)
           -> walk_source_files() again
           -> fingerprint_file() again for each source file
           -> for each public symbol:
              -> symbol_graph::trace_symbol_callers()
                 -> engine::codebase_scan::walk_files() again
                 -> read all candidate files again
                 -> parse imports again
```

The worst offender is `trace_symbol_callers()` being per-symbol while each call walks/reads the codebase. That can approach `files * exported_symbols` scans.

## Existing engine primitives involved

- `engine::codebase_scan` already owns generic walking/search.
- `engine::symbol_graph` already owns import/caller tracing.
- `engine::baseline` is already used correctly by audit baseline comparisons.
- `engine::edit_op` is already used by autofix application.

The issue is not total bypass of `engine`; it is uncoordinated repeated use of its primitives.

## Proposed direction

Introduce a shared analysis context built once per command/run and passed to audit/fixability/refactor consumers.

```text
CodebaseSnapshot
  files: [path, content, metadata]

FingerprintIndex
  file -> FileFingerprint

SymbolIndex
  symbol -> callers
  symbol -> importers
  file -> imports
  file -> calls

AuditContext / AnalysisContext
  snapshot
  fingerprints
  symbol_index
  changed_scope
  reference_fingerprints
```

Then migrate consumers:

```text
audit detectors      -> borrow AnalysisContext
compute_fixability   -> borrow same AnalysisContext
refactor planning    -> borrow same AnalysisContext
```

## Concrete implementation slices

[...]

PR #1515, opened 2026-04-25
feat(audit): CodebaseSnapshot + FingerprintIndex primitives (slice 1 of #1492)

chubes4:
## Summary

- Slice 1 of #1492. Pure addition — `CodebaseSnapshot` owns one walk + one read pass; `FingerprintIndex` builds once from a snapshot.
- New `fingerprint_content(path, root, content)` extracts the disk-free path from `fingerprint_file()`, which now delegates to it. Behavior is preserved byte-for-byte.
- Zero consumer migration in this PR. Audit detectors, fixability, refactor — all untouched. Subsequent slices will swap them onto the shared state and ratchet bench numbers will land then.

## What this is / what it isn't

This is **slice 1 of 7** from #1492. It's the in-memory shared-state scaffolding waiting for consumers. There is no behavior change in this PR — `homeboy audit homeboy --changed-since=v0.93.0 --json-summary` produces identical findings counts (9 / 3 warnings / 6 info / drift +1) before and after this change. The single line-level diff observed across runs is pre-existing nondeterminism in the `repeated_field_pattern` description string (HashMap iteration order), reproducible on `main` without these changes.

## API surface

`src/core/engine/codebase_scan.rs`:

```rust
pub struct CodebaseSnapshot { /* … */ }
impl CodebaseSnapshot {
    pub fn build(root: &Path, config: &ScanConfig) -> Self;
    pub fn root(&self) -> &Path;
    pub fn len(&self) -> usize;
    pub fn is_empty(&self) -> bool;
    pub fn iter(&self) -> impl Iterator<Item = (&Path, &str)>;
}
```

`src/core/code_audit/fingerprint.rs`:

```rust
pub fn fingerprint_content(path: &Path, root: &Path, content: &str) -> Option<FileFingerprint>;
// fingerprint_file() now delegates to fingerprint_content() after a single disk read.

pub struct FingerprintIndex { /* … */ }
impl FingerprintIndex {
    pub fn from_snapshot(snapshot: &CodebaseSnapshot) -> Self;
    pub fn get(&self, path: &Path) -> Option<&FileFingerprint>;
    pub fn len(&self) -> usize;
    pub fn is_empty(&self) -> bool;
    pub fn iter(&self) -> impl Iterator<Item = (&Path, &FileFingerprint)>;
}
```

No shims. `fingerprint_file` is a one-line convenience wrapper around `fingerprint_content`, not a deprecated path. Callers that already have content in hand (snapshot consumers) skip the disk read; callers that don't keep working unchanged.

## Tests

Six new unit tests, all green; full lib suite goes from 1403 → 1409 passing serial.

`codebase_scan` tests:
- `snapshot_build_matches_walk_files` — set-equivalence vs `walk_files` on a fixture with skip dirs, nested dirs, and an extension filter.
- `snapshot_reads_each_file_exactly_once` — every fixture file's content appears exactly once in the snapshot.
- `snapshot_empty_for_empty_dir` — empty / `is_empty` / iter count.

`fingerprint` tests:
- `fingerprint_content_matches_fingerprint_file` — same input via both code paths produces identical structural fields (compared as sorted sets where HashMap iteration is involved).
- `fingerprint_index_matches_per_file_calls` — every snapshot path resolves to the same fingerprint via `index.get(path)` and `fingerprint_file(path, root)`.
- `fingerprint_index_empty_for_empty_snapshot` — empty / `is_empty` / `get` miss.

## Validation

```
cargo build                                                    # clean
cargo test --lib -- --test-threads=1                           # 1409 passed; 0 failed
cargo test codebase_scan                                       # 19 passed (3 new)
cargo test fingerprint                                         # 40 passed (3 new)
homeboy audit homeboy --changed-since=v0.93.0 --json-summary   # parity vs main
```

## Out of scope

Tracked under #1492:
- Slice 2: migrate audit detectors to borrow from `AnalysisContext` (the wrapper struct lands when there's a consumer).
- Slice 3: batched `SymbolIndex` builder.
- Slice 4: `ModuleSurfaceIndex::from_fingerprints_and_symbols(...)`.
- Slice 5: `generate_audit_fixes_with_context(...)`.
- Slice 6: wire `report::compute_fixability()` to the shared context.
- Slice 7: ensure changed-scope audit keeps exact functionality with fewer scans.
- Bench baseline capture (lands when slice 2 has something measurable).

## AI assistance

- **AI assistance:** Yes
- **Used for:** drafted the slice-1 primitives + tests; Chris reviewed scope and approved the cook before I started.

### DIFF
--- a/src/core/code_audit/fingerprint.rs
+++ b/src/core/code_audit/fingerprint.rs
@@ -1,9 +1,10 @@
 //! fingerprint — extracted from conventions.rs.
 
 use std::collections::HashMap;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 
 use super::conventions::Language;
+use crate::core::engine::codebase_scan::CodebaseSnapshot;
 
 /// A structural fingerprint extracted from a single source file.
 #[derive(Debug, Clone, Default)]
@@ -83,14 +84,28 @@ pub struct FileFingerprint {
     pub trait_impl_methods: Vec<String>,
 }
 
-/// Extract a structural fingerprint from a source file.
+/// Extract a structural fingerprint from a source file on disk.
+///
+/// Reads `path` from disk, then delegates to [`fingerprint_content`]. Use
+/// `fingerprint_content` directly when content has already been loaded
+/// (e.g., from a [`CodebaseSnapshot`]) to avoid double-reads.
+pub fn fingerprint_file(path: &Path, root: &Path) -> Option<FileFingerprint> {
+    let content = std::fs::read_to_string(path).ok()?;
+    fingerprint_content(path, root, &content)
+}
+
+/// Extract a structural fingerprint from already-loaded file content.
 ///
 /// Tries the grammar-driven core engine first (no subprocess, faster, testable).
 /// Falls back to the extension fingerprint script if no grammar is available
 /// or the core engine can't handle the file.
-pub fn fingerprint_file(path: &Path, root: &Path) -> Option<FileFingerprint> {
+///
+/// This is the content-taking primitive used by [`FingerprintIndex::from_snapshot`]
+/// to avoid re-reading every file from disk after a [`CodebaseSnapshot`] has
+/// already loaded them. [`fingerprint_file`] is a convenience wrapper that
+/// reads from disk and delegates here so behavior stays identical.
+pub fn fingerprint_content(path: &Path, root: &Path, content: &str) -> Option<FileFingerprint> {
     let ext = path.extension()?.to_str()?;
-    let content = std::fs::read_to_string(path).ok()?;
     let relative_path = path
         .strip_prefix(root)
         .unwrap_or(path)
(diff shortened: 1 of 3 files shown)


chubes4:
## Holding merge — CI audit findings are detector bugs, not code issues

The 8 new findings CI flagged are all false positives from two distinct detector bugs. Filed as standalone issues so they get fixed properly instead of papered over with a baseline refresh:

- **#1517** — `near_duplicate` `count_body_lines()` has an off-by-one (counts signature + body + close as 3 lines, so the `< MIN_BODY_LINES` filter at `< 3` lets one-statement delegation methods through). Compounds because `is_trivial_method` from `test_coverage.rs` already lists `len`/`is_empty`/`iter` but `near_duplicate` doesn't consult it. Fix is structural: lift `is_trivial_method` to a shared helper + fix the line-counting math.

- **#1518** — `missing_test_method` enforces literal `test_<name>` prefix as the only coverage signal, ignoring behavior-describing test names that are the dominant Rust idiom. Fix is a token-bounded substring match (with body-mention as a future enhancement on top of the symbol_graph machinery slice 3 of #1492 will add anyway).

Both are the dogfood loop working: we audited homeboy with homeboy on this PR and the audit caught real bugs in the audit. Fixing those before merging this slice means the slice-2 bench parity gate lands on a determinism-stable, false-positive-free base — which is the whole point of the bench dogfood plan in #1492.

Plan:
1. Cook fixes for #1517 + #1518 in parallel.
2. Land them.
3. Rebase this PR (or just retrigger CI; the source code in this PR doesn't change).
4. Watch CI go green.
5. Merge.

No baseline refresh on this PR. The audit was right that something needed fixing — it just pointed at the wrong place.

PR #1630, opened 2026-04-27
fix(audit): route heuristic findings to review-only issues

chubes4:
## Summary
- Route heuristic audit categories to review-only issue reconciliation by default so noisy threshold/similarity findings stay visible without auto-filing tracker debt.
- Stabilize `repeated_field_pattern` field ordering so JSON descriptions are deterministic across processes.

## Changes
- Adds `IssueGroupRouting` with `actionable` / `review_only` modes and a `ReviewOnlyDefault` reconcile skip reason.
- Defaults `homeboy issues reconcile` audit groups to `review_only` when the category's `FindingConfidence` is heuristic, while preserving explicit `routing` / `issue_routing` overrides.
- Sorts repeated-field groups before rendering descriptions and suggestions to remove HashMap-order nondeterminism.
- Adds regression coverage for routing defaults, routing overrides, review-only reconcile behavior, and deterministic repeated-field description ordering.

## Tests
- `cargo test core::code_audit::field_patterns`
- `cargo test core::issues::reconcile`
- `cargo test commands::issues`
- `cargo test --lib -- --test-threads=1`

Note: one parallel `cargo test --lib` run had three transient failures (`signature_check_majority_wins`, `signature_check_adds_to_existing_outliers`, `test_run_check`); each passed when rerun individually, and the full library suite passed serially.

## Closes
- Closes #1516
- Closes #1621
- Refs #1478
- Refs #1622
- Refs #1492

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** OpenCode (GPT-5.5)
- **Used for:** Implemented the review-only issue routing slice, deterministic repeated-field ordering, and associated tests. Chris remains responsible for review and merge.

PR #1636, opened 2026-04-27
fix(audit): keep review-only findings out of issue filing

### LINKED ISSUE #1478: audit: separate structural findings from heuristic findings — confidence tier on detectors to gate autofix
## The pattern, not a bug

Today's investigation into [#1471](https://github.com/Extra-Chill/homeboy/issues/1471) (the \`orphanedtest\` detector deleting \`ExtensionManifest::test_script\` / \`test_mapping\` for the third time) and [#1476](https://github.com/Extra-Chill/homeboy/issues/1476) (\`homeboy.json\` doesn't arm the autofix verify gate) surfaced a deeper issue the audit framework doesn't currently address: **it treats high-confidence structural findings and low-confidence heuristic findings as the same kind of thing.**

Both get:
- Emitted as \`Finding { severity, kind, ... }\` into the same pipeline.
- Pushed into the same \`refactor --from audit --write\` auto-apply path.
- Committed to user branches by the CI autofix bot without per-finding confidence gating.

That's an architectural flattening. A \`class.notFound\` from PHPStan is a structural fact. \"Method named \`test_*\` is an orphaned test\" is a name-based heuristic. Both at 100% confidence look identical to downstream consumers — but the heuristic can easily be wrong (and has been, on the same code, three times).

## The three-time pattern

The \`test_script\`/\`test_mapping\` case is diagnostic:

1. **#1176** (original breakage, 2026-04-13 era) — \"post-write verify gate for automated refactor\" commit accidentally deleted the methods. Fixed by restoration in PR #1183.
2. **Autofix commit on a later branch** — deleted them again. Fixed by \`a76f51ed\`.
3. **Commits \`cd3f0b38\` and \`f808dd48\`** on the bench PR (#1385) branch — deleted them twice in 7 minutes. Had to force-push twice in this session.

Each fix was **localized** (\"restore these methods\"). No one has changed the data model to make the **class** of bug impossible. The autofix keeps winning because the detector keeps being wrong in the same way, and nothing in the pipeline knows the difference between a finding that should block CI and a finding that should nudge a reviewer.

### Layer 1 — Data model loses information at the boundary

\`core_fingerprint.rs\` knows which functions are tests (internal \`Function\` struct has \`is_test\` and \`has_test_attr\`). It then **flattens** that into a single \`methods: Vec<String>\`, re-attaching a \`test_\` prefix to disambiguate. The detector reverse-engineers test-ness from the prefix, which is ambiguous for legitimately-named source methods. Classic \"encode-then-decode via string convention\" smell — same root cause as the WP stubs narrative-vs-phpstan-return gap from homeboy-extensions#225.

### Layer 2 — Detectors make language-specific semantic claims from language-agnostic structural data

The fingerprint is deliberately language-agnostic (nice for extensibility). The test-coverage detector makes language-specific semantic claims (\"this name-prefixed method is a test\"). That's only safe when the fingerprint guarantees the prefix invariant. For Rust, it doesn't: \`pub fn test_foo()\` outside a test module is a perfectly valid non-test method.

### Layer 3 — Findings don't carry confidence metadata

Every \`Finding\` is the same shape regardless of whether it was derived from:

- **Structural facts** (parser output: \`function X is declared at line N\`, \`file Y imports Z\`). Near-100% reliable.
- **Visibility/graph analysis** (\`this public fn has zero call sites across the workspace\`). High reliability if the graph walker is complete.
- **Name-pattern heuristics** (\`this method starts with \`test_\` so it must be a test\`). Low reliability; easy to wrong-classify.
- **Behavioral guesses** (\`this method hash matches another, so it's a duplicate\`). Depends entirely on hash fidelity.

All four levels currently feed the same auto-apply pipeline. An auto-apply of a structural finding (missing import) is almost always safe. An auto-apply of a heuristic finding (orphaned test) can delete working code.

### Layer 4 — Autofix has too much authority over a detector it doesn't trust enough

The autofix design commits \"fixes\" **before** verifying they compile, and the verify gate (from #1176) is **opt-in per extension**. For a system where detectors can be heuristic and wrong, this is backwards — verify should be default-on, and auto-apply should be gated on finding confidence.

### 1. Add a confidence tier to \`Finding\`

\`\`\`rust
pub enum FindingConfidence {
    /// Derived from parser output or explicit language constructs.
    /// Safe to auto-apply without human review.
    /// Example: unused import, class.notFound from PHPStan.
    Structural,

    /// Derived from graph analysis (call sites, reference resolution).
    /// Safe to auto-apply when the graph walker is provably complete.
    /// Example: unreferenced_export after full cross-file walk.
    Graph,

    /// Derived from name patterns, body hashes, or other heuristics.
    /// NEVER safe to auto-apply. Surfaces as reviewer hint only.
    /// Example: orphanedtest (name-prefix), near-duplicate (hash).
    Heuristic,
}
\`\`\`

Each detector declares its tier. Autofix \`on-failure\` mode only applies \`Structural\` fixes. \`Graph\` fixes require opt-in per detector. \`Heuristic\` findings never auto-apply — they become \`review\` comments.

### 2. Make the verify gate default-on

Move \`autofix_verify\` from per-extension opt-in to per-extension opt-out. For extensions that don't declare a verify command, use a sane default (e.g. \`cargo check --release\` for Rust, \`php -l\` for PHP) or refuse to auto-apply at all.

### 3. Move name-pattern heuristics out of the fingerprinter

The fingerprint contract should be: \"structural facts extracted from the parse tree, never interpretations.\" Test-ness classification (is this function a test?) stays in the detector — and when available, uses structural signals (\`#[cfg(test)] mod tests { #[test] fn ... }\` on Rust, \`extends TestCase\` on PHPUnit) rather than name prefixes.

When structural signals aren't available (PHP without a declared class extends, JS without a formal test framework), heuristic classifiers can kick in **as \`FindingConfidence::Heuristic\` only** — still generate the finding, but flag it as reviewer-hint rather than auto-apply.

### 4. Separate the detector/enforcer/mutator roles

The audit framework currently fuses three roles:

- **Detector** (observes code) — different languages, different accuracy requirements.
- **Enforcer** (blocks CI) — needs findings to be near-100% reliable.
- **Mutator** (applies fixes) — needs findings to be provably correct.

Each role has different correctness requirements. Today they share one pipeline. A clean separation probably looks like:

- Detectors produce findings with confidence tiers.
- Enforcer gates on confidence: \`Structural\` blocks CI; \`Graph\` warns; \`Heuristic\` informs.
- Mutator only consumes \`Structural\` findings by default.

This is the architectural shape question. Doesn't need to land in one PR — probably a wiki article first to settle the design, then incremental PRs that add the confidence field, migrate detectors to declare their tier, and adjust autofix to respect it.

## Scope

This is foundational design work. Likely 1-2 weeks of thinking + incremental code landing, not a single PR. Children would include:

- **Design wiki article** laying out the confidence tiers, the refuge-from-autofix policy for heuristics, and the migration plan.
- **Rust struct change**: add \`confidence\` field to \`Finding\`, default to \`Heuristic\` (safest).
- **Per-detector audit**: classify every detector in \`src/core/code_audit/\` into a tier. Many will be \`Heuristic\` today.
- **Autofix policy change**: gate auto-apply on \`Structural\` tier by default; others become review-only.
- **Verify-gate default**: flip to opt-out with a sensible default per language.
- **Fingerprinter hygiene pass**: move name-pattern heuristics out of the fingerprint data path.

## Relationship to existing issues

Symptoms/children of this broader pattern:

- **#1471** — \`orphanedtest\` detector misclassifying source methods named \`test_*\` as tests (today's specific breakage).
- **#1476** — \`homeboy.json\` needs \`autofix_verify\` — same root pattern, just the config-level manifestation.
- **#1183** (closed) — restore \`test_script\` / \`test_mapping\` accessors — same false positive, first occurrence.
- **homeboy-extensions#225** — WP stubs narrative-vs-machine-readable return types — different domain, same \"structured info flattened into string convention\" smell.

This issue is the **tracker for the pattern**. The band-aid fix for #1471 (add \`test_methods\` field to \`FileFingerprint\`) is still worth shipping as a near-term unblock, but the real work is here.

## Motivating case study

[WordPress/wordpress-develop#11387](https://github.com/WordPress/wordpress-develop/pull/11387) is the reference thesis for \"make homeboy catch what core committers catch before they have to.\" But the autofix loop currently runs BEFORE those catches, with lower-confidence detectors that can delete working code. Until the confidence-tier separation exists, every new detector added is a new way for the bot to break branches.

The WP-core-readiness checklist from today's rig/bench framing depends on this: a reviewer-simulation harness that sometimes deletes methods is worse than one that doesn't exist, because the damage is silent and the developer loses trust in the tool.

## AI assistance

This issue was drafted by Claude Code (Opus 4.7) during a session that traced through \`src/core/code_audit/test_coverage.rs\`, \`core_fingerprint.rs\`, \`fingerprint.rs\`, and the CI workflow to identify the pattern across #1471, #1476, #1183 as instances of a shared architectural gap rather than separate bugs. Chris reviewed the framing and directed the filing of this issue in lieu of a band-aid code fix.

### LINKED ISSUE #1516: audit: repeated_field_pattern description has nondeterministic field order (HashMap iteration)
## Symptom

Two back-to-back invocations of `homeboy audit homeboy --changed-since=<tag> --json-summary` on a clean tree can produce findings whose `description` strings differ by field-name ordering, even when the underlying source is byte-identical. Surfaced during validation of #1515 (slice 1 of #1492):

> The single line-level diff observed across runs is pre-existing nondeterminism in the `repeated_field_pattern` description string (HashMap iteration order), reproducible on `main` without these changes.

This is harmless for the count-level aggregates the audit baseline ratchet uses today, but it's a real problem for:

- **#1492 slice 2 onward**, which wants byte-for-byte findings parity asserted as part of the bench dogfood gate (and after each refactor slice). Nondeterministic descriptions defeat parity assertions.
- **JSON consumers**, including homeboy-action's PR-comment renderer and any CI diff that surfaces `description` strings.

## Root cause

`src/core/code_audit/field_patterns.rs::detect_repeated_field_patterns()`.

```rust
// line 87
let mut field_locations: HashMap<FieldSignature, Vec<(String, String)>> = HashMap::new();

// line 101-105 — iteration order is HashMap-dependent
let repeated_fields: Vec<&FieldSignature> = field_locations
    .iter()
    .filter(|(_, locs)| locs.len() >= MIN_OCCURRENCES)
    .map(|(field, _)| field)
    .collect();

// line 109 — values are Vec<FieldSignature> built by .push() in repeated_fields order
let mut struct_set_to_fields: HashMap<Vec<(String, String)>, Vec<FieldSignature>> =
    HashMap::new();

// line 133 — field_names order is therefore HashMap-iteration-order
let field_names: Vec<&str> = fields.iter().map(|f| f.name.as_str()).collect();

// line 151 — embedded into the description string
description: format!(
    "Repeated field group [{}] appears in {} structs: {}",
    field_names.join(", "),
    locations.len(),
    struct_names.join(", ")
),

// line 166 — final sort is keyed on (file, description), so a permuted
// field_names list shifts which finding wins as the canonical one for a file.
findings.sort_by(|a, b| a.file.cmp(&b.file).then(a.description.cmp(&b.description)));
```

`HashMap` in Rust uses a randomized hasher per process by default. Two processes hash the same keys in different orders. The fix is structural: sort before stringifying.

## Proposed fix

Tiny, mechanical — sort `fields` (and the `repeated_fields` collection it derives from) by a stable key before constructing description strings:

```rust
// after line 121, before iterating struct_set_to_fields:
let mut entries: Vec<(&Vec<(String, String)>, &Vec<FieldSignature>)> =
    struct_set_to_fields.iter().collect();
entries.sort_by(|a, b| a.0.cmp(b.0));

for (locations, fields) in entries {
    // ...
    let mut sorted_fields = fields.clone();
    sorted_fields.sort_by(|a, b| a.name.cmp(&b.name));
    let field_names: Vec<&str> = sorted_fields.iter().map(|f| f.name.as_str()).collect();
    // ...
}
```

Equivalent treatment for `struct_names` (already iterating an already-sorted `locations` Vec — verify, but I think that one is fine because `sorted_locs.sort()` on line 115 stabilizes it).

## Acceptance criteria

- Two back-to-back `homeboy audit homeboy --changed-since=<tag> --json-summary` runs produce byte-identical JSON.
- Existing tests in `field_patterns.rs` still pass; add one regression test that constructs a fixture with multiple repeated fields per group and asserts `description` contains the field names in lexicographic order.
- No baseline churn — existing baselines should still match (the description format string didn't change, only the order of the comma-separated list inside it). If a baseline fingerprint flips because the canonical winner changed for one file, refresh with `homeboy audit homeboy --baseline` and call out the delta in the PR.

## Why now

Slice 2 of #1492 introduces a parity assertion over `homeboy audit` JSON output as the bench dogfood gate. This bug would trip that gate on every run for reasons unrelated to the refactor. Better to fix it as a one-commit prerequisite than to paper over it in the parity check.

## AI assistance

- **AI assistance:** Yes
- **Used for:** traced the nondeterminism through `field_patterns.rs` after the slice 1 cook (#1515) reported it as a pre-existing condition; drafted the fix sketch from code inspection. Chris flagged the issue and asked for it to be filed before continuing to slice 2.

### LINKED ISSUE #1621: audit: make heuristic and threshold categories review-only by default
## Problem

A Data Machine broad-audit triage pass closed ten CI-filed audit issues as not worth acting on in the component:

- Extra-Chill/data-machine#1286 unused parameter
- Extra-Chill/data-machine#1284 shared scaffolding
- Extra-Chill/data-machine#1283 repeated literal shape
- Extra-Chill/data-machine#1281 parallel implementation
- Extra-Chill/data-machine#1274 intra method duplicate
- Extra-Chill/data-machine#1273 high item count
- Extra-Chill/data-machine#1272 god file
- Extra-Chill/data-machine#1271 directory sprawl
- Extra-Chill/data-machine#1279 missing test method
- Extra-Chill/data-machine#1278 missing test file

The findings were not all "false positives" in the narrow sense. Many are useful signals:

- threshold pressure (`god_file`, `high_item_count`, `directory_sprawl`)
- naming/convention test coverage hints (`missing_test_file`, `missing_test_method`)
- broad structural similarity hints (`shared_scaffolding`, `parallel_implementation`, `repeated_literal_shape`)
- callback/interface-shape ambiguity (`unused_parameter`)

But they were poor defaults for automatic per-component GitHub issue filing because the correct outcome was often "do nothing in this component right now." That creates issue churn and makes normal triage look like cleanup failure.

## Proposed direction

This is related to #1478, but focused on default presentation/routing rather than the underlying confidence data model.

Once detectors have confidence/category metadata, Homeboy should treat heuristic and threshold categories differently from high-confidence structural findings by default:

- **Structural / high-confidence defects**: can fail CI and file component issues.
- **Graph findings with known contracts**: can file issues when confidence is high enough.
- **Heuristic findings**: should be review-only unless explicitly opted into issue filing.
- **Threshold findings**: should probably aggregate into a report/dashboard or advisory section, not open recurring component issues by default.
- **Coverage naming findings**: should be opt-in or advisory unless a repo declares that one-source-file ↔ one-test-file naming is an enforced standard.

## Why

The Data Machine pass showed a healthy component-level outcome:

> We inspected representative findings, disagreed with doing churny broad refactors, and closed the component issues.

But that outcome suggests the Homeboy defaults are too eager to convert heuristic/threshold signals into actionable issue debt.

The tool should preserve the signal without forcing every repo to repeatedly close the same classes of non-actionable issues.

## Relationship to #1478

#1478 is the foundational confidence-tier design. This issue is the product/policy follow-up: use those tiers to decide which audit categories create GitHub issues, fail CI, become review hints, or stay in local reports.

## AI assistance

- **AI assistance:** Yes
- **Used for:** Drafting this issue from a live Data Machine audit triage session; Chris challenged the initial framing and directed the distinction between component triage and Homeboy defaults.

chubes4:
## Summary
- Route broad threshold and heuristic audit categories into review-only handling by default.
- Preserve updates and closures for existing audit tracker issues while preventing noisy new/refiled issues.
- Stabilize `repeated_field_pattern` descriptions so field ordering is deterministic across processes.

## Changes
- Adds review-only category defaults and configuration through homeboy.json and CLI flags.
- Threads finding confidence into issue groups and treats heuristic findings as review-only.
- Records skipped review-only categories explicitly in reconcile plans.
- Sorts repeated-field groups before rendering descriptions and suggestions to remove HashMap-order nondeterminism.

## Tests
- `cargo fmt --check`
- `cargo test core::code_audit::field_patterns`
- `cargo test core::issues::reconcile`
- `cargo test commands::issues`
- Previously on this branch: `cargo test issues --lib`
- Previously on this branch: `cargo test --lib -- --test-threads=1`
- Previously on this branch: `cargo test -- --test-threads=1`
- Previously on this branch: `homeboy audit homeboy --path /Users/chubes/Developer/homeboy@audit-confidence-routing`

Closes Extra-Chill/homeboy#1478
Closes Extra-Chill/homeboy#1516
Closes Extra-Chill/homeboy#1621

## AI assistance
- **AI assistance:** Yes
- **Used for:** Implementing the audit issue-routing change, deterministic repeated-field ordering, tests, verification, and PR drafting; Chris reviewed the direction and constraints.

### DIFF
--- a/homeboy.json
+++ b/homeboy.json
@@ -3,8 +3,8 @@
   "baselines": {
     "audit": {
       "context_id": "homeboy",
-      "created_at": "2026-04-27T01:48:54Z",
-      "item_count": 761,
+      "created_at": "2026-04-27T01:56:20Z",
+      "item_count": 760,
       "known_fingerprints": [
         "Commands::src/commands/docs.rs::NamespaceMismatch",
         "Stack (Tests)::tests/core/stack/inspect_test.rs::MissingMethod",
@@ -50,6 +50,7 @@
         "dead_code::src/core/git/github.rs::UnreferencedExport",
         "dead_code::src/core/git/operations.rs::UnreferencedExport",
         "dead_code::src/core/git/operations.rs::UnreferencedExport",
+        "dead_code::src/core/issues/plan.rs::UnreferencedExport",
         "dead_code::src/core/issues/plan.rs::UnreferencedExport",
         "dead_code::src/core/paths.rs::UnreferencedExport",
         "dead_code::src/core/paths.rs::UnreferencedExport",
@@ -241,6 +242,7 @@
         "structural::src/commands/file.rs::HighItemCount",
         "structural::src/commands/fleet.rs::HighItemCount",
         "structural::src/commands/git.rs::GodFile",
+        "structural::src/commands/issues.rs::HighItemCount",
         "structural::src/commands/project.rs::HighItemCount",
         "structural::src/commands/refactor.rs::GodFile",
         "structural::src/commands/refactor.rs::HighItemCount",
--- a/src/commands/issues.rs
+++ b/src/commands/issues.rs
@@ -11,6 +11,7 @@ use std::collections::BTreeMap;
 use std::io::Read;
 use std::path::PathBuf;
 
+use homeboy::code_audit::FindingConfidence;
 use homeboy::issues::{
     apply_plan, reconcile, GithubTracker, IssueGroup, ReconcileConfig, ReconcilePlan,
     ReconcileResult, Tracker,
@@ -80,6 +81,12 @@ enum IssuesCommand {
         #[arg(long, value_name = "LABEL")]
         suppress_label: Vec<String>,
 
+        /// Override review-only categories (repeatable). Replaces the default
+        /// heuristic/threshold list and homeboy.json's
+        /// `issues.review_only_categories` when set.
+        #[arg(long, value_name = "CATEGORY")]
+        review_only_category: Vec<String>,
+
         /// Don't refresh the body of closed-not_planned issues with the
         /// latest finding count. Default is to refresh (so the closed
         /// issue stays useful as a "current state" reference).
@@ -146,6 +153,7 @@ pub fn run(args: IssuesArgs, _global: &super::GlobalArgs) -> CmdResult<IssuesCom
             suppress_from_config,
             suppress_category,
             suppress_label,
+            review_only_category,
             no_refresh_closed,
             list_limit,
             apply,
@@ -163,6 +171,7 @@ pub fn run(args: IssuesArgs, _global: &super::GlobalArgs) -> CmdResult<IssuesCom
                 suppress_from_config,
                 suppress_category,
                 suppress_label,
+                review_only_category,
                 no_refresh_closed,
             )?;
 
@@ -222,6 +231,7 @@ struct GroupRow {
     count: usize,
     label: String,
     body: String,
+    confidence: Option<FindingConfidence>,
 }
 
 impl FindingsInput {
@@ -235,6 +245,7 @@ impl FindingsInput {
                 count: row.count,
                 label: row.label,
                 body: row.body,
+                confidence: row.confidence,
             })
             .collect()
     }
@@ -307,7 +318,7 @@ fn parse_findings_value(value: Value) -> homeboy::Result<FindingsInput> {
             let row_obj = row_value.as_object().ok_or_else(|| {
                 homeboy::Error::validation_invalid_argument(
                     &format!("findings.groups.{}", category),
-                    "Each group must be a JSON object with `count`, optional `label`, optional `body`",
+                    "Each group must be a JSON object with `count`, optional `label`, optional `body`, optional `confidence`",
                     None,
                     None,
                 )
@@ -327,11 +338,32 @@ fn parse_findings_value(value: Value) -> homeboy::Result<FindingsInput> {
                 .and_then(|v| v.as_str())
                 .unwrap_or_default()
                 .to_string();
-            groups.insert(category.clone(), GroupRow { count, label, body });
+            let confidence = row_obj
+                .get("confidence")
+                .and_then(|v| v.as_str())
+                .and_then(parse_confidence);
+            groups.insert(
+                category.clone(),
+                GroupRow {
+                    count,
+                    label,
+                    body,
+                    confidence,
+                },
+            );
         }
     }
 
     Ok(FindingsInput { command, groups })
+}
+
+fn parse_confidence(raw: &str) -> Option<FindingConfidence> {
+    match raw.trim().to_ascii_lowercase().as_str() {
+        "structural" => Some(FindingConfidence::Structural),
+        "graph" => Some(FindingConfidence::Graph),
+        "heuristic" => Some(FindingConfidence::Heuristic),
+        _ => None,
+    }
 }
 
 // ---------------------------------------------------------------------------
@@ -344,18 +376,22 @@ fn build_reconcile_config(
     suppress_from_config: bool,
     cli_categories: Vec<String>,
     cli_labels: Vec<String>,
+    cli_review_only_categories: Vec<String>,
     no_refresh_closed: bool,
 ) -> homeboy::Result<ReconcileConfig> {
     let mut config = ReconcileConfig {
-        suppressed_categories: Vec::new(),
-        suppression_labels: Vec::new(),
         refresh_closed_not_planned: !no_refresh_closed,
+        ..ReconcileConfig::default()
     };
 
     if suppress_from_config {
-        if let Some((suppressed, labels)) = read_suppressions(component_id, path)? {
+        if let Some(reconcile_config) = read_suppressions(component_id, path)? {
+            let (suppressed, labels, review_only) = reconcile_config;
             config.suppressed_categories = suppressed;
             config.suppression_labels = labels;
+            if let Some(review_only) = review_only {
+                config.review_only_categories = review_only;
+            }
         }
     }
 
@@ -365,6 +401,9 @@ fn build_reconcile_config(
     }
     if !cli_labels.is_empty() {
         config.suppression_labels = cli_labels;
+    }
+    if !cli_review_only_categories.is_empty() {
+        config.review_only_categories = cli_review_only_categories;
     }
 
     // Sane default for suppression_labels when neither config nor CLI set
@@ -383,7 +422,7 @@ fn build_reconcile_config(
 fn read_suppressions(
     component_id: &str,
     path: Option<&str>,
-) -> homeboy::Result<Option<(Vec<String>, Vec<String>)>> {
+) -> homeboy::Result<Option<(Vec<String>, Vec<String>, Option<Vec<String>>)>> {
     let component_dir = match path {
         Some(p) => PathBuf::from(p),
         None => match homeboy::component::resolve_effective(Some(component_id), None, None) {
@@ -417,10 +456,19 @@ fn read_suppressions(
         })
         .unwrap_or_default();
 
-    if categories.is_empty() && labels.is_empty() {
+    let review_only = raw
+        .pointer("/issues/review_only_categories")
+        .and_then(|v| v.as_array())
+        .map(|arr| {
+            arr.iter()
+                .filter_map(|x| x.as_str().map(String::from))
+                .collect::<Vec<_>>()
+        });
+
+    if categories.is_empty() && labels.is_empty() && review_only.is_none() {
         Ok(None)
     } else {
-        Ok(Some((categories, labels)))
+        Ok(Some((categories, labels, review_only)))
     }
 }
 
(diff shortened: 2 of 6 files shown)


chubes4:
Stabilization pass result: this is the keeper for the duplicate audit-routing work.

Why this one wins over #1630:
- This preserves convergence for existing tracker state: review-only categories still update existing open issues and close resolved ones; they only skip brand-new filings / re-files.
- The routing policy is configurable via `homeboy.json` / CLI instead of only being embedded into the incoming group shape.
- I folded #1630's unique non-routing fix into this branch as `bb6bf329` (`fix(audit): stabilize repeated field pattern ordering`), so #1516 is now accounted for here too.

Hardcoding check from Chris's extra constraint:
- I checked the #1630/#1636 changed paths for WordPress, WP-CLI, REST, Data Machine, plugin, and project-specific concepts.
- The canonical implementation here is generic audit/reconcile routing plus deterministic ordering. It does not bake WordPress/Data Machine knowledge into Homeboy core.

Validation after folding #1516:
- `cargo fmt --check`
- `cargo test core::code_audit::field_patterns`
- `cargo test core::issues::reconcile`
- `cargo test commands::issues`

Recommendation: keep #1636 open as the canonical PR; close #1630 as superseded.

Issue #1275, opened 2026-04-23
Cross-reference with open GitHub issues


chubes4:
When an audit run finds a new finding that overlaps an already-open homeboy-filed audit issue (by file path + kind + finding signature), **update the existing issue** instead of filing a duplicate. Two benefits:

- Keeps the audit board focused (currently possible to end up with issue #X and issue #Y both naming the same file for different kinds, and also two audit-kind issues for the same kind where one has new findings the other doesn't).
- Lets humans watch a single stable issue per kind per component rather than chasing transient ones.

Stretch: when findings are **resolved** (disappear from a newer run), edit the issue body to strike them through rather than rewriting the whole table, so there's a visible history of what got fixed when. The current `--ratchet` mode does some of this at the baseline level — this issue proposes surfacing it in the issue body visibly.

## Motivation

Concrete signal from the current `Extra-Chill/data-machine` audit board: multiple open audit issues name overlapping file paths, and successive runs can open new issues for the same `(component, kind)` tuple instead of updating the existing one. This makes the board churn-heavy and the history per kind hard to follow.

## Proposed detector

1. Before filing, query open homeboy-filed audit issues on the target repo filtered by `label:audit label:audit:<kind>`.
2. Compute a stable grouping key per issue: `(component, kind)` at minimum, optionally sharded by top-level file path when a single issue would get too large.
3. For each finding in the new run, resolve it to an existing issue by grouping key. If found, update the existing issue body in place (merge the findings table). If not found, file a new issue as today.
4. Track finding signatures (file path + kind + normalized locator) inside the issue body so subsequent runs can diff against them.

## Fixer

Not applicable — this is a reporting/triage behavior change rather than a code-level finding.

Stretch behavior: when a prior finding disappears from a new run, strike it through (`~~…~~`) in the issue body and append a dated "resolved in run YYYY-MM-DD" annotation, preserving history inline instead of rewriting the table.

## Related

- #1270
- #1271
- #1272
- #1273
- #1274

## AI assistance

- **AI assistance:** Yes
- **Tool(s):** Claude Code (Opus 4.7)
- **Used for:** Drafted issue body from manual audit findings on Extra-Chill/data-machine v0.78.0.

chubes4:
Superseded by the issue reconciliation work that landed in #1563, #1687, #1695, and follow-ups, with the remaining reconciliation/core ownership tracked in #1551. Closing this older grouping proposal to keep one current tracker for the remaining problem.

PR #1637, opened 2026-04-27
feat(audit): detect facade passthrough classes

### LINKED ISSUE #1273: Facade-passthrough detector
Flag classes where >N% (e.g. 70%) of public methods are single-expression `return $this->member->method(...args)` delegates to another class. Two common shapes:

- Facade over multiple sub-services (split-then-rejoin anti-pattern).
- Legacy "main" class that's been gutted into thin wrappers over a newer refactored internal class.

## Motivation

Concrete numeric signal from `Extra-Chill/data-machine` v0.78.0:

`inc/Core/Database/Jobs/Jobs.php` is **504 lines** of `return $this->operations->foo()` passthrough to `JobsOperations.php` + `JobsStatus.php`. **16+** call sites do `new Jobs()` and consume the facade API.

## Proposed detector

1. For each class, classify each public method body:
   - Single-expression `return $this->member->method(...args)` delegating to a field.
   - Single-expression `$this->member->method(...args);` with no return (void delegate).
   - Anything else (real implementation).
2. If the ratio of delegate methods to total public methods exceeds a configurable threshold (default 70%), emit a `facade_passthrough` finding.
3. Annotate the finding with the target member(s) being delegated to, the count of passthroughs, and a call-site count for `new Foo()` / static usage of the facade.

## Fixer

Plan-only. The fixer proposes one of:
- Inlining the delegate(s) at every call site when the facade adds no real value.
- Deleting the facade and updating call sites to construct the inner class directly.

The plan lists every call site so a human can decide which option to take.

## Related

- #1270
- #1271
- #1272
- #1274
- #1275

## AI assistance

- **AI assistance:** Yes
- **Tool(s):** Claude Code (Opus 4.7)
- **Used for:** Drafted issue body from manual audit findings on Extra-Chill/data-machine v0.78.0.

chubes4:
## Summary
- Adds an audit detector for PHP facade classes whose public methods mostly pass through to an inner member.
- Reports delegate ratio, target member breakdown, and external constructor call-site count so cleanup can be planned safely.
- Leaves issue grouping to the existing `homeboy issues reconcile` path, which already updates one issue per category instead of filing duplicates.

## Changes
- Adds `facade_passthrough` as a first-class `AuditFinding` kind.
- Runs the detector during the main audit pass after repeated field detection.
- Adds unit coverage for all-delegate facades, mixed real classes, ignored constructors/private methods, call-site counts, and small-class suppression.

## Tests
- `cargo test facade_passthrough`
- `cargo test core::issues::reconcile`

## Closes
- Closes #1273

## AI assistance
- **AI assistance:** Yes
- **Used for:** Implemented the detector, wrote focused tests, verified existing issue reconciliation behavior for #1275, and prepared the PR.

### DIFF
--- a/src/core/code_audit/conventions.rs
+++ b/src/core/code_audit/conventions.rs
@@ -200,6 +200,9 @@ pub enum AuditFinding {
     /// method-shape (same method names + visibilities + order) and have high
     /// per-method body similarity — candidates for a shared base class.
     SharedScaffolding,
+    /// Class whose public methods are mostly single-expression delegates to an
+    /// internal member — usually a split-then-rejoin facade or legacy wrapper.
+    FacadePassthrough,
 }
 
 impl AuditFinding {
@@ -247,6 +250,7 @@ impl AuditFinding {
             "dead_guard",
             "upstream_workaround",
             "shared_scaffolding",
+            "facade_passthrough",
         ]
     }
 }
--- a/src/core/code_audit/facade_passthrough.rs
+++ b/src/core/code_audit/facade_passthrough.rs
@@ -1,0 +1,431 @@
+//! Facade-passthrough detection.
+//!
+//! Flags PHP classes where most public methods are single-statement delegates
+//! to the same inner member, e.g. `return $this->operations->create(...)`.
+
+use std::collections::{HashMap, HashSet};
+
+use regex::Regex;
+
+use super::conventions::{AuditFinding, Language};
+use super::findings::{Finding, Severity};
+use super::fingerprint::FileFingerprint;
+
+const MIN_PUBLIC_METHODS: usize = 3;
+const DELEGATE_RATIO_THRESHOLD: f32 = 0.70;
+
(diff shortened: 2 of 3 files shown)


chubes4:
Hardcoding check from the audit-routing stabilization pass:

This PR does not appear to bake WordPress, WP-CLI, REST, or Data Machine-specific concepts into core. It is, however, a PHP-class-specific detector implemented directly in Homeboy core (`facade_passthrough.rs` parses PHP methods, `$this->member` delegates, and `new ClassName(...)` call sites).

Chris's constraint for this pass: Homeboy core must remain language/framework agnostic. If this detector needs PHP-specific syntax knowledge, it should live behind a generic detector abstraction, language-extension hook, or project/config-driven rule rather than becoming another core detector with embedded PHP semantics.

Recommendation: review the placement before landing. The detector idea may be useful, but the Homeboy-core shape should stay generic.

PR #1638, opened 2026-04-27
fix(audit): source false-positive rules from config

### LINKED ISSUE #1544: audit: unreferenced_export rule misses WP-CLI subcommand methods
## Summary

The \`unreferenced_export\` audit rule flags public methods on WP-CLI command classes as unreferenced even when they're reachable via WP-CLI subcommand invocation. Static call-site analysis can't see WP-CLI's runtime registration → subcommand → method dispatch path.

## Concrete repro

**Project:** [Extra-Chill/data-machine](https://github.com/Extra-Chill/data-machine)
**Issue created from this finding:** Extra-Chill/data-machine#684

**Finding:**
> \`inc/Cli/Commands/EmailCommand.php\` — Public function \`test_connection\` is not referenced by any other file

**Reality:** \`test_connection\` is the \`wp datamachine email test-connection\` subcommand. The class is registered via:

\`\`\`php
// inc/Cli/Bootstrap.php:35
WP_CLI::add_command( 'datamachine email', Commands\EmailCommand::class );
\`\`\`

WP-CLI introspects the class at runtime — every public method becomes a subcommand. The method's docblock declares the routing:

\`\`\`php
/**
 * Test the IMAP connection.
 *
 * ## EXAMPLES
 *
 *     wp datamachine email test-connection
 *
 * @subcommand test-connection
 */
public function test_connection( array \$args, array \$assoc_args ): void {
\`\`\`

There are no PHP call sites for this method (and there shouldn't be — it's invoked by the CLI runtime, not by other PHP code). The auditor's \"public method must be called from another file\" heuristic is wrong here.

## Why this matters

Every WP-CLI command class on a WordPress plugin gets falsely flagged. data-machine has dozens of subcommands across multiple command classes — each one would trip this rule. Repos that try to clean up the finding by removing the \"unused\" method break their CLI surface. Repos that suppress the finding leave the rule mis-firing for the next consumer.

## Proposed fix

Detect WP-CLI command registration and treat every public method on a registered command class as reachable.

**Detection signals (any of):**
1. Class is the second argument to a \`WP_CLI::add_command()\` call anywhere in the codebase.
2. Class extends a known WP-CLI base class (e.g. \`WP_CLI_Command\`, or in this repo, \`DataMachine\\Cli\\BaseCommand\`).
3. Method's docblock contains \`## OPTIONS\` or \`## EXAMPLES\` or \`@subcommand\` (WP-CLI's documented conventions).

The first signal is the most precise. The third is the cheapest if scanning every \`add_command()\` callsite is expensive.

**Less rigorous fallback (acceptable):** if a class extends \`WP_CLI_Command\` (or a project-specific base flagged in \`homeboy.json\`), suppress \`unreferenced_export\` findings on its public methods.

## Detection one-liner for repo audits

\`\`\`bash
grep -rn 'WP_CLI::add_command' --include='*.php' .
\`\`\`

Each match's second argument is a class whose public methods are CLI-reachable.

## Out of scope

Other invoke-by-name registries that may have the same blind spot (REST routes registered via \`register_rest_route()\` callbacks, AS hook handlers, action/filter callbacks). Worth tracking separately — the WP-CLI case is the most common false positive on plugin codebases.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** Claude Sonnet 4.5 (via Kimaki / Claude Code)
- **Used for:** Diagnosed the false positive while triaging audit issues on data-machine, traced the registration site in Bootstrap.php, drafted this upstream report. Chris confirmed the upstream-fix-first rule applies.

### LINKED ISSUE #1545: audit: missing_interface rule misclassifies traits and uses pure pattern-matching
## Summary

The \`missing_interface\` audit rule flags classes that don't implement what siblings in the same directory implement — without checking (a) whether the missing thing is actually an interface or just a trait, and (b) whether the flagged class actually needs the methods the trait/interface provides. The result is **majority-rule pattern matching**, not real conformance analysis.

## Concrete repro

**Project:** [Extra-Chill/data-machine](https://github.com/Extra-Chill/data-machine)
**Issue created from this finding:** Extra-Chill/data-machine#1134

**Finding:**
> \`inc/Abilities/Chat/SendMessageAbility.php\` — Missing interface: ChatSessionHelpers
> Suggestion: Implement ChatSessionHelpers to match the convention in Chat

**Reality (two distinct bugs in the same finding):**

### Bug 1: \`ChatSessionHelpers\` is a trait, not an interface

\`\`\`php
// inc/Abilities/Chat/ChatSessionHelpers.php:20
trait ChatSessionHelpers {
    protected ConversationStoreInterface \$chat_db;
    // ... session DB access + ownership helpers
}
\`\`\`

The audit categorizes this as \"missing interface\" but the symbol is declared with \`trait\`. The rule conflates traits with interfaces purely on use-site shape (\`use Foo;\` looks similar inside a class body whether \`Foo\` is a trait or a class with the same name).

### Bug 2: SendMessageAbility doesn't need session helpers

The trait provides:
- \`initDatabase()\` — wires up \`\$this->chat_db\`
- \`checkPermission()\` — chat capability check
- \`can_access_user_sessions()\` / \`verifySessionOwnership()\` — session ownership

\`SendMessageAbility\` (which actually sends a message via \`ChatOrchestrator::processChat()\`) doesn't access session DB rows, doesn't list sessions, doesn't verify ownership at the session level. The other 4 chat abilities (\`Get/List/Delete/MarkRead/Create ChatSession\`) all manipulate session records and need the trait's helpers — that's why they use it. Adding the trait to \`SendMessageAbility\` would inject unused fields and methods.

The audit conclusion (\"4 of 5 use this thing, the 5th must be wrong\") is majority-rule pattern matching against a directory rather than real \"this class uses an undefined symbol the trait would provide\" analysis.

## Why this matters

Any directory with N+1 sibling classes where N use a shared trait/interface and 1 legitimately doesn't will get false-positive findings. This happens frequently in:
- Ability classes where a subset shares a common dependency (DB access, OAuth, etc.)
- Handler hierarchies where some handlers need session/state and others are stateless
- Test directories where most files use a fixture trait and some are unit-pure

The current rule's \"convention\" heuristic punishes intentional differentiation.

## Proposed fix

The rule needs two distinct corrections:

1. **Distinguish traits from interfaces.** Resolve the symbol kind (\`class\` / \`trait\` / \`interface\`) before reporting — the finding should say \"missing trait usage\" or \"missing interface implementation,\" not conflate them. Mislabeling makes the suggestion (\"implement ChatSessionHelpers\") nonsensical because you can't \`implements\` a trait.

2. **Switch from majority-rule to capability-detection.** The rule should fire only when the flagged class actually uses (or could benefit from) symbols the trait/interface provides. Detection signals:
   - Class references an undefined symbol that the trait/interface defines.
   - Class re-implements logic that's identical to a method on the trait/interface (genuine duplicate-effort case).
   - Class declares a property that overlaps with a property on the trait (potential type confusion).

   Pure \"sibling files use this thing\" should not be a finding — that's a convention preference, not a defect.

## Less invasive interim mitigation

Add a per-rule allowlist in \`homeboy.json\`:
\`\`\`json
{
  \"audit\": {
    \"missing_interface\": {
      \"allowlist\": [
        \"inc/Abilities/Chat/SendMessageAbility.php\"
      ]
    }
  }
}
\`\`\`

Repos can suppress per-file false positives without disabling the whole rule. Doesn't fix the upstream rule but contains the bleeding.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** Claude Sonnet 4.5 (via Kimaki / Claude Code)
- **Used for:** Diagnosed the false positive while triaging audit issues on data-machine, read the trait + the flagged class, identified the two distinct bugs (trait vs interface, majority-rule vs capability-detection), drafted this upstream report.

### LINKED ISSUE #1546: audit: missing_registration rule flags REST utility classes that aren't endpoints
## Summary

The \`missing_registration\` audit rule flags every PHP class in a directory that contains REST endpoints, regardless of whether the flagged class is itself an endpoint registrant. Utility classes (verifiers, value objects, resolvers) consumed by REST controllers get false-positive findings asking them to call \`rest_api_init\`.

## Concrete repro

**Project:** [Extra-Chill/data-machine](https://github.com/Extra-Chill/data-machine)
**Issue created from these findings:** Extra-Chill/data-machine#950

**Findings (5):**
> \`inc/Api/WebhookVerifier.php\` — Missing registration: rest_api_init
> \`inc/Api/WebhookSignatureVerifier.php\` — Missing registration: rest_api_init
> \`inc/Api/WebhookAuthResolver.php\` — Missing registration: rest_api_init
> \`inc/Api/WebhookVerificationResult.php\` — Missing registration: rest_api_init
> \`inc/Api/Flows/FlowScheduling.php\` — Missing registration: rest_api_init
>
> Suggestion: Add rest_api_init call to match the convention in Api

**Reality:** all 5 files are utility classes consumed by REST controllers, not controllers themselves.

\`WebhookVerifier.php\` is the most obvious example — its file docblock literally says:

> Webhook verifier — provider-agnostic template engine.
> Verifies HMAC-family webhook signatures from a declarative config

And the class itself:
\`\`\`php
class WebhookVerifier {
    const DEFAULT_MAX_BODY_BYTES    = 1048576;
    const DEFAULT_TOLERANCE_SECONDS = 300;

    /**
     * Verify a webhook request against a template config.
     *
     * @param string \$raw_body     Raw request body bytes (as signed).
     * @param array  \$headers      Lower-case-keyed headers.
     * ...
     */
\`\`\`

That's a stateless verification engine with constants and a static-shaped \`verify()\` method. It's called from REST endpoints; it isn't one. Adding \`add_action('rest_api_init', ...)\` to it would either be a no-op or actively wrong (the engine has no routes to register).

The same applies to \`WebhookSignatureVerifier\` (signature math), \`WebhookAuthResolver\` (auth resolution), \`WebhookVerificationResult\` (result value object), and \`FlowScheduling\` (scheduling helper). None are REST controllers.

## Why this matters

Any REST-heavy directory will accumulate utility classes alongside the controllers. The current rule penalizes that organization by demanding utility classes pretend to be endpoints. Two failure modes:

1. **Repos that try to fix the finding** add no-op \`rest_api_init\` callbacks to utility classes — pollutes the boot path with hooks that do nothing.
2. **Repos that suppress** leave the rule mis-firing for the next consumer.

Either way, the rule's directory-presence heuristic is too loose.

## Proposed fix

Only flag classes that are plausibly REST controllers. Detection signals (any of):

1. Class extends \`WP_REST_Controller\` (or a project-specific base flagged in \`homeboy.json\`).
2. Class has a method that returns or calls \`register_rest_route()\` (the actual route registration call).
3. Class is registered via a constructor call inside an existing \`rest_api_init\` hook (deeper analysis but most precise).

Conversely, **suppress** classes that look like utility/value types:
- Class has no constructor side effects (no hook registration in \`__construct\`).
- Class is referenced as a type hint by other classes (consumed as a dependency, not registered as a hook target).
- Class is mostly static methods or constants.

## Less invasive interim mitigation

Per-rule allowlist in \`homeboy.json\`:
\`\`\`json
{
  \"audit\": {
    \"missing_registration\": {
      \"allowlist\": [
        \"inc/Api/WebhookVerifier.php\",
        \"inc/Api/WebhookSignatureVerifier.php\",
        \"inc/Api/WebhookAuthResolver.php\",
        \"inc/Api/WebhookVerificationResult.php\",
        \"inc/Api/Flows/FlowScheduling.php\"
      ]
    }
  }
}
\`\`\`

## Detection one-liner

\`\`\`bash
grep -rn 'register_rest_route' --include='*.php' inc/Api/
\`\`\`

Files that match are real endpoint registrants. Files that don't match but live in the same directory are likely utility classes — those should NOT be flagged for missing \`rest_api_init\`.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** Claude Sonnet 4.5 (via Kimaki / Claude Code)
- **Used for:** Diagnosed the false positives while triaging audit issues on data-machine, read each flagged file to confirm utility-class shape, drafted this upstream report.

### LINKED ISSUE #1547: audit: missing_method rule has same convention-pattern-matching gap as missing_registration (related to #1546)
## Summary

Companion to Extra-Chill/homeboy#1546. The \`missing_method\` audit rule has the same shape problem: it uses pattern-matching against sibling files in a directory rather than capability-detection. Most findings are utility classes flagged for missing methods that only true REST controllers / domain-class instances need.

## Concrete repro

**Project:** [Extra-Chill/data-machine](https://github.com/Extra-Chill/data-machine)
**Issue created from these findings:** Extra-Chill/data-machine#719

Selected examples from the 38 findings:

| Flagged file | Missing method | Why it's wrong |
|---|---|---|
| \`inc/Api/WebhookVerifier.php\` | \`register\`, \`register_routes\` | Static engine class. Not a REST controller. (See homeboy#1546.) |
| \`inc/Api/WebhookSignatureVerifier.php\` | \`register\`, \`register_routes\` | Signature math utility. Not a REST controller. |
| \`inc/Api/WebhookAuthResolver.php\` | \`register\`, \`register_routes\` | Auth resolution utility. |
| \`inc/Api/WebhookVerificationResult.php\` | \`register\`, \`register_routes\` | Value object. |
| \`inc/Core/Database/Chat/ConversationStoreFactory.php\` | 11 chat methods (\`update_title\`, \`delete_session\`, \`mark_session_read\`, etc.) | **Factory pattern**. The methods live on the returned store instance, not on the factory. |
| \`inc/Core/Auth/RemoteAgentClient.php\` | \`__construct\` | Has a constructor. Auditor missed it. |
| \`inc/Core/Auth/CallerContext.php\` | \`__construct\` | Has a constructor. |
| \`inc/Engine/AI/Directives/MemoryFilesReader.php\` | \`get_outputs\` | Reader, not a directive. The convention applies only to directive classes. |

Out of 38 findings, the majority fit one of three patterns:
1. **Utility class flagged for REST controller methods** (same root as homeboy#1546)
2. **Factory pattern misread** as needing the methods of the type it produces
3. **Auditor parsing failure** — the method actually exists but the parser missed it (e.g. \`__construct\` on \`RemoteAgentClient\`)

## Proposed fix

Three corrections, in priority order:

1. **Reuse the controller-detection signal from homeboy#1546.** If a class isn't a plausible REST controller, don't flag it for missing controller methods. Same allowlist (extends \`WP_REST_Controller\`, has \`register_rest_route()\` calls).

2. **Distinguish factory classes from concrete types.** Factories typically have:
   - A static \`get()\`, \`for()\`, \`create()\`, or \`make()\` method that returns an instance of the actual type
   - Class name ends in \`Factory\`
   - Method bodies that instantiate and return objects of another type

   Don't flag a factory for missing methods of the type it produces — the convention check should look at the produced type, not the factory.

3. **Fix the \`__construct\` parser miss.** Two files in this batch (\`RemoteAgentClient\`, \`CallerContext\`) clearly have \`__construct\` defined and the rule still flagged them. Possible parser issue — worth a unit test against real source files.

## Less invasive interim mitigation

Per-rule allowlist in \`homeboy.json\`, same shape as proposed for homeboy#1546.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** Claude Sonnet 4.5 (via Kimaki / Claude Code)
- **Used for:** Diagnosed the false positives while triaging audit issues on data-machine, identified the three sub-patterns, drafted this upstream report. Filed alongside homeboy#1546 as a related fix.

### LINKED ISSUE #1548: audit: naming_mismatch rule treats every directory file as needing the directory's primary suffix
## Summary

The \`naming_mismatch\` audit rule flags every PHP file in a directory whose name doesn't end with the directory's most-common suffix. Utility classes, traits, helpers, factories, value objects, and other legitimately-not-the-primary-type files trip this on every codebase.

## Concrete repro

**Project:** [Extra-Chill/data-machine](https://github.com/Extra-Chill/data-machine)
**Issue created from these findings:** Extra-Chill/data-machine#721

All 13 findings flag legitimate utility/helper/trait files for not having the \`Ability\` or \`Abilities\` suffix:

| Flagged file | What it actually is | Why the convention doesn't apply |
|---|---|---|
| \`inc/Abilities/PermissionHelper.php\` | Permission utility | Helper for ability checks, not an ability itself |
| \`inc/Abilities/AbilityCategories.php\` | Constants/enum | Category definitions consumed by abilities |
| \`inc/Abilities/Chat/ChatSessionHelpers.php\` | Trait | \`trait\`, mixed into ability classes via \`use\` |
| \`inc/Abilities/Content/ContentActionHandlers.php\` | Action callbacks | WordPress action hook handlers |
| \`inc/Abilities/Content/BlockSanitizer.php\` | Sanitizer utility | Pure function class |
| \`inc/Abilities/Content/CanonicalDiffPreview.php\` | Preview rendering | Used by abilities, not one |
| \`inc/Abilities/Engine/PipelineBatchScheduler.php\` | Scheduler service | Engine subsystem |
| \`inc/Abilities/Engine/EngineHelpers.php\` | Helper trait | \`trait\`, mixed into engine ability classes |
| \`inc/Abilities/File/FileConstants.php\` | Constants class | File-related const definitions |
| \`inc/Abilities/Flow/FlowHelpers.php\` | Helper class | Used by flow abilities |
| \`inc/Abilities/FlowStep/FlowStepHelpers.php\` | Helper class | Used by flow step abilities |
| \`inc/Abilities/Job/JobHelpers.php\` | Helper trait | \`trait\` |
| \`inc/Abilities/Pipeline/PipelineHelpers.php\` | Helper class | Used by pipeline abilities |

The audit suggestion itself acknowledges the false-positive nature:

> Treat this as a utility/helper or rename it to match the 'Ability' convention

\"Treat this as a utility/helper\" means there's no fix — it just **is** a utility/helper. The check is firing because the file lives in a directory named \`Abilities/\` next to files named \`*Ability.php\`.

## Why this matters

Every directory with a primary type + auxiliary helpers/utilities/traits will trip this rule. WordPress conventions especially encourage helper traits (mixed in via \`use Foo;\`), constants files, sanitizers, and renderers as siblings of the main domain class.

## Proposed fix

Detect file purpose, don't pattern-match on directory residency. Suppress the finding when the file is any of:

1. **A trait** — \`trait Foo {\` declaration. Traits are mixed in, not registered as the directory's primary type.
2. **A constants/enum class** — class body is mostly \`const\` or static read-only properties.
3. **A helper class with an obvious naming suffix** — \`*Helper\`, \`*Helpers\`, \`*Constants\`, \`*Sanitizer\`, \`*Renderer\`, \`*Validator\`, \`*Resolver\`, \`*Factory\`, \`*Builder\`. These are widely-recognized PHP/WordPress conventions for non-primary types.
4. **A class without the primary type's interface implementation** — if siblings implement \`SomeInterface\` and this class doesn't, AND it doesn't extend the primary base, it's structurally not the same type.

## Less invasive interim mitigation

Per-rule suffix allowlist in \`homeboy.json\`:

\`\`\`json
{
  \"audit\": {
    \"naming_mismatch\": {
      \"allowed_suffixes\": [\"Helper\", \"Helpers\", \"Constants\", \"Sanitizer\", \"Renderer\", \"Validator\", \"Resolver\", \"Factory\", \"Builder\", \"Categories\"]
    }
  }
}
\`\`\`

## Detection one-liner

\`\`\`bash
# Files declared as traits should never be flagged for class-naming convention:
grep -l '^trait ' --include='*.php' -r inc/Abilities/
\`\`\`

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** Claude Sonnet 4.5 (via Kimaki / Claude Code)
- **Used for:** Diagnosed the false positives while triaging audit issues on data-machine, classified each flagged file by actual purpose, drafted this upstream report.

### LINKED ISSUE #1550: audit: upstream_workaround rule misclassifies historical @see references as active workarounds
## Summary

The \`upstream_workaround\` audit rule fires on any docblock containing \`@see <url-to-issue>\` and labels the file as containing an upstream-bug workaround that should be removed when the upstream is closed. The rule conflates **historical design-source pointers** (\"this file was designed in response to issue N\") with **active workaround markers** (\"this code paper-overs an upstream bug — remove when fixed\"). Most files that survive an issue closure cite the issue as design provenance, not as a workaround target.

## Concrete repro

**Project:** [Extra-Chill/data-machine](https://github.com/Extra-Chill/data-machine)
**Issue created from these findings:** Extra-Chill/data-machine#1242

Two of the three findings:

> \`inc/Api/WebhookAuthResolver.php\` lines 2-21: Webhook auth config resolver.
> Workaround references @see https://github.com/Extra-Chill/data-machine/issues/1179.
> *Suggestion: Check whether the upstream issue/PR is closed or whether the fix has shipped — if so, remove this branch and its comment.*

> \`inc/Api/WebhookSignatureVerifier.php\` lines 2-14: Webhook Signature Verifier — DEPRECATED.
> Workaround references @see https://github.com/Extra-Chill/data-machine/issues/1179.

The actual file docblocks (verbatim from \`WebhookAuthResolver.php\`):

\`\`\`php
/**
 * Webhook auth config resolver.
 *
 * Produces a canonical verifier config for an HMAC flow from its
 * scheduling_config. Stored shape (post-migration) is always:
 * ...
 *
 * No provider names live in this file.
 *
 * @package DataMachine\\Api
 * @since 0.79.0
 * @see https://github.com/Extra-Chill/data-machine/issues/1179
 */
\`\`\`

Issue #1179 is *the design issue that produced this code*. The file is the **shipped result** of #1179, not a workaround for it. The \`@see\` pointer is historical provenance — \"if you want to understand why this code exists, read issue 1179.\"

The rule's suggestion (\"remove this branch and its comment\") would delete:

1. The pointer to the design that produced the file (loss of context for future readers)
2. Possibly the file itself (per the suggestion's framing) — but the file *is* the fix, not the workaround

## How to distinguish workarounds from design pointers

Real upstream-bug workarounds typically have explicit language calling them out as such:

- \"Workaround for X — remove when X is fixed\"
- \"HACK:\" / \"XXX:\" / \"TODO: remove when N closes\"
- \"This branch exists only because <upstream>\"
- @see reference next to a conditional that special-cases something

Design-provenance pointers don't:

- \"@see <issue> for context\"
- \"@see <issue> for the design discussion\"
- \"Implements the contract from <issue>\"
- @see in a docblock without surrounding workaround language

Both contain \`@see <url>\`. The difference is what's around it.

## Proposed fix

Two-tier detection:

1. **Promote to finding only when surrounding language indicates workaround intent.** Keywords/phrases the rule should require:
   - \`workaround\`, \`hack\`, \`HACK:\`, \`XXX:\`
   - \`work[-\\s]?around\`
   - \`special[-\\s]?case\`
   - \`only because\`, \`only exists because\`, \`only here because\`
   - \`temporary\` near \`upstream\` or \`@see\`
   - \`TODO.*remove\`, \`FIXME.*remove\`

   Without one of these, treat \`@see <issue>\` as historical provenance and don't flag.

2. **When flagging, distinguish open from closed upstream.** Currently the rule flags \`@see <closed-issue-url>\` as \"upstream is closed, remove the workaround.\" If the rule can fetch upstream issue state, it can:
   - **Open upstream** → low-priority informational finding (\"workaround referencing open issue\")
   - **Closed upstream** → high-priority actionable finding (\"workaround referencing closed issue, verify the fix shipped\")
   - **No upstream-state lookup** → don't promote based on URL presence alone

   The current behavior promotes on URL presence regardless of intent or upstream state.

## Less invasive interim mitigation

Per-file allowlist in \`homeboy.json\`:

\`\`\`json
{
  \"audit\": {
    \"upstream_workaround\": {
      \"design_provenance_files\": [
        \"inc/Api/WebhookAuthResolver.php\",
        \"inc/Api/WebhookSignatureVerifier.php\"
      ]
    }
  }
}
\`\`\`

\"This file's \`@see\` is provenance, not a workaround marker.\"

## Why this matters

Every project accumulates @see refs to the issues that produced files. A rule that flags all of them as workarounds floods the audit pile with noise and trains contributors to ignore \`upstream_workaround\` findings — which means the *real* workaround findings (the third one in #1242, a legitimate dead PHP-version guard) get lost in the noise.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** Claude Sonnet 4.5 (via Kimaki / Claude Code)
- **Used for:** Diagnosed the false positives while triaging audit issues on data-machine, identified the workaround-vs-provenance distinction, drafted this upstream report. Chris confirmed the upstream-fix-first rule applies.

### LINKED ISSUE #1552: audit: dead_guard rule can't distinguish runtime context (production/test/uninstall/activation)
## Summary

The \`dead_guard\` rule flags every \`class_exists()\` / \`function_exists()\` check whose target is \"guaranteed at runtime\" by the plugin's declared requirements. But \"runtime\" isn't one context — it's at least four:

1. **Production runtime** (plugin booted, WP loaded) — most guards genuinely dead here
2. **Plugin activation/uninstall** (no plugin order guarantee, sibling plugins may be unloaded)
3. **PHPUnit / WP test environment** (WP_Ability available via test bootstrap; WP procedural functions may not be)
4. **Pure-PHP smoke tests** (no WordPress at all — guards are defining stubs, not consuming symbols)

A guard that's dead in (1) is often **load-bearing** in (2)–(4). The rule treats them all the same, producing a high false-positive rate that makes findings unactionable without per-guard manual verification.

## Concrete repro

**Project:** [Extra-Chill/data-machine](https://github.com/Extra-Chill/data-machine)
**Issue created from these findings:** Extra-Chill/data-machine#1210

37 findings reported. Reading 8 representative ones:

### Truly dead (production runtime, AS is a hard composer dep)

- \`inc/Abilities/Engine/ScheduleFlowAbility.php:97,151\` — \`function_exists('as_unschedule_all_actions')\` / \`as_schedule_single_action\`
- \`inc/Engine/Tasks/RecurringScheduler.php:123,174,225,239,384\` — multiple \`as_*\` guards
- \`inc/Core/ActionScheduler/BatchScheduler.php:165,271\`
- \`inc/Engine/AI/System/Tasks/SystemTask.php:185\`, \`ImageGenerationTask.php:261\`
- ~10–15 of 37 total fall in this bucket

### Load-bearing in non-production contexts

- \`data-machine.php:40\` — \`if ( ! class_exists( 'ActionScheduler' ) ) { require __DIR__ . '/vendor/.../action-scheduler.php'; }\`
  **Removing breaks installs where another plugin (WooCommerce, etc.) ships AS first** — double-load fatal.
- \`uninstall.php:113\` — runs at uninstall, when sibling plugins may already be deactivated
- \`inc/migrations/flows.php:19\` — runs at plugin activation, before AS is guaranteed loaded
- \`inc/Core/Auth/CallerContext.php:202\` — \`if ( function_exists('wp_generate_uuid4') ) { ... } else { return bin2hex(random_bytes(16)); }\`
  Docblock explicitly says: \"falls back to a hex random string for testability in environments where WordPress functions aren't loaded.\" Load-bearing for tests.
- \`tests/Unit/Abilities/JobAbilitiesTest.php:14\` — \`if ( ! function_exists('as_schedule_single_action') ) { function as_schedule_single_action(...) {...} }\`
  **This isn't a usage guard — it's a stub definition.** Removing this either breaks PHPUnit (function undefined) or fatals (function double-defined). Comment at line 12 says \"Stub for Action Scheduler function not available in PHPUnit test environment.\"
- \`tests/queueable-trait-smoke.php:18\` — same stub-definition pattern for \`wp_json_encode\`

The rule flagged all of these identically.

### Already removed in #1259

- \`tests/Unit/Abilities/LogAbilitiesTest.php\` — 9 \`class_exists('WP_Ability')\` guards inside test methods. Those *were* truly dead (WP test bootstrap loads core), but removing them required reading each one to confirm they weren't stub definitions.

### Auditor parsing bug

- \`homeboy.json:29,33\` — flagged as \`class_exists('WP_Ability')\` dead guards. They're regex *strings* inside the existing \`drop_wp_ability_dead_guards\` transform definition, not live PHP. The rule is reading JSON content as code.

## Proposed fix

Three corrections, in priority order:

### 1. Distinguish stub-definition from usage-guard

The shape \`if ( ! function_exists( X ) ) { function X(...) {...} }\` is a **stub definition**, not a usage guard. The rule must:

- Suppress findings when the guard's body contains a \`function X(...)\` declaration matching the guarded symbol.
- Suppress findings when the guard's body contains a \`require_once\` of a vendor file that defines the guarded symbol (covers \`data-machine.php:40\`).

These two patterns alone account for ~5 of the 37 findings on data-machine and 100% of the load-bearing-in-tests false positives.

### 2. Detect non-production contexts by file path

Files under \`tests/\`, \`uninstall.php\`, \`activate.php\`, \`inc/migrations/*.php\`, and similar lifecycle paths run in contexts where the plugin's runtime requirements don't apply. Either:

- **Suppress entirely** for these paths (cheap, high precision)
- **Downgrade severity** to informational (\"may be intentional in this context\")

### 3. Don't read JSON content as code

The rule scanned \`homeboy.json\` and flagged regex *strings* as dead guards. JSON files (and other non-PHP source) should be excluded from guard analysis. Cheap parser-level fix.

## Less invasive interim mitigation

Per-rule path allowlist in \`homeboy.json\`:

\`\`\`json
{
  \"audit\": {
    \"dead_guard\": {
      \"exclude_paths\": [
        \"tests/**\",
        \"uninstall.php\",
        \"data-machine.php\",
        \"inc/migrations/*.php\",
        \"inc/Core/Auth/CallerContext.php\"
      ]
    }
  }
}
\`\`\`

Plus exclude non-PHP files from the rule's input glob (probably a fix in the rule itself, not the project config).

## Why this matters

37 findings is a small enough pile that reading each one is feasible. But the false-positive rate (~50%) means contributors need per-guard judgment about reachability across 4 distinct runtime contexts — which is exactly the analysis the rule was meant to do.

If left to its current behavior, contributors will either:
- Skip this rule's findings (the audit pile grows, signal lost)
- Apply blanket fixes (breaks tests, breaks uninstall, double-loads in WC environments)

Both failure modes lose the value of the rule. Fixing the four context-distinction cases above gets the rule to ~5–10% false positive — actionable for the high-precision portion, with the rest tagged correctly.

## Companion fixes

- homeboy#1551 (this same triage session) — closed audit issues get re-filed on every CI run, undoing the false-positive triage work above. The two issues compound: the rule produces noisy false positives AND triage decisions don't stick.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** Claude Sonnet 4.5 (via Kimaki / Claude Code)
- **Used for:** Walked 8 of the 37 findings in #1210 to classify each by runtime context, identified the 4 distinct context categories, identified the JSON-as-PHP parser bug, and the stub-definition vs usage-guard distinction. Drafted this upstream report.

chubes4:
## Summary

- Keep Homeboy core framework-agnostic while preserving the false-positive reductions from this branch.
- Add generic audit detector rules that can come from extension manifests and component config.
- Remove the WordPress/WP-CLI-shaped regression fixtures from the new tests; the detector tests now use generic runtime-dispatch/lifecycle/utility examples.

## Changes

- Adds `audit.detector_rules` to extension manifests, plus component-level `audit` rules as an additive override.
- Merges extension-provided rules before component-provided rules when running audit, preferring the worktree's portable `homeboy.json` over the registered primary checkout for path-based audits.
- Moves audit detector config into `src/core/component/audit.rs` to keep `component/mod.rs` below the structural threshold.
- Removes new compatibility wrapper functions that were becoming audit/compiler dead-code findings.
- Keeps runtime entrypoint markers, lifecycle path globs, utility suffixes, and convention exceptions as data, not framework knowledge in core detectors.

## Tests

- `cargo test core::code_audit::conventions --lib`
- `cargo test core::code_audit::dead_code --lib`
- `cargo test core::code_audit::dead_guard --lib`
- `cargo test commands::audit::tests::audit_detects_outliers_in_convention_group --bin homeboy`
- `homeboy audit homeboy --path /Users/chubes/Developer/homeboy@audit-false-positive-rules` — fails on unrelated pre-existing `src/core/triage.rs` drift; the branch-introduced audit-config issues from this rework are cleared.

## Closes

- Closes #1544
- Closes #1545
- Closes #1546
- Closes #1547
- Closes #1548
- Closes #1550
- Closes #1552

## AI assistance

- **AI assistance:** Yes
- **Used for:** Refactored the audit false-positive rules into generic extension/component-provided detector config and updated tests; Chris remains responsible for review and merge.

### DIFF
--- a/src/commands/audit.rs
+++ b/src/commands/audit.rs
@@ -225,12 +225,12 @@ mod tests {
 
         fs::write(
             root.join("commands/good_one.rs"),
-            "pub fn run() {}\npub fn helper() {}\n",
+            "pub fn run() {}\npub fn execute() {}\n",
         )
         .unwrap();
         fs::write(
             root.join("commands/good_two.rs"),
-            "pub fn run() {}\npub fn helper() {}\n",
+            "pub fn run() {}\npub fn execute() {}\n",
         )
         .unwrap();
         fs::write(root.join("commands/bad.rs"), "pub fn run() {}\n").unwrap();
--- a/src/core/code_audit/conventions.rs
+++ b/src/core/code_audit/conventions.rs
@@ -351,9 +351,15 @@ pub fn discover_conventions(
         }
     }
 
+    let declared_traits: Vec<String> = fingerprints
+        .iter()
+        .filter_map(declared_trait_name)
+        .collect();
+
     let expected_interfaces: Vec<String> = interface_counts
         .iter()
         .filter(|(_, count)| **count >= threshold)
+        .filter(|(name, _)| !declared_traits.contains(name))
         .map(|(name, _)| name.clone())
         .collect();
 
@@ -412,10 +418,11 @@ pub fn discover_conventions(
                     .iter()
                     .all(|name| !suffix_matches(name, suffix))
         });
+        let utility_like = helper_like && is_utility_like_file(fp);
 
         let mut deviations = Vec::new();
 
-        if helper_like {
+        if helper_like && !utility_like {
             let suffix = naming_suffix.as_deref().unwrap_or("member");
             deviations.push(Deviation {
                 kind: AuditFinding::NamingMismatch,
@@ -569,6 +576,41 @@ pub fn discover_conventions(
         total_files: total,
         confidence,
     })
+}
+
+fn declared_trait_name(fp: &FileFingerprint) -> Option<String> {
+    let re = regex::Regex::new(r"(?m)^\s*trait\s+([A-Za-z_][A-Za-z0-9_]*)\b").ok()?;
+    re.captures(&fp.content)
+        .and_then(|cap| cap.get(1))
+        .map(|m| m.as_str().to_string())
+}
+
+fn is_utility_like_file(fp: &FileFingerprint) -> bool {
+    let suffixes = [
+        "Helper",
+        "Helpers",
+        "Constants",
+        "Categories",
+        "Sanitizer",
+        "Renderer",
+        "Validator",
+        "Verifier",
+        "Resolver",
+        "Factory",
+        "Builder",
+        "Result",
+        "Scheduling",
+    ];
+    let names_to_check: Vec<&str> = if !fp.type_names.is_empty() {
+        fp.type_names.iter().map(|s| s.as_str()).collect()
+    } else {
+        fp.type_name.as_deref().into_iter().collect()
+    };
+
+    declared_trait_name(fp).is_some()
+        || names_to_check
+            .iter()
+            .any(|name| suffixes.iter().any(|suffix| name.ends_with(suffix)))
 }
 
 // ============================================================================
@@ -854,7 +896,7 @@ mod tests {
     }
 
     #[test]
-    fn helper_like_outlier_collapses_to_naming_mismatch() {
+    fn utility_like_outlier_is_not_promoted_to_naming_mismatch() {
         let fingerprints = vec![
             FileFingerprint {
                 relative_path: "abilities/CreateAbility.php".to_string(),
@@ -882,13 +924,92 @@ mod tests {
         let convention =
             discover_conventions("Abilities", "abilities/*.php", &fingerprints).unwrap();
 
+        assert!(
+            convention.outliers.is_empty(),
+            "recognized helper files are intentional utilities, got: {:?}",
+            convention.outliers
+        );
+    }
+
+    #[test]
+    fn non_utility_helper_like_outlier_still_reports_naming_mismatch() {
+        let fingerprints = vec![
+            FileFingerprint {
+                relative_path: "abilities/CreateAbility.php".to_string(),
+                language: Language::Php,
+                methods: vec!["execute".to_string(), "register".to_string()],
+                type_name: Some("CreateAbility".to_string()),
+                ..Default::default()
+            },
+            FileFingerprint {
+                relative_path: "abilities/UpdateAbility.php".to_string(),
+                language: Language::Php,
+                methods: vec!["execute".to_string(), "register".to_string()],
+                type_name: Some("UpdateAbility".to_string()),
+                ..Default::default()
+            },
+            FileFingerprint {
+                relative_path: "abilities/FlowThing.php".to_string(),
+                language: Language::Php,
+                methods: vec!["formatFlow".to_string()],
+                type_name: Some("FlowThing".to_string()),
+                ..Default::default()
+            },
+        ];
+
+        let convention =
+            discover_conventions("Abilities", "abilities/*.php", &fingerprints).unwrap();
+
         assert_eq!(convention.outliers.len(), 1);
-        assert!(convention.outliers[0].noisy);
-        assert_eq!(convention.outliers[0].deviations.len(), 1);
         assert!(matches!(
             convention.outliers[0].deviations[0].kind,
             AuditFinding::NamingMismatch
         ));
+    }
+
+    #[test]
+    fn declared_traits_do_not_become_missing_interfaces() {
+        let fingerprints = vec![
+            FileFingerprint {
+                relative_path: "chat/ListChatSessionsAbility.php".to_string(),
+                language: Language::Php,
+                methods: vec!["execute".to_string()],
+                type_name: Some("ListChatSessionsAbility".to_string()),
+                implements: vec!["ChatSessionHelpers".to_string()],
+                ..Default::default()
+            },
+            FileFingerprint {
+                relative_path: "chat/DeleteChatSessionAbility.php".to_string(),
+                language: Language::Php,
+                methods: vec!["execute".to_string()],
+                type_name: Some("DeleteChatSessionAbility".to_string()),
+                implements: vec!["ChatSessionHelpers".to_string()],
+                ..Default::default()
+            },
+            FileFingerprint {
+                relative_path: "chat/ChatSessionHelpers.php".to_string(),
+                language: Language::Php,
+                methods: vec!["verifySessionOwnership".to_string()],
+                type_name: Some("ChatSessionHelpers".to_string()),
+                content: "<?php\ntrait ChatSessionHelpers {}".to_string(),
+                ..Default::default()
+            },
+        ];
+
+        let convention = discover_conventions("Chat", "chat/*.php", &fingerprints).unwrap();
+        assert!(
+            convention.expected_interfaces.is_empty(),
+            "traits should not be treated as interfaces: {:?}",
+            convention.expected_interfaces
+        );
+        assert!(
+            convention
+                .outliers
+                .iter()
+                .flat_map(|o| &o.deviations)
+                .all(|d| d.kind != AuditFinding::MissingInterface),
+            "declared trait should not produce MissingInterface deviations"
+        );
     }
 
     #[test]
--- a/src/core/code_audit/conventions.rs
+++ b/src/core/code_audit/conventions.rs
@@ -285,19 +285,6 @@ impl std::str::FromStr for AuditFinding {
 /// The algorithm:
 /// 1. Find methods that appear in ≥ 60% of files (the "convention")
 /// 2. Find files that are missing any of those methods (the "outliers")
-pub fn discover_conventions(
-    group_name: &str,
-    glob_pattern: &str,
-    fingerprints: &[FileFingerprint],
-) -> Option<Convention> {
-    discover_conventions_with_config(
-        group_name,
-        glob_pattern,
-        fingerprints,
-        &AuditConfig::default(),
-    )
-}
-
 pub fn discover_conventions_with_config(
     group_name: &str,
     glob_pattern: &str,
@@ -853,7 +840,20 @@ pub fn check_signature_consistency(conventions: &mut [Convention], root: &Path) 
 mod tests {
     use super::*;
 
-    fn wordpress_like_audit_config() -> AuditConfig {
+    fn discover_conventions(
+        group_name: &str,
+        glob_pattern: &str,
+        fingerprints: &[FileFingerprint],
+    ) -> Option<Convention> {
+        discover_conventions_with_config(
+            group_name,
+            glob_pattern,
+            fingerprints,
+            &AuditConfig::default(),
+        )
+    }
+
+    fn framework_like_audit_config() -> AuditConfig {
         AuditConfig {
             utility_suffixes: vec![
                 "Helper".to_string(),
@@ -958,7 +958,7 @@ mod tests {
             "Abilities",
             "abilities/*.php",
             &fingerprints,
-            &wordpress_like_audit_config(),
+            &framework_like_audit_config(),
         )
         .unwrap();
 
@@ -1038,7 +1038,7 @@ mod tests {
             "Chat",
             "chat/*.php",
             &fingerprints,
-            &wordpress_like_audit_config(),
+            &framework_like_audit_config(),
         )
         .unwrap();
         assert!(
@@ -1057,38 +1057,38 @@ mod tests {
     }
 
     #[test]
-    fn rest_utility_classes_do_not_need_route_registration() {
+    fn utility_classes_do_not_need_dispatch_registration() {
         let fingerprints = vec![
             FileFingerprint {
-                relative_path: "api/PostsController.php".to_string(),
+                relative_path: "endpoints/PostsEndpoint.php".to_string(),
                 language: Language::Php,
-                methods: vec!["register_routes".to_string()],
-                registrations: vec!["rest_api_init".to_string()],
-                type_name: Some("PostsController".to_string()),
+                methods: vec!["register".to_string()],
+                registrations: vec!["runtime_dispatch".to_string()],
+                type_name: Some("PostsEndpoint".to_string()),
                 ..Default::default()
             },
             FileFingerprint {
-                relative_path: "api/PagesController.php".to_string(),
+                relative_path: "endpoints/PagesEndpoint.php".to_string(),
                 language: Language::Php,
-                methods: vec!["register_routes".to_string()],
-                registrations: vec!["rest_api_init".to_string()],
-                type_name: Some("PagesController".to_string()),
+                methods: vec!["register".to_string()],
+                registrations: vec!["runtime_dispatch".to_string()],
+                type_name: Some("PagesEndpoint".to_string()),
                 ..Default::default()
             },
             FileFingerprint {
-                relative_path: "api/WebhookVerifier.php".to_string(),
+                relative_path: "endpoints/SignatureVerifier.php".to_string(),
                 language: Language::Php,
                 methods: vec!["verify".to_string()],
-                type_name: Some("WebhookVerifier".to_string()),
+                type_name: Some("SignatureVerifier".to_string()),
                 ..Default::default()
             },
         ];
 
         let convention = discover_conventions_with_config(
             "Api",
-            "api/*.php",
+            "endpoints/*.php",
             &fingerprints,
-            &wordpress_like_audit_config(),
+            &framework_like_audit_config(),
         )
         .unwrap();
         assert!(
@@ -1098,7 +1098,7 @@ mod tests {
                 .flat_map(|o| &o.deviations)
                 .all(|d| d.kind != AuditFinding::MissingRegistration
                     && d.kind != AuditFinding::MissingMethod),
-            "REST utility classes should not be treated as endpoint registrants"
+            "utility classes should not be treated as runtime endpoint registrants"
         );
     }
 
@@ -1132,7 +1132,7 @@ mod tests {
             "Chat",
             "chat/*.php",
             &fingerprints,
-            &wordpress_like_audit_config(),
+            &framework_like_audit_config(),
         )
         .unwrap();
         assert!(
--- a/src/core/code_audit/conventions.rs
+++ b/src/core/code_audit/conventions.rs
@@ -1013,6 +1013,83 @@ mod tests {
     }
 
     #[test]
+    fn rest_utility_classes_do_not_need_route_registration() {
+        let fingerprints = vec![
+            FileFingerprint {
+                relative_path: "api/PostsController.php".to_string(),
+                language: Language::Php,
+                methods: vec!["register_routes".to_string()],
+                registrations: vec!["rest_api_init".to_string()],
+                type_name: Some("PostsController".to_string()),
+                ..Default::default()
+            },
+            FileFingerprint {
+                relative_path: "api/PagesController.php".to_string(),
+                language: Language::Php,
+                methods: vec!["register_routes".to_string()],
+                registrations: vec!["rest_api_init".to_string()],
+                type_name: Some("PagesController".to_string()),
+                ..Default::default()
+            },
+            FileFingerprint {
+                relative_path: "api/WebhookVerifier.php".to_string(),
+                language: Language::Php,
+                methods: vec!["verify".to_string()],
+                type_name: Some("WebhookVerifier".to_string()),
+                ..Default::default()
+            },
+        ];
+
+        let convention = discover_conventions("Api", "api/*.php", &fingerprints).unwrap();
+        assert!(
+            convention
+                .outliers
+                .iter()
+                .flat_map(|o| &o.deviations)
+                .all(|d| d.kind != AuditFinding::MissingRegistration
+                    && d.kind != AuditFinding::MissingMethod),
+            "REST utility classes should not be treated as endpoint registrants"
+        );
+    }
+
+    #[test]
+    fn factories_do_not_need_methods_of_created_type() {
+        let fingerprints = vec![
+            FileFingerprint {
+                relative_path: "chat/DatabaseConversationStore.php".to_string(),
+                language: Language::Php,
+                methods: vec!["update_title".to_string(), "delete_session".to_string()],
+                type_name: Some("DatabaseConversationStore".to_string()),
+                ..Default::default()
+            },
+            FileFingerprint {
+                relative_path: "chat/MemoryConversationStore.php".to_string(),
+                language: Language::Php,
+                methods: vec!["update_title".to_string(), "delete_session".to_string()],
+                type_name: Some("MemoryConversationStore".to_string()),
+                ..Default::default()
+            },
+            FileFingerprint {
+                relative_path: "chat/ConversationStoreFactory.php".to_string(),
+                language: Language::Php,
+                methods: vec!["get".to_string()],
+                type_name: Some("ConversationStoreFactory".to_string()),
+                ..Default::default()
+            },
+        ];
+
+        let convention = discover_conventions("Chat", "chat/*.php", &fingerprints).unwrap();
+        assert!(
+            convention
+                .outliers
+                .iter()
+                .flat_map(|o| &o.deviations)
+                .all(|d| d.kind != AuditFinding::MissingMethod),
+            "factories produce stores; they should not implement store methods"
+        );
+    }
+
+    #[test]
     fn no_interface_convention_when_none_shared() {
         let fingerprints = vec![
             FileFingerprint {
--- a/src/core/code_audit/conventions.rs
+++ b/src/core/code_audit/conventions.rs
@@ -11,6 +11,7 @@ use super::fingerprint::FileFingerprint;
 use super::import_matching::has_import_with_context;
 use super::naming::{detect_naming_suffix, suffix_matches};
 use super::signatures::{compute_signature_skeleton, tokenize_signature};
+use crate::component::AuditConfig;
 
 #[derive(Debug, Clone, Default, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
 #[serde(rename_all = "lowercase")]
@@ -289,6 +290,20 @@ pub fn discover_conventions(
     glob_pattern: &str,
     fingerprints: &[FileFingerprint],
 ) -> Option<Convention> {
+    discover_conventions_with_config(
+        group_name,
+        glob_pattern,
+        fingerprints,
+        &AuditConfig::default(),
+    )
+}
+
+pub fn discover_conventions_with_config(
+    group_name: &str,
+    glob_pattern: &str,
+    fingerprints: &[FileFingerprint],
+    audit_config: &AuditConfig,
+) -> Option<Convention> {
     if fingerprints.len() < 2 {
         return None; // Need at least 2 files to detect a pattern
     }
@@ -418,11 +433,12 @@ pub fn discover_conventions(
                     .iter()
                     .all(|name| !suffix_matches(name, suffix))
         });
-        let utility_like = helper_like && is_utility_like_file(fp);
+        let utility_like = helper_like && is_utility_like_file(fp, audit_config);
+        let convention_exempt = is_convention_exception(fp, audit_config);
 
         let mut deviations = Vec::new();
 
-        if helper_like && !utility_like {
+        if helper_like && !utility_like && !convention_exempt {
             let suffix = naming_suffix.as_deref().unwrap_or("member");
             deviations.push(Deviation {
                 kind: AuditFinding::NamingMismatch,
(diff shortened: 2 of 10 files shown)


chubes4:
Hardcoding check from the audit-routing stabilization pass:

This PR appears to add WordPress/PHP-plugin-specific false-positive handling directly to Homeboy core. Examples from the diff include:
- `is_wp_cli_command_file()` and WP-CLI command heuristics (`WP_CLI_Command`, `@subcommand`, `## OPTIONS`, `## EXAMPLES`).
- lifecycle/path assumptions such as `uninstall.php`, `activate.php`, and `inc/migrations/`.
- REST/route registration expectations in tests (`rest_api_init`, `register_routes`) and Data Machine-flavoured examples.

Chris's constraint for this pass: Homeboy core must remain language/framework agnostic and must not know what WordPress, WP-CLI, REST, PHP plugin conventions, or Data Machine-specific class names/options are.

Recommendation: do not land this shape in Homeboy core as-is. Move this behind generic detector abstractions, fixture/config-driven rules, language-extension hooks, or project config so WordPress/PHP-plugin knowledge lives outside core.

chubes4:
Stabilization pass note: I would not treat this as just a CI/audit cleanup PR in its current shape.

Homeboy core needs to stay language/framework agnostic. This branch reduces WordPress false positives by baking WordPress-specific concepts directly into core audit logic, including:

- `src/core/code_audit/dead_code.rs`: `is_wp_cli_command_file()` checks `WP_CLI_Command`, `BaseCommand`, `@subcommand`, `## OPTIONS`, and `## EXAMPLES`.
- `src/core/code_audit/dead_guard.rs`: lifecycle/test context detection knows `uninstall.php`, `activate.php`, and `inc/migrations/` paths.
- `src/core/code_audit/conventions.rs`: tests and messaging explicitly model “REST utility classes”.
- `src/core/code_audit/requirements.rs` already models WordPress plugin headers and WP core symbol/version baselines; this PR leans further into that coupling.
- `src/core/code_audit/upstream_workaround.rs` treats `$wp_version` as a first-class version-guard input.

Those are real false-positive classes, but the implementation puts framework knowledge in the generic auditor. The generic shape should be something like:

- Runtime entrypoint patterns supplied by project config or extension metadata, e.g. class-extends patterns, docblock markers, command-section markers, and lifecycle paths.
- Language-neutral convention suppressions for utility/factory/verifier/resolver classes, driven by configurable suffix/category patterns rather than REST/WordPress names.
- Requirement/baseline symbol providers supplied by extensions or fixture packs, e.g. “this project guarantees these functions/classes/constants at version X”, not hardcoded WP core symbol tables.
- Fixture-driven tests that prove the detector accepts externally supplied framework patterns, with WordPress fixtures living outside core or behind a generic extension hook.

Recommendation: block this PR until the WordPress-specific rules are moved behind generic detector inputs/config/extension hooks. Passing CI with these rules embedded in core would make Homeboy a WordPress-aware auditor, which violates the core boundary.

chubes4:
This needs rework before merge under the Homeboy-core agnosticism constraint.

The current diff hardcodes framework/project-specific knowledge in core audit logic:

- `dead_code.rs`: `is_wp_cli_command_file()`, `WP_CLI_Command`, `@subcommand`, `## OPTIONS`, `## EXAMPLES`.
- `dead_guard.rs`: lifecycle/test path assumptions like `uninstall.php`, `activate.php`, `inc/migrations/`.
- `conventions.rs` tests and framing are WordPress/REST-specific even where the underlying heuristic could be generic.

The desired shape is a generic extension/config mechanism: runtime entrypoint markers, lifecycle/test path globs, utility suffixes, and convention exceptions should come from component config or language-extension rules, not Homeboy core knowing WordPress/WP-CLI/Data Machine conventions.

Please refactor in that direction or split this into a generic primitive plus a repo-local config update. Do not merge this form.

chubes4:
Refactored the blocker examples out of Homeboy core in commit 55e7d2d3. The PR now uses generic component audit config for runtime entrypoint extends/markers, lifecycle path globs, utility suffixes, and convention exception globs. WordPress/WP-CLI-shaped values only appear in tests as supplied config data, not as detector defaults.\n\nTests rerun:\n- cargo test core::code_audit::conventions --lib\n- cargo test core::code_audit::dead_code --lib\n- cargo test core::code_audit::dead_guard --lib\n- cargo test commands::audit::tests::audit_detects_outliers_in_convention_group --bin homeboy\n- cargo test -- --test-threads=1

chubes4:
Reworked the branch so the false-positive reductions are driven by generic extension/component detector config rather than WordPress/WP-CLI-shaped core rules. Pushed ebc3a01e. Focused tests pass; full branch audit still fails only on unrelated existing src/core/triage.rs drift.

chubes4:
Reviewed the reworked branch against the prior architecture blocker. The final detector logic is generic: runtime entrypoints, lifecycle globs, utility suffixes, and convention exceptions now come from extension/component AuditConfig rather than Homeboy core hardcoding WordPress/WP-CLI/Data Machine rules. CI is green. Residual risk is mainly config coverage: framework-specific extensions/components need to provide the right detector rules for their ecosystems.

PR #1644, opened 2026-04-27
fix(audit): normalize abstract signature declarations

### LINKED ISSUE #1623: audit: signature_mismatch still counts PHP return type tokens
## Summary

`signature_mismatch` is still flagging PHP methods where the only reported structural difference is a return type token. This appears to be the same class as closed issue #571, but it still reproduces on `homeboy 0.102.0+b622c7d1` via Extra-Chill/data-machine#1321.

## Concrete repro

**Downstream issue:** Extra-Chill/data-machine#1321

**Reported finding:**

```text
inc/Engine/AI/System/Tasks/Retention/RetentionTask.php
Signature mismatch for runRetentionCleanup: different structure — expected 5 tokens, found 6.
Example: abstract protected function runRetentionCleanup(): array;
Suggestion: protected function runRetentionCleanup ( )
```

**Actual code:**

```php
// RetentionTask.php
abstract protected function runRetentionCleanup(): array;

// Every concrete retention task.
protected function runRetentionCleanup(): array {
    // ...
}
```

All concrete implementations match the abstract method shape, including the `array` return type. The suggested fix would remove useful type information and make the abstract signature less precise.

## Expected behavior

The rule should treat `abstract protected function runRetentionCleanup(): array;` as structurally compatible with `protected function runRetentionCleanup(): array { ... }`.

If the detector intentionally ignores visibility modifiers or `abstract`, it should still preserve return type declarations as part of the canonical signature instead of counting `: array` as unexpected drift.

## Actual behavior

The return type appears to add an extra token, producing a false positive and an unsafe suggestion to remove the return type.

## Related

- Extra-Chill/homeboy#571 — closed earlier report for return-type token false positives.
- Extra-Chill/data-machine#1321 — current downstream audit issue generated by Homeboy Action.

## AI assistance
- **AI assistance:** Yes
- **Used for:** Reproduced the downstream finding, compared the abstract and concrete signatures, and drafted this report. Chris remains responsible.

chubes4:
## Summary
- Fix `signature_mismatch` false positives when comparing abstract PHP method declarations with concrete implementations.
- Normalize declaration-only syntax before token comparison so return-type-bearing abstract declarations preserve the callable shape.

## Changes
- Strip `abstract` as a declaration-only modifier during signature normalization.
- Strip trailing `;` from body-less declarations after return-type normalization.
- Add PHP-focused token regression tests for abstract return types and declaration terminators.

## Tests
- `cargo fmt --check`
- `cargo test core::code_audit::signatures::tests`

Closes #1623

## AI assistance
- **AI assistance:** Yes
- **Used for:** Inspected the issue and detector, implemented the signature normalization fix, added focused regression tests, and ran targeted validation. Chris remains responsible.

### DIFF
--- a/src/core/code_audit/signatures.rs
+++ b/src/core/code_audit/signatures.rs
@@ -33,6 +33,8 @@ pub(crate) fn normalize_signature(sig: &str) -> String {
         .replace_all(&normalized, "")
         .to_string();
 
+    let normalized = strip_declaration_only_modifiers(&normalized);
+
     // Strip trailing body markers FIRST — the `{ }` or `{ ... }` are not part of
     // the function's structural contract and contain parens (e.g., `Ok(())`) that
     // confuse `strip_return_type`'s `rfind(')')`.
@@ -44,6 +46,7 @@ pub(crate) fn normalize_signature(sig: &str) -> String {
     // PHP:  "function foo(): ?array" → "function foo()"
     // Rust: "fn foo() -> Result<T>" → "fn foo()"
     let normalized = strip_return_type(&normalized);
+    let normalized = strip_declaration_terminator(&normalized);
 
     // Strip parameter type annotations — only arity and parameter names matter
     // for structural comparison.  This is language-agnostic: for each comma-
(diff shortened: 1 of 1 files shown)


chubes4:
Reviewed the signature normalization change. I did not find blocking issues.

The ordering in `src/core/code_audit/signatures.rs` looks correct for #1623: `abstract` is removed before body/return normalization, and the declaration terminator is removed after return-type stripping, which covers both `abstract protected function runRetentionCleanup(): array;` and no-return-type declarations. The new tests pin both cases.

Residual risk: the `abstract` stripping is intentionally broad across signature strings, but it only removes the standalone keyword plus following whitespace; I do not see a practical false-negative path for the supported PHP/Rust signature shapes. CI is green, including Build, Audit, Lint, Test, and Auto-refactor.

Recommendation: merge.

PR #1941, opened 2026-04-29
fix(audit): recognize WP-CLI command exports

### LINKED ISSUE #1544: audit: unreferenced_export rule misses WP-CLI subcommand methods

chubes4:
## Summary
- Recognize PHP classes registered through `WP_CLI::add_command(..., SomeCommand::class)` as runtime-dispatched command classes.
- Suppress `unreferenced_export` findings for public methods on those registered command classes, while keeping docblock-only `@subcommand` methods flagged without registration evidence.

## Tests
- `cargo fmt --check`
- `cargo test core::code_audit::dead_code -- --test-threads=1`
- `homeboy audit homeboy --path /Users/chubes/Developer/homeboy@fix-audit-wp-cli-subcommand-exports --changed-since origin/main`
- `homeboy lint homeboy --path /Users/chubes/Developer/homeboy@fix-audit-wp-cli-subcommand-exports --file src/core/code_audit/dead_code.rs`

Closes #1544

## AI assistance
- **AI assistance:** Yes
- **Used for:** Implemented the detector change and focused tests; Chris remains responsible for review and merge.

### DIFF
--- a/src/core/code_audit/dead_code.rs
+++ b/src/core/code_audit/dead_code.rs
@@ -43,6 +43,35 @@ fn build_caller_map(
     map
 }
 
+/// Find PHP classes registered as WP-CLI commands.
+///
+/// WP-CLI dispatches public methods on registered command classes by runtime
+/// reflection, so those methods have no normal PHP call sites.
+fn wp_cli_command_classes<'a>(
+    fingerprints: impl Iterator<Item = &'a FileFingerprint>,
+) -> HashSet<String> {
+    static WP_CLI_COMMAND_RE: std::sync::LazyLock<regex::Regex> = std::sync::LazyLock::new(|| {
+        regex::Regex::new(r"WP_CLI::add_command\s*\([^,]+,\s*([A-Za-z_\\][A-Za-z0-9_\\]*)::class")
+            .unwrap()
+    });
+
+    let mut classes = HashSet::new();
+    for fp in fingerprints {
+        if !matches!(fp.language, super::conventions::Language::Php) {
+            continue;
+        }
+
+        for caps in WP_CLI_COMMAND_RE.captures_iter(&fp.content) {
+            let class_name = caps[1].trim_start_matches('\\');
+            classes.insert(class_name.to_string());
+            if let Some(short_name) = class_name.split('\\').next_back() {
+                classes.insert(short_name.to_string());
+            }
+        }
+    }
+    classes
+}
+
 /// Analyze fingerprints for dead code patterns.
 ///
 /// Performs four checks on `owned` fingerprints:
@@ -77,6 +106,9 @@ pub(crate) fn analyze_dead_code_with_config(
 
     // Build cross-file caller map for parameter analysis (#824).
     let caller_map = build_caller_map(owned, reference);
+
+    let wp_cli_command_classes =
+        wp_cli_command_classes(owned.iter().chain(reference.iter()).copied());
 
     // Only check owned fingerprints for dead code — reference fingerprints
     // just provide call/import data for the cross-reference set.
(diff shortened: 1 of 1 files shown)


chubes4:
Closing this because it hardcodes WP-CLI runtime-dispatch semantics in Homeboy core. That violates the core/extension boundary. Replacement work is tracked in #1945 and should move the framework-specific detection into extension/fingerprint metadata while keeping core generic.

Issue #2778, opened 2026-05-24
audit: detect inconsistent dry-run and skipped-action status reporting


chubes4:
## Problem

Recent QA found command status semantics that are partly domain-specific but still have a generic audit shape:

- #2748 release `--deploy` dry-run reports deployment planned when release is skipped
- #2766 refactor transform no-match dry-runs should be successful empty results

Both are examples of skipped/no-op work being reported inconsistently. Homeboy commands should distinguish planned, skipped, empty, dry-run, failed, and completed actions in a machine-readable way.

## Proposed generic audit

Add a no-op/dry-run status contract detector for commands that support dry-run, planning, or skip paths.

The detector should verify that declared command scenarios expose consistent status semantics:

- Dry-run planned action.
- Dry-run skipped action.
- Successful empty/no-match action.
- Failed action.

## Acceptance criteria

- Commands with dry-run/planning modes can declare scenario fixtures for planned/skipped/empty/failure paths.
- Audit flags outputs where a skipped action is reported as planned or completed.
- Audit flags no-match/no-op dry-runs that exit as errors when the command contract says empty success is expected.
- Findings point at the scenario and the inconsistent status fields.

PR #2798, opened 2026-05-25
Treat empty refactor transforms as success

### LINKED ISSUE #2766: Treat refactor transform no-match dry-runs as a successful empty result
## Problem

`homeboy refactor transform` reports a no-match dry-run as `success: false` with exit code `1`, even though the command completed normally and returned a structured empty result. In automation, this makes "nothing to change" look like a command failure.

## Reproduction

On Homeboy `43544a9e` / `v0.197.10`, use any component path with files that do not contain the searched token, then run:

```sh
/usr/bin/time -lp homeboy refactor transform --path <component-or-fixture> --find NO_MATCH_TOKEN --replace NEW --files 'src/**/*.txt' --output refactor-transform-no-match.json > refactor-transform-no-match.stdout 2> refactor-transform-no-match.time
```

The bounded QA fixture used 10,000 text files and no matches for `NO_MATCH_TOKEN`.

## Observed

The command exits `1` and writes:

```json
{
  "success": false,
  "data": {
    "command": "refactor.transform",
    "name": "ad-hoc",
    "rules": [
      {
        "description": "",
        "id": "ad-hoc",
        "matches": [],
        "replacement_count": 0
      }
    ],
    "total_files": 0,
    "total_replacements": 0,
    "written": false
  }
}
```

The measured 10,000-file no-match run took `4.44s`, peaked at `19,644,416` bytes RSS, and produced only a normal empty result. There was no validation error or runtime failure.

Source shape:

- `src/commands/refactor.rs` sets `let exit_code = if result.total_replacements == 0 { 1 } else { 0 };` for transform.
- The generic response wrapper maps non-zero exit with data to `success: false`.

## Expected

A completed dry-run with zero matches should be a successful empty result, similar to many search/preview operations that return structured zero-count output. If callers need grep-like semantics, that should be an explicit mode or a clearly named field, not the default command success state.

## Acceptance criteria

- No-match `refactor transform` dry-runs exit `0` and return `success: true` with `total_replacements: 0`.
- If write mode also finds no matches, behavior is documented and consistent with dry-run behavior.
- Tests cover no-match transform JSON output and exit code.

chubes4:
## Summary
- Treat `homeboy refactor transform` no-match runs as successful empty results instead of command failures.
- Add a generic `command_status_contracts` audit detector for declared JSON scenario fixtures and wire Homeboy's refactor no-match fixture through it.
- Add focused regression coverage for transform empty results and status-contract mismatches.

Fixes #2766.
Refs #2778.

## Evidence
- Reproduced pre-fix no-match transform output as `success: false` with exit code `1`.
- Post-fix no-match transform CLI output returns `success: true`, `total_replacements: 0`, `written: false`, and exit code `0`.
- `cargo fmt --check`
- `cargo test transform_no_match_dry_run_is_successful_empty_result`
- `cargo test command_status_contracts`
- `cargo run --quiet --bin homeboy -- audit homeboy --path /Users/chubes/Developer/homeboy@fix-refactor-dry-run-empty-success --only command_status_contract_violation --ignore-baseline --json-summary`
- `cargo test --lib`

## Notes
- One earlier full `cargo test` run failed in unrelated `commands::runs::remote_artifact::tests::cleanup_downloads_plans_and_removes_runner_cache`; rerunning that exact test passed, and the subsequent full library suite passed.

## AI assistance
- **AI assistance:** Yes
- **Used for:** Implemented the refactor transform status fix, generic audit detector and fixture coverage, and local verification; Chris remains responsible for review and merge.

### DIFF
--- a/homeboy.json
+++ b/homeboy.json
@@ -1,5 +1,21 @@
 {
   "auto_cleanup": false,
+  "audit": {
+    "command_status_contracts": {
+      "scenarios": [
+        {
+          "expected_fields": {
+            "/data/command": "refactor.transform",
+            "/data/total_replacements": 0,
+            "/data/written": false,
+            "/success": true
+          },
+          "file": "tests/fixtures/refactor_transform_no_match.json",
+          "id": "refactor-transform-no-match-dry-run"
+        }
+      ]
+    }
+  },
   "autofix_verify": {
     "args": [
       "check",
--- a/homeboy.json
+++ b/homeboy.json
@@ -14,7 +14,12 @@
           "id": "refactor-transform-no-match-dry-run"
         }
       ]
-    }
+    },
+    "convention_exception_globs": [
+      "src/core/code_audit/detectors/test_coverage.rs",
+      "src/core/code_audit/detectors/test_vacuity.rs",
+      "src/core/code_audit/detectors/upstream_workaround.rs"
+    ]
   },
   "autofix_verify": {
     "args": [
--- a/src/commands/refactor/transform_command.rs
+++ b/src/commands/refactor/transform_command.rs
@@ -89,7 +89,7 @@ fn run_transform_single(
     log_transform_rules(&result);
     log_transform_summary(&result, write);
 
-    let exit_code = if result.total_replacements == 0 { 1 } else { 0 };
+    let exit_code = 0;
     Ok((RefactorOutput::Transform { result }, exit_code))
 }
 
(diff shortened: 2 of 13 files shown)


chubes4:
Test is failing before producing structured counts. Please update the branch before merge.\n\nFailure details:\n- Selected changed-test scope includes only `tests/fixtures/refactor_transform_no_match.json`\n- Cargo reported all filtered suites ran 0 tests, then Homeboy failed with: `WARNING: cargo test ran 0 tests`\n- Final message: `Found 85 test files but no tests were executed. This may indicate a configuration issue.`\n\nThis likely needs the fixture-only changed-scope case to map to an actual regression test, or the PR needs to include/update a test file that exercises the fixture.\n\nRun: https://github.com/Extra-Chill/homeboy/actions/runs/26379892265

Issue #3401, opened 2026-06-03
audit: migrate remaining whole-tree detectors onto shared snapshots


chubes4:
## Problem

#1492 introduced shared audit analysis context for the audit/fixability/refactor path, but several audit detectors still appear to do independent whole-tree walks and file reads after discovery has already built an in-memory source snapshot and fingerprints.

Examples from current source inspection:

- `src/core/code_audit/structural.rs` calls `codebase_scan::walk_files()` and `std::fs::read_to_string()` for every source file.
- `src/core/code_audit/detectors/field_patterns.rs` walks and rereads source files.
- `src/core/code_audit/detectors/enum_dispatch_contracts.rs` walks and rereads Rust files.
- Docs drift, test topology, layer ownership, rust test wiring, and related root-level detectors also have independent walk/read paths.

## Why it matters

Audit already pays to walk and fingerprint the codebase during `discovery::auto_discover_groups()`. Repeating whole-tree scans makes `homeboy audit` scale poorly on larger components and makes the `audit-self` benchmark harder to improve. The issue is especially visible when audit is used as a frequent hot command or Lab-offloaded workload.

## Proposed direction

Migrate detectors that only need source file paths/content or existing fingerprints to consume shared audit state instead of walking/rereading the tree independently.

Potential shape:

```text
AuditAnalysisContext
  snapshot/content view
  fingerprints
  source path filters
  reference fingerprints
```

Then update detectors in slices:

1. Structural analysis from snapshot/content.
2. Field-pattern and enum-dispatch detectors from snapshot/content or fingerprints.
3. Test topology / layer ownership / rust test wiring from shared source file inventory where applicable.
4. Docs drift from a docs snapshot or a shared scan primitive if markdown content is outside source fingerprints.

## Acceptance criteria

- `homeboy audit homeboy` avoids unnecessary second whole-tree source walks for detectors that can use audit-owned shared state.
- Detector findings remain byte-for-byte equivalent before and after migration, except for intentionally documented ordering/metadata changes.
- Tests cover at least structural and one additional migrated detector using shared state.
- Per-detector phase timing from #3400 shows reduced scan/read overhead after the migration.
- The old root-only filtered path remains correct for `--only` selections that do not require discovery.

## Related

- #1492 shared scan/fingerprint/symbol context for audit/fixability/refactor and was closed after PR #2287.
- #3400 should make the remaining detector-level cost visible before/after this work.

## AI assistance

- **AI assistance:** Yes
[...]

PR #3409, opened 2026-06-03
audit: run structural detector from snapshot

chubes4:
## Summary
- Migrates the structural audit detector to analyze an already-loaded `CodebaseSnapshot` instead of walking and rereading files inside the detector.
- Keeps the detector-specific source extension coverage intact so structural findings stay unchanged.
- Adds a regression test that compares snapshot-backed structural findings against the root wrapper output.

Refs #3401.

## Tests
- `cargo test structural`
- `cargo run --bin homeboy -- audit --path . --only god_file --ignore-baseline --json-summary`

## Follow-ups
- `field_patterns` still does its own whole-tree walk/read.
- `enum_dispatch_contracts` still does its own Rust-only whole-tree walk/read.
- Other root-level detectors called out in #3401 should be migrated only when the shared state preserves their current coverage and finding payloads.

## AI assistance
- **AI assistance:** Yes
- **Used for:** Inspecting audit detector scan paths, implementing the structural snapshot-backed slice, and drafting this PR description.

### DIFF
--- a/src/core/code_audit/discovery.rs
+++ b/src/core/code_audit/discovery.rs
@@ -5,8 +5,11 @@ use std::path::Path;
 
 use super::conventions::Language;
 use super::fingerprint::{fingerprint_content, normalize_convention_tags, FileFingerprint};
-use super::walker::{is_test_path, walk_source_files_snapshot};
+use super::walker::{
+    extension_provided_file_extensions, is_extension_provided_source_file, is_test_path,
+};
 use crate::core::component::AuditConfig;
+use crate::core::engine::codebase_scan::CodebaseSnapshot;
 
 type DiscoveryGroupKey = (String, Language, bool, Vec<String>);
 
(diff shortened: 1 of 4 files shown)


chubes4:
Changes needed before merge.

This does not appear to deliver the intended performance improvement yet.

The PR says structural now analyzes an already-loaded `CodebaseSnapshot`, but `audit_internal()` calls `structural::build_snapshot(root)` immediately before `analyze_snapshot()`. That still performs a fresh structural-specific walk/read after discovery has already walked/read the source tree. In other words, the duplicated whole-tree work remains; it is just hidden behind a new snapshot wrapper.

Please either:

- Thread/reuse the existing audit discovery snapshot/shared state so structural does not do a second source-tree walk/read, or
- Narrow this PR to preparatory refactoring and make the PR title/body explicit that it does not yet reduce audit IO.

As-is, this adds API surface without improving the hot path, which is exactly the kind of performance-motivated complexity we should avoid.

chubes4:
Updated in b661f567 to address the review finding.

What changed:
- `audit_internal()` now builds one shared source snapshot with the union of extension-provided convention files and structural's detector-owned source extensions.
- Discovery now consumes that caller-provided snapshot and filters it back to extension-provided non-index files, preserving its previous selection semantics.
- Structural now consumes the same shared snapshot and filters it to its existing hardcoded source extensions, preserving structural coverage while avoiding a second full tree walk/read.
- The root-only path still uses the structural wrapper because it intentionally bypasses discovery/shared full-audit state.

Verification:
- `cargo test structural`
- `cargo test discovery`
- `cargo test core_agnostic`
- `cargo run --bin homeboy -- audit --path . --only god_file --ignore-baseline --json-summary`

PR #3416, opened 2026-06-03
audit: optimize parallel implementation scoring

### LINKED ISSUE #3403: audit: optimize parallel-implementation pairwise scoring
## Problem

The `parallel_implementation` detector appears to be the highest-risk CPU scaling path in the audit duplication family.

Current shape in `src/core/code_audit/duplication.rs`:

- `extract_call_sequences()` extracts call sequences for eligible methods.
- `detect_parallel_implementations()` loops over every pair of eligible sequences.
- For each pair it builds filtered signal vectors, checks body-shape compatibility, computes Jaccard similarity, and computes LCS ratio.
- LCS allocates a `Vec<Vec<usize>>` dynamic-programming matrix per compared pair.

That is useful signal, but the all-pairs scoring path can become expensive as method count grows.

## Why it matters

Audit should remain cookable on larger components and should be safe to run frequently. Parallel implementation detection is valuable, but it needs cheap prefilters and reusable derived state so the detector scales with real signal rather than every eligible method pair.

## Proposed direction

Keep the detector semantics but reduce pairwise work.

Possible optimizations:

- Precompute `signal_calls` once per method sequence instead of rebuilding it inside every pair.
- Precompute call sets for Jaccard instead of rebuilding `HashSet`s per pair.
- Add an inverted index from call name to method ids and only compare methods sharing enough signal calls.
- Skip LCS unless Jaccard and shared-call count already clear thresholds.
- Replace full LCS matrix allocation with a lower-allocation implementation or bounded early-exit scoring.
- Use per-detector timing from #3400 to validate the gain.

## Acceptance criteria

- Existing `parallel_implementation` findings remain equivalent for current Homeboy fixtures/tests unless an intentional false-positive reduction is documented.
- The detector performs less pairwise work on `homeboy audit homeboy`, demonstrated by timing or a focused unit/bench assertion.
- Per-detector timing shows `detector.duplication.parallel_implementation` separately once #3400 lands.
- The implementation preserves deterministic output ordering.
- Add focused tests around any new indexing/prefilter behavior.

## Related

Prior issues tuned false positives in this detector, but this issue tracks CPU cost/scaling rather than finding quality alone.

- #2333
- #2334
- #2398
- #1728
- #3400

## AI assistance

- **AI assistance:** Yes
- **Used for:** Code inspection of `detect_parallel_implementations()` and drafting this performance-focused follow-up.

chubes4:
## Summary
- Precompute filtered signal call sequences and unique signal sets once per method before pair scoring.
- Preserve the existing all-pairs ordering and finding semantics while skipping LCS for pairs that fail the already-required shared-call floor.
- Replace the per-pair full LCS matrix with a rolling-row implementation to reduce allocation pressure.

Fixes #3403.

## Verification
- `cargo fmt --check`
- `cargo test core::code_audit::duplication::tests::parallel --lib`
- `cargo test core::code_audit::duplication --lib`
- `cargo run --bin homeboy -- audit --force-hot --only parallel-implementation homeboy`

Full `cargo run --bin homeboy -- audit homeboy` smoke was attempted, but the lab runner health check failed, the local machine was hot, and the command hit the 20-minute tool timeout before completion.

## AI assistance
- **AI assistance:** Yes
- **Used for:** Inspecting the parallel-implementation scoring path, drafting the signal-preserving optimization, updating focused tests, and running verification.

### DIFF
--- a/src/core/code_audit/duplication.rs
+++ b/src/core/code_audit/duplication.rs
@@ -698,6 +698,18 @@ struct MethodCallSequence {
     shape: BodyShape,
 }
 
+/// Per-method call sequence after detector-level signal filtering.
+#[derive(Debug)]
+struct ScoredCallSequence {
+    file: String,
+    method: String,
+    /// Ordered signal calls used for LCS scoring.
+    signal_calls: Vec<String>,
+    /// Unique signal calls used for Jaccard scoring and cheap prefilters.
+    signal_set: HashSet<String>,
+    shape: BodyShape,
+}
+
 /// Generic looping markers — substrings that indicate the body iterates over
 /// something. Covers control-flow keywords (`for`, `while`, `loop`,
 /// `foreach`) shared by most languages and common iterator-pipeline calls
@@ -867,6 +879,31 @@ fn signal_calls(
         .collect()
 }
 
+fn scored_call_sequences(
+    sequences: Vec<MethodCallSequence>,
+    extra_plumbing: &HashSet<&str>,
+    common_calls: &HashSet<String>,
+) -> Vec<ScoredCallSequence> {
+    sequences
+        .into_iter()
+        .filter_map(|sequence| {
+            let signal_calls = signal_calls(&sequence.calls, extra_plumbing, common_calls);
+            if signal_calls.len() < MIN_CALL_COUNT {
+                return None;
+            }
+
+            let signal_set = signal_calls.iter().cloned().collect();
+            Some(ScoredCallSequence {
+                file: sequence.file,
+                method: sequence.method,
+                signal_calls,
+                signal_set,
+                shape: sequence.shape,
+            })
+        })
+        .collect()
+}
+
 /// Check if a name is a language keyword (not a function call).
 fn is_keyword(name: &str) -> bool {
     matches!(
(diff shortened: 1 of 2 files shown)


chubes4:
Changes needed before merge: CI Audit failed.

Failure:

`vacuous_test`: `shared_signal_call_count_counts_unique_overlap` exercises only stdlib/fixture behavior.

Please replace that helper-only test with a product-behavior test that proves the optimization preserves detector behavior. For example, construct fingerprints/call sequences that would only produce a `parallel_implementation` finding when the shared-call floor is met, and assert the detector output rather than testing `HashSet` overlap directly. The implementation itself looks directionally fine, but the audit gate is correct here: the added coverage should prove audit signal preservation, not just the helper math.

chubes4:
Still blocked: the replacement run fixed the previous helper-only test, but Audit now fails on another helper-only test.

Failure:

`vacuous_test`: `jaccard_identical_sets` exercises only stdlib/fixture behavior.

Please remove or replace the remaining direct helper math tests with detector-level behavior assertions. The test suite should prove `detect_parallel_implementations` emits/suppresses findings for the optimized scoring behavior, not assert set/Jaccard arithmetic in isolation.

PR #3553, opened 2026-06-05
Harden parallel test observation isolation

### LINKED ISSUE #3545: Harden parallel test isolation for observation store and runtime temp dirs
## Problem

While fixing `homeboy test` on latest `main`, multiple failures pointed at shared global state leaking across parallel tests:

- `writes_summary_to_run_dir_artifact` read an unexpected/missing resource summary from a run dir.
- `export_trace_spans_when_present` hit an observation-store migration race around `artifacts.artifact_type`.
- `export_since_writes_multiple_runs` observed extra recent runs from shared state and failed an exact export-count assertion.

The immediate PR can patch the observed cases, but the broader gap is that parallel tests can still collide through Homeboy's observation DB, runtime tmpdir, and environment overrides.

## Expected behavior

Tests that touch Homeboy home/runtime state should be isolated by default or use a single reliable guard/helper that prevents cross-test contamination.

## Acceptance criteria

- Test helpers make it hard to create shared observation stores or runtime temp dirs accidentally.
- Observation-store migrations are idempotent under interrupted/concurrent test setup.
- Tests that export/list recent runs avoid relying on globally empty state unless they create an isolated home.
- The full `homeboy test` path can run repeatedly without order-dependent failures.

chubes4:
## Summary
- Isolate Homeboy runtime temp dirs in the shared `with_isolated_home` test helper so run-dir artifacts do not collide with shared runtime state.
- Make observation-store migrations safer under interrupted or concurrent initialization by adding a SQLite busy timeout and idempotent handling for the `artifacts.artifact_type` migration.
- Add regression coverage for interrupted migration recovery, concurrent observation-store initialization, and run-dir resource artifact isolation.

Closes #3545.

## Verification
- `cargo fmt`
- `cargo test initialization_ -- --nocapture`
- `cargo test core::engine::resource::tests -- --nocapture`
- `cargo test export_ -- --nocapture`
- `cargo test` passed unit coverage, then failed in existing integration baseline drift: `tests/core_agnostic_test.rs::core_owned_source_stays_language_and_framework_agnostic` reports pre-existing core-boundary baseline findings in unrelated files.
- `./target/debug/homeboy test homeboy --path .` failed on the same existing `core_owned_source_stays_language_and_framework_agnostic` baseline drift after 3938 passed / 1 failed / 18 skipped.

## AI assistance
- **AI assistance:** Yes
- **Used for:** Drafted the isolation and migration hardening patch, added regression tests, and ran local verification. Chris remains responsible for review and merge.

### DIFF
--- a/src/core/engine/resource.rs
+++ b/src/core/engine/resource.rs
@@ -289,6 +289,7 @@ fn homeboy_rss_bytes(warnings: &mut Vec<String>) -> Option<u64> {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::test_support::with_isolated_home;
 
     #[derive(Debug)]
     struct FakeProbe {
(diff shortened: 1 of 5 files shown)


chubes4:
Pushed a follow-up for the failing Test job.

Failure was `core::observation::store::store_test::initialization_is_safe_under_concurrent_setup` hitting SQLite `database is locked` while multiple threads entered migration 1 concurrently:

https://github.com/Extra-Chill/homeboy/actions/runs/27022146749/job/79753668345

Fix: serialize in-process observation-store migration application with a static mutex, while keeping the existing SQLite busy timeout for connection-level contention.

Local verification:
- `cargo test core::observation::store::store_test::initialization_is_safe_under_concurrent_setup -- --nocapture`
- 5x stress loop of the same targeted test passed.

PR #4150, opened 2026-06-12
Stop duplicate Workflow Bench lab attempts after publish

### LINKED ISSUE #4146: Stop or expose bench retry loops after a result has already published
## Problem
Long Lab Workflow Bench runs appeared to publish a completed result, then continue with another attempt/retry process for the same run id until manually stopped.

Observed in Studio Web runs (`r10` through `r12`):
- `homeboy-summary.json` showed `result_counts.passed=1` and `status=completed`.
- `published/manifest.json` existed.
- Process checks still showed another `scripts/workflow-bench.mjs run --run-id ...` and `wp-codebox recipe-run ...` for the same run id.
- Manual `pkill` cleanup was needed to avoid stale retries consuming Lab resources.

## Why this hurts
After a successful published result, operators should not have to inspect process trees to know whether the run is actually done. If retries are intentional, their policy and state should be visible.

## Desired behavior
- If a pass is accepted/published, the bench wrapper should stop retrying.
- If retries continue for a reason, the summary/status should expose the retry policy, active attempt, and cancellation command.
- Controller timeout should not accidentally leave duplicate attempts running after a successful publication.

## Acceptance criteria
- A passed Workflow Bench run leaves no same-run-id retry process running.
- Active retries are visible through Homeboy status/run surfaces.
- A regression test covers pass-after-retry and verifies cleanup.

chubes4:
## Summary
- Adds a Lab pre-exec guard for Workflow Bench commands with `--run-id` so a run that already has `homeboy-summary.json` with a passing result and `published/manifest.json` is treated as terminal.
- Returns a structured `homeboy/lab-workflow-bench-published-guard/v1` envelope exposing the retry policy and marks the Lab plan with `lab.workflow_bench_published_guard` instead of launching another remote process.
- Covers pass-after-publish and failed/unpublished cases with regression tests.

Fixes #4146.

## Tests
- `cargo test -p homeboy core::runner::lab::tests::workflow_bench_published_guard -- --nocapture`
- `cargo test -p homeboy core::runner::lab::tests:: -- --nocapture`
- `git diff --check`

## AI assistance
- **AI assistance:** Yes
- **Used for:** Implemented the Lab Workflow Bench published-result guard and drafted regression tests; Chris remains responsible for review and merge.

### DIFF
--- a/src/core/runner/lab.rs
+++ b/src/core/runner/lab.rs
@@ -1,5 +1,5 @@
 use std::fs;
-use std::path::Path;
+use std::path::{Path, PathBuf};
 use std::process::Command;
 
 use crate::core::agent_task_lifecycle;
@@ -88,6 +88,16 @@ pub enum LabOffloadOutcome {
         stderr: String,
         exit_code: i32,
     },
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct PublishedWorkflowBenchRun {
+    run_id: String,
+    run_dir: PathBuf,
+    summary_path: PathBuf,
+    manifest_path: PathBuf,
+    passed_count: u64,
+    status: Option<String>,
 }
 
 fn lab_offload_git_fetch_refs(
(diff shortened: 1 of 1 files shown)


chubes4:
Updated in `3f302dae` to address the core-genericity review.

What changed:
- Replaced the Workflow Bench-specific guard with a generic `lab.run_idempotency_guard` for any Lab command carrying `--run-id`.
- Removed Workflow Bench command/path detection from Homeboy core.
- Terminal evidence is now generic: `homeboy-summary.json` with a passed/completed result plus `published/manifest.json`.
- Explicit output/status paths are resolved generically, including relative `--output-dir`, `--output-root`, `--artifact-dir`, `--artifacts-dir`, and `--status-file` values relative to the source checkout.
- Summaries without `run_id` are accepted only when the evidence directory itself is scoped by the requested run id, avoiding root-level false positives.
- Added a bounded source-tree scan for run-id-scoped `homeboy-summary.json` files so producers do not need Workflow Bench-specific paths in core.

Tests run:
- `cargo test -p homeboy core::runner::lab::tests::lab_run_idempotency_guard -- --nocapture`
- `cargo test -p homeboy core::runner::lab::tests:: -- --nocapture`
- `git diff --check`

PR #4153, opened 2026-06-12
fix: support secret-safe runner env

### LINKED ISSUE #4147: Support secret-safe runner environment configuration
## Problem
Runner environment config can contain plaintext secrets alongside ordinary settings. During debugging, reading the runner config exposed an `OPENAI_API_KEY` value in tool output.

Observed with `homeboy-lab` local config:
- runner env included normal paths and `HOMEBOY_PUBLIC_ARTIFACT_BASE_URL`
- the same env object also included `OPENAI_API_KEY`
- inspecting config to debug runner behavior risked leaking secrets

## Why this hurts
Lab/run debugging often requires checking effective env. If secrets live in the same printable JSON config, normal diagnostics become risky and easy to leak into chat/logs.

## Desired behavior
Homeboy should support and encourage secret-safe runner env configuration:
- secret env refs instead of plaintext values
- redaction when printing configs/status/errors
- an `env doctor` or `runner env` view that redacts known secret names

## Acceptance criteria
- Keys matching secret patterns are redacted in Homeboy-rendered config/status output.
- Runner config supports references to secret stores/files without embedding raw values.
- Documentation shows how to configure runner secrets separately from non-secret env.

chubes4:
## Summary
- Add `secret_env` refs to runner/server-runner config so credentials can be sourced from runner-side environment variables or files at execution time instead of stored as plaintext config.
- Make runner and server config output redact only sensitive env keys while leaving non-secret diagnostics like `HOMEBOY_PUBLIC_ARTIFACT_BASE_URL` visible.
- Document the `env` versus `secret_env` split with dummy examples.

Closes #4147.

## Tests
- `cargo test redacts --no-fail-fast`
- `cargo test secret_env --no-fail-fast`
- `cargo test --no-fail-fast` *(fails only in existing `tests/core_agnostic_test.rs::core_owned_source_stays_language_and_framework_agnostic`, which reports unrelated pre-existing ecosystem boundary findings in files not touched by this PR)*

## AI assistance
- **AI assistance:** Yes
- **Used for:** Drafted and implemented the runner env redaction/secret-reference changes, documentation, tests, and local verification under Chris's direction.

### DIFF
--- a/docs/commands/runner.md
+++ b/docs/commands/runner.md
@@ -8,6 +8,12 @@ homeboy runner <COMMAND>
 
 `runner` manages durable execution backends. SSH runners are a capability on a `homeboy server` record, so the common Lab flow uses one ID for the machine and its runner. Local runners remain standalone because they describe this machine rather than an SSH server.
 
+Runner configuration separates printable environment from secrets:
+
+- `env` is for non-secret values that are useful in diagnostics, such as `HOMEBOY_PUBLIC_ARTIFACT_BASE_URL`.
+- `secret_env` is for execution-time secret references like `{ "env": "NAME" }` or `{ "file": "~/.config/homeboy/secrets/name" }`.
+- Command output redacts sensitive names in `env` and prints only `secret_env` references, never resolved secret values.
+
 ## Subcommands
 
 ### `add`
--- a/docs/commands/runner.md
+++ b/docs/commands/runner.md
@@ -346,10 +346,10 @@ Path rules:
 
 Runner job environment:
 
-- `homeboy runner env <runner-id>` shows the effective environment Homeboy injects into runner jobs before command execution.
-- Values are redacted by default because runner env commonly contains tokens. Use `--show-values` only in trusted local/operator contexts.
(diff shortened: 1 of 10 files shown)


chubes4:
Review follow-up pushed after rebasing onto current `main` (`a544b518`, includes #4151):

- `homeboy runner env` now reports configured public `env` separately from `secret_env` refs.
- `secret_env` output includes keys and reference metadata only; resolved secret values are not printed, even with `--show-values`.
- SSH runner process prep no longer resolves `secret_env` on the controller before daemon/reverse submission.
- Added regression coverage that controller-side SSH prep leaves secret refs unresolved, while daemon-local prep resolves the same refs on the runner side.

Verification:
- `cargo test redacts --no-fail-fast` passed.
- `cargo test secret_env --no-fail-fast` passed.
- `cargo test runner_env_output --no-fail-fast` passed.
- `cargo test --no-fail-fast` still has the unrelated existing `tests/core_agnostic_test.rs::core_owned_source_stays_language_and_framework_agnostic` guard failure; the reported findings are in files not touched by this PR.

PR #4689, opened 2026-06-16
fix: clarify rig install reinstall flags

### LINKED ISSUE #4679: rig install force flag ergonomics are easy to misuse
## Problem
`homeboy rig install` rejects `--force` and suggests `--force-hot`:

`error: unexpected argument --force found`

`tip: a similar argument exists: --force-hot`

For install/reinstall semantics, `--force-hot` reads like a resource-policy flag rather than an overwrite/reinstall flag. This made it easy to issue the wrong command during rig validation.

## Expected behavior
The command should make reinstall/overwrite semantics clear and separate from resource policy semantics.

## Possible fixes
- Support `--force` as an alias when the intent is reinstall/overwrite.
- Or add a clearer flag like `--reinstall` / `--replace` and reserve `--force-hot` for hot-resource policy.
- Improve the error suggestion to explain what `--force-hot` does and whether it is safe for reinstall.

## Acceptance criteria
- `rig install --help` clearly documents reinstall/overwrite behavior.
- Invalid `--force` usage points to the correct semantic flag.
- Tests cover the accepted reinstall flag and error message.

chubes4:
## Summary
- Add `homeboy rig install --reinstall` to make reinstall/refresh intent explicit.
- Accept `--force` as a reinstall alias so users reaching for overwrite wording no longer get pointed at `--force-hot`.
- Document that reinstall refreshes matching installs while resource policy remains controlled by `--force-hot`.

Fixes #4679.

## Tests
- `cargo test --test command_surface_test rig_install -- --nocapture`
- `cargo test install_refreshes_existing_matching_local_rig_without_metadata -- --nocapture`
- `cargo test install_rejects_existing_rig_with_different_declared_id -- --nocapture`
- `cargo test install_rejects_existing_stack_with_different_content -- --nocapture`

Note: one parallel rerun of the stack conflict test was SIGKILLed while competing for Cargo's artifact lock; the same test passed when rerun sequentially.

### DIFF
--- a/docs/commands/rig.md
+++ b/docs/commands/rig.md
@@ -128,10 +128,13 @@ Reports service PIDs, stale service state, and last recorded `up` / `check` time
 ```sh
 homeboy rig install https://github.com/chubes4/homeboy-rigs.git//packages/studio --id studio
 homeboy rig install ./packages/studio
+homeboy rig install ./packages/studio --reinstall
 homeboy rig install https://github.com/chubes4/homeboy-rigs.git//packages --all
 ```
 
 Installs rigs from a local directory or git-backed package. Package discovery accepts either a single `rig.json` or a package layout with `rigs/<id>/rig.json`. If the selected package also contains `stacks/*.json`, those stack specs are installed alongside the rig. Existing stack specs with different content are treated as user-owned config and are left in place instead of being overwritten.
+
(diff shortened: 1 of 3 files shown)


chubes4:
Minion ping: please continue/finish this one. Important constraint from Chris: do not run cargo commands; they freeze up RAM on this host. Use targeted non-cargo inspection/tests only, and report if verification is blocked by the no-cargo constraint.

Issue #4659, opened 2026-06-15
Preview client hangs forwarding reviewer auth bootstrap responses


chubes4:
## Summary
When a held `wp-codebox` preview exposes a reviewer auth bootstrap route through Homeboy preview ingress, the bootstrap endpoint works locally on the runner but hangs through the public tunnel.

## Evidence
- Broker auth is fixed and preview-client registration succeeds.
- Public route reports ready: `ready https://wpcom-ai-edit-tunnel.dev.chubes.net`.
- Runner-local bootstrap succeeds:
  - `http://127.0.0.1:48631/__wp-codebox/reviewer-auth-bootstrap?token=...`
  - Returns `302` to `/ai/`
  - Sets 3 auth cookies
- Public-translated bootstrap hangs:
  - `https://wpcom-ai-edit-tunnel.dev.chubes.net/__wp-codebox/reviewer-auth-bootstrap?token=...`
  - Client read times out before headers/body are returned.
- Public `/ai/` can return `200`, so the route itself is registered.

## Reproduction Shape
1. Start a runner-owned preview client against `https://preview-broker.dev.chubes.net`.
2. Local origin is a held `wp-codebox` preview on `127.0.0.1:48631`.
3. Hit a normal public page path like `/ai/`: works intermittently/slowly.
4. Hit `__wp-codebox/reviewer-auth-bootstrap?token=...`: hangs through ingress, while local origin responds immediately.

## Expected
Preview ingress should forward the bootstrap response, including `302 Location` and multiple `Set-Cookie` headers, without hanging.

## Notes
Do not log or expose bootstrap tokens or cookie values. Only validate status, redirect target, and cookie count.

chubes4:
Fixed in https://github.com/Extra-Chill/homeboy/pull/4665. Evidence: added local-origin coverage for a bootstrap-shaped 302 with multiple Set-Cookie headers; added reverse-channel ingress coverage that forwards 302 Found, Location, two separate set-cookie headers, and content-length: 0 for /__wp-codebox/reviewer-auth-bootstrap?token=... using fake token/cookie values only. Targeted validation: cargo fmt; cargo test preview (46 passed). Broader cargo test --lib still has 5 unrelated/environment failures outside preview forwarding, noted on the PR.

chubes4:
[...]

PR #4716, opened 2026-06-16
test: harden preview bootstrap forwarding regression

chubes4:
## Summary
- Confirms the merged preview bootstrap forwarding fix already preserves `302 Found`, `Location`, and repeated `Set-Cookie` headers through public ingress.
- Hardens the reverse-channel regression test to assert exact status/location and cookie count without logging cookie values.
- Adds a bounded raw HTTP read timeout so a missing response terminator/content length fails deterministically instead of hanging.

## Verification
- `cargo fmt --check`
- `cargo test preview` (46 passed)

Refs #4659.

### DIFF
--- a/src/core/preview_ingress_tests.rs
+++ b/src/core/preview_ingress_tests.rs
@@ -206,6 +206,9 @@ fn reverse_channel_client_serves_public_request() {
 fn reverse_channel_client_forwards_bootstrap_redirect_and_repeated_cookies() {
     test_support::with_isolated_home(|_| {
         let token = "test-preview-token";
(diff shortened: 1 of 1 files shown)


chubes4:
Rebased onto current `origin/main` and resolved the test-helper conflict by keeping both the new byte-response helper from `main` and this PR's header-value assertions. PR is now mergeable.

Validation after rebase:
- `cargo fmt --check`
- `cargo test preview` (48 passed)

No token or cookie values were logged; the regression asserts only status line, location header value, repeated cookie count, and `content-length: 0`.

PR #6358, opened 2026-06-24
fix: restore 10 RED-main tests (post-merge-burst drift)

chubes4:
## Summary

Main went RED (5960 passed; 10 failed) after a high merge volume, blocking the release pipeline. This restores all 10 failing tests, fixing the **source of drift** wherever possible (only updating a stale assertion when the new behavior was clearly the intended one).

## Fixes by cluster

**A — component `local_path` error-hint wording (3 tests, source fix)**
`validate_component_local_paths` / `validate_component_local_path` packed the missing-path blockers only into `details.tried`, never as `hints`, but the tests (added in `cc91c81a`) assert the blocker appears in `err.hints`. Now the blockers are attached as hints (still also in `tried`). Covers deploy `run`, resolution `resolve_project_component`, and readiness validation.

**B — agent-task controller (3 tests, source fix + 1 stale-assertion update)**
- `compile_plan_from_spec_rejects_unbacked_artifact_consumption`: `validate_artifact_flow_bindings` treated a producer merely declaring an artifact in its `artifacts` set as backing a consume edge, so unbacked consumption compiled. Backing now requires an explicit `emits` or an `artifact_flow` edge (matches the sibling `..._accepts_emit_consume_pairing` test which sets `emits`).
- `init_from_spec_for_resume_fork_isolates_under_derived_loop_id`: `derive_fork_loop_id` produced `…/fork-<hash>`, but the loop_id becomes a single sanitized path segment (`sanitize_path_segment` collapses `/`→`_`), so the persisted `report.loop_id` diverged. Switched the derived id to the sanitization-stable `…-fork-<hash>` and updated the assertion to match.
- `run_failure_summary_normalizes_nested_provider_failure`: two drifts — owner-surface classification ignored the `provider`/`failure_phase` codebox signal (fell through to `wordpress_runtime`), and declared artifacts carried their intrinsic `kind` (`log_bundle`) instead of the evidence-ref taxonomy. Codebox classification now considers provider/phase, and declared-artifact refs are classified by container as `artifact_bundle`.

**C — route Lab routing (2 tests, test-isolation fix)**
Both failed due to the process-global `LAB_OFFLOAD_METADATA_ENV` leaking from a parallel test (its presence makes `route_after_parse` short-circuit as a Lab-offload subprocess → `Ok(None)`). Added the same `EnvGuard::remove(LAB_OFFLOAD_METADATA_ENV)` scoping guard the sibling tests already use, so the preconditions hold regardless of test ordering.

**D — cli_surface manifest (1 test, source fix)**
`command_registry_manifest_and_docs_metadata_align` enforces that the registry is authoritative for `output_notes`, but `da14a66d` added single-segment `output_notes` overrides in `command_safety_metadata` for `trace`/`lint`/`cleanup`/`refactor`/`undo` without syncing `COMMAND_REGISTRY`. Moved those strings into dedicated registry constructors and dropped the per-path `output_notes` overrides (keeping path-specific `mutates`/`dangerous_flags`).

**E — process_tree SIGKILL escalation (1 test, source fix)**
`terminate_process_tree` computed `surviving_pids` immediately after sending SIGKILL, but kill delivery/reaping is asynchronous, so a freshly-killed child could still report running → racy `surviving_pids`. Now polls `wait_for_exit` on the killed set (`SIGKILL_GRACE`) before computing survivors, making escalation deterministic (assertion not weakened).

## Verification
- `cargo build --lib` clean.
- `cargo fmt` (out-of-scope reformatting reverted; only the edited closure kept).
- Each fixed test verified via targeted `cargo test --lib '<path>' -- --exact`; full `cli_surface::tests` (17) and `process_tree_tests` (4) modules pass.

## Notes
- All fixes are at the source of drift except the resume-fork loop_id prefix (intended-behavior assertion update) and the two route env-isolation guards (test-hardening). No new ecosystem literals were introduced into agnostic core behavioral code.

### DIFF
--- a/src/cli_surface.rs
+++ b/src/cli_surface.rs
@@ -995,8 +995,9 @@ fn command_safety_metadata(path: &[String]) -> CommandSafetyMetadata {
                 "mutates persisted audit baseline data in component configuration";
         }
         ["refactor"] => {
+            // output_notes is sourced from the authoritative command registry
+            // (refactor_command_registry_entry).
             metadata.mutates = true;
-            metadata.output_notes = "refactor subcommands can rewrite source files; use planning/dry-run modes where available";
             metadata.dangerous_flags = vec!["--write", "--commit"];
         }
         ["refactor", "rename"]
--- a/src/cli_surface.rs
+++ b/src/cli_surface.rs
@@ -742,12 +742,14 @@ fn command_safety_metadata(path: &[String]) -> CommandSafetyMetadata {
             metadata.dangerous_flags = vec!["--force", "--upgrade-runner"];
         }
         ["trace"] => {
+            // output_notes is sourced from the authoritative command registry
+            // (trace_command_registry_entry); only the path-specific mutates
+            // flag is set here.
             metadata.mutates = true;
-            metadata.output_notes = "runs trace workflows and records observation artifacts unless using read-only subcommands";
         }
         ["lint"] => {
-            metadata.output_notes =
-                "runs lint workflows; pass --fix to apply auto-fixable findings in place";
+            // output_notes is sourced from the authoritative command registry
+            // (lint_command_registry_entry).
             metadata.dangerous_flags = vec!["--fix", "--force"];
         }
         ["deps", "install"] | ["deps", "update"] | ["deps", "stack", "apply"] => {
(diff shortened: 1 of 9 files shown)


chubes4:
Closing as redundant/superseded. This branch's RED-main fixes already landed on main via #6355 (clusters A/B/C — component-path hints, agent-task controller, route Lab guards) and #6357 (cli_surface trace manifest + process_tree SIGKILL). The original cook on this branch went silent mid-run, so the work was salvaged into fresh branches and merged; all 10 RED-main tests are green on main and v0.260.0 has been released. This PR is now DIRTY (conflicts) because its content is fully duplicated on main.

PR #6371, opened 2026-06-25
feat(runner): domain-specific run names for runner exec jobs (#6362)

### LINKED ISSUE #6362: Use domain-specific run names for runner exec jobs
## Problem
Some generic runner exec helper commands were persisted with unrelated run names such as `woo-db-api-rest-query-profile-...` while inspecting SSI fixture matrix artifacts. That makes evidence harder to trust and search.

## Desired behavior
Runner exec jobs should derive run ids/titles from the command domain or explicit labels instead of falling back to stale/unrelated workload naming.

## Acceptance criteria
- `homeboy runner exec` records a run name that reflects the command being executed when no explicit name is passed.
- Callers can pass an explicit run label/id for ad hoc evidence commands.
- Existing workload-specific naming is not reused outside that workload family.
- `homeboy runs list` makes it obvious which runs are matrix executions, summary extraction, or generic diagnostics.

chubes4:
Closes #6362. `runner exec` now derives a run id/title from the command being executed when no explicit name is passed (instead of inheriting a stale/unrelated workload name), and honors an explicit run label for ad hoc evidence commands. Workload-specific naming is no longer reused outside its workload family. Kept generic/agnostic (names derived from command data).

## What changed
- Added `derive_runner_exec_run_label(command, explicit)` in `evidence/mirror.rs`: explicit label wins (sanitized to a lowercase slug); otherwise the label is derived from the command's first meaningful (non flag-looking) token, reduced to its basename + stem. Falls back to a generic `exec` token only when no usable token exists. No ecosystem literals — derivation reads the command argv as data.
- Persisted runner-exec runs now record `metadata_json.run_label`, so `homeboy runs list` can distinguish matrix executions, summary extraction, and generic diagnostics.
- Added a `--run-label` flag to `homeboy runner exec`, threaded through `RunnerExecOptions.run_label` -> daemon/reverse-broker transports -> `persist_lab_offload_handoff_run` -> `mirror_job_run`.
- Unit tests covering: derivation from command domain, executable basename/extension stripping, skipping leading flags, explicit-label sanitization, blank-label fallback, generic fallback for all-flags commands, and that an unrelated prior workload name is never reused.

## Verification
- `cargo build --lib` (clean)
- `cargo fmt` (in-scope files only; out-of-scope drift reverted)


chubes4:
Closing: superseded/out-of-scope. The runner-exec domain-naming work (#6362) was deprioritized; this auto-opened branch is DIRTY.

PR #6653, opened 2026-06-27
chore: reduce trace/report.rs HighItemCount via nested mods

chubes4:
Groups pub structs + fns into nested mod blocks with pub use re-exports to bring src/core/extension/trace/report.rs under the 30-item threshold. Zero API change (pub use preserves paths + field access; private fns made pub(crate) where re-exported). serde attrs intact. Verified cargo build --lib --tests (exit 0). field_patterns row left in place.


chubes4:
Superseded \u2014 content landed via the -v2 branch (#6656). Closing stale original.

Issue #6855, opened 2026-06-28
core: extract hardcoded language/grammar knowledge (Rust/PHP/JS) out of core into extensions


chubes4:
## Summary

The CoreBoundaryLeak detector (`core-agnostic-source`) currently has **275 baselined findings** — ~90% of the entire audit baseline. After a full session working through the baseline, the conclusion is: **these are true positives, not detector noise.** Homeboy core genuinely should not know what Rust or PHP are, and right now it does, in several engines that do language-specific parsing and code generation.

The fix is **not** to exclude these paths from the detector or annotate them away (that would hide real debt and weaken the gate). The fix is to **move the language-aware logic into extensions** and have core drive it through generic, config/manifest-supplied primitives — exactly the pattern already used by the vacuity detector (`TestVacuityPolicy` / `PackageNameSource`) and the thin-command-adapter detector (all markers config-driven).

## Concrete violations (representative, not exhaustive)

- **`src/core/code_audit/structural.rs`** — hardcoded extension list `["rs","php","js",...]` and a literal dispatch: `"rs" => count_rust_items(content)`, `"php" => count_php_items(content)`, with `count_rust_items` / `count_php_items` / `is_rust_item_declaration` etc. living in core.
- **`src/core/refactor/plan/generate/signatures.rs`** — a `Language::{Rust,Php}` enum baked into core that emits language-specific stubs, e.g. `Language::Rust => "todo!(...)"`, `Language::Php => ...`.
- **`src/core/engine/symbol_graph.rs`** — `parse_rust_import_path`, language-specific import parsing / module-path derivation.
- **`src/core/engine/edit_op_apply.rs`** — Rust/PHP-specific type-declaration and namespace/anchor handling (`try_replace_namespace`, `is_type_declaration_line`).
- Term distribution across the 275: `php` (112), `rust` (97), `cargo` (17), `npm` (12), `composer` (12), `wordpress` (7), `wp-content` (4), `WooCommerce` (2).
- Clustering by dir: `refactor` (94), `code_audit` (48), `engine` (38), `extension` (30), `runner` (25), `release` (18), `setup` (10).

## Why this matters

These engines (refactor planning, structural audit, symbol graph, edit application) are supposed to be **language-agnostic orchestration**. Today they encode grammar and codegen for specific languages, which:
- makes adding a new language a core change instead of an extension change,
- couples core releases to ecosystem details,
- and produces 275 standing audit findings that train everyone to treat the gate as noise.

[...]

Issue #6808, opened 2026-06-27
Audit gate: baseline drift on rebased PRs causes false failures


chubes4:
**Observed:** PR #6782 passed audit when cut, but after rebasing onto an advanced main (other merges had moved the audit baseline in homeboy.json), the audit gate failed with ~300 findings that are mostly pre-existing/baseline, not introduced by the PR. The branch's baseline snapshot went stale relative to main.

**Fix directions:**
- Regenerate/merge the audit baseline at merge time (or compare against main's current baseline) rather than the snapshot pinned at branch-cut.
- Make `--changed-since` audit scope authoritative for the gate so only the PR's actual delta is judged.

Surfaced during a multi-PR cleanup fleet (rebasing PRs onto a fast-moving main).

PR #6896, opened 2026-06-28
feat(audit): extract item-counting grammar to config (#6855 phase 1)

chubes4:
Phase 1 of #6855. Replaces hardcoded count_rust_items/count_php_items/count_js_items/is_rust_item_declaration in structural.rs with a generic config-driven count_items primitive reading a new LanguageGrammar contract (item_declaration_prefixes / visibility_prefixes / modifier_prefixes / ignore_after_line_equals). Rust/PHP/JS item keywords move OUT of core into homeboy.json audit config. GodFile/HighItemCount counts are byte-identical (regression tests assert parity). Removes language literals from structural.rs. Verified cargo build --lib --tests + cargo test --lib structural unchanged.

### DIFF
--- a/homeboy.json
+++ b/homeboy.json
@@ -125,6 +125,81 @@
       "src/core/code_audit/detectors/test_vacuity.rs",
       "src/core/code_audit/detectors/upstream_workaround.rs"
     ],
+    "language_grammars": [
+      {
+        "file_extensions": [
+          "rs"
+        ],
+        "ignore_after_line_equals": [
+          "#[cfg(test)]"
+        ],
+        "item_declaration_prefixes": [
+          "fn ",
+          "struct ",
+          "enum ",
+          "const ",
+          "static ",
+          "type ",
+          "trait ",
+          "impl ",
+          "impl<"
+        ],
+        "modifier_prefixes": [
+          "async ",
+          "unsafe "
+        ],
+        "visibility_prefixes": [
+          "pub(crate) ",
+          "pub(super) ",
+          "pub "
+        ]
+      },
+      {
+        "file_extensions": [
+          "php"
+        ],
+        "item_declaration_prefixes": [
+          "function ",
+          "class ",
+          "interface ",
+          "trait ",
+          "const "
+        ],
+        "modifier_prefixes": [
+          "static ",
+          "abstract ",
+          "final "
+        ],
+        "visibility_prefixes": [
+          "public ",
+          "protected ",
+          "private "
+        ]
+      },
+      {
+        "file_extensions": [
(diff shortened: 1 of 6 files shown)


chubes4:
## ⚠️ Do not merge — Audit gate exposes a runtime regression

Lint ✅ and Test ✅ (the structural parity unit tests pass — `count_items` reproduces `count_rust_items`/`count_php_items`/`count_js_items` byte-for-byte). But the **Audit gate fails with `delta: +293, drift_increased: true`** — and crucially, this persists **after rebasing onto current main** (so it is NOT baseline drift / #6808).

The +293 new findings are spread across **every detector category** (158 CoreBoundaryLeak, 78 HighItemCount, 72 RepeatedFieldPattern, 28 GodFile, NamingMismatch/MissingMethod/DuplicateFunction in dozens of untouched files like `http_api.rs`, `agent_task_dispatch.rs`, `project/types/*`). That broad, cross-detector explosion is the signature of **the audit config not being fully applied at runtime** on this branch — e.g. `exclude_path_contains`/allowlists/detector configs effectively reverting to defaults, so every file (incl. generated/test) gets scanned and CoreBoundaryLeak/GodFile/HighItemCount fire everywhere.

### What I ruled out
- Not baseline drift (persists post-rebase).
- Not the new `#[serde(default)]` `language_grammars` field breaking deserialization (all fields defaulted, no `deny_unknown_fields`).
- Not the `component/mod.rs` change (pure `pub use` reorder).
- Parity is correct (unit tests pass) — so `count_items` itself is fine.

### Likely root cause (needs runtime trace)
Something in the wiring (`descriptor_runtime.rs` passing `&config.language_grammars`, or the `AuditConfig::is_empty()`/`merge` change) causes the **live audit run** to diverge from the unit-test path — most plausibly the structural/symbol-graph analysis no longer receives the loaded config the same way, collapsing item/structure understanding so cross-reference detectors mis-fire. The unit tests don't catch it because they construct config in-Rust and bypass the runtime config-load + multi-detector assembly.

### Recommendation
Hold this PR. Phase 1's migration touched a primitive shared by more than just `count_top_level_items` (symbol-graph + structural share understanding of "what's an item"), so it needs the **full audit run traced locally** (or a test that loads the real `homeboy.json` audit config and asserts the live finding count is unchanged on a fixture tree) before re-attempting. Adding such a config-round-trip regression test should be part of Phase 1 so this class of break can't pass Lint+Test again.

chubes4:

chubes4:
Closing as superseded by #6915, which is the current grammar phase 1 draft.

PR #6915, opened 2026-06-28
feat(audit): item-counting grammar to config (#6855 phase 1, harness-gated)

chubes4:
Third attempt at #6855 Phase 1, gated by the fidelity-upgraded runtime audit harness (#6914) which now exercises test-path skipping AND the agnostic detector — the two behaviors #6896/#6906 broke. Extracts structural.rs item counting into config-driven LanguageGrammar; enum Language/symbol_graph/is_empty untouched. All 3 audit_runtime_regression harness tests + structural parity pass with unchanged findings.

### DIFF
--- a/homeboy.json
+++ b/homeboy.json
@@ -125,6 +125,51 @@
       "src/core/code_audit/detectors/test_vacuity.rs",
       "src/core/code_audit/detectors/upstream_workaround.rs"
     ],
+    "language_grammars": [
+      {
+        "file_extensions": ["rs"],
+        "item_declaration_prefixes": [
+          "fn ",
+          "struct ",
+          "enum ",
+          "const ",
+          "static ",
+          "type ",
+          "trait ",
+          "impl ",
+          "impl<"
+        ],
+        "visibility_prefixes": ["pub(crate) ", "pub(super) ", "pub "],
+        "modifier_prefixes": ["async ", "unsafe "],
+        "ignore_after_line_equals": ["#[cfg(test)]"]
+      },
+      {
+        "file_extensions": ["php"],
+        "item_declaration_prefixes": [
+          "function ",
+          "class ",
+          "interface ",
+          "trait ",
+          "const "
+        ],
+        "visibility_prefixes": ["public ", "protected ", "private "],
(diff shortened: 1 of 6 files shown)


chubes4:
## Held as draft — root cause of the harness fidelity gap FOUND

Third attempt. The fidelity-upgraded harness (#6914) passes all 3 tests (matches_snapshot, **skips_test_paths**, deterministic) AND structural parity passes (24 tests). Yet CI Audit fails **identically** to #6896/#6906: `delta: 300`, 158 CoreBoundaryLeak (82 on `tests/` paths), 94 IntraMethodDuplicate, 84 HighItemCount...

### Root cause: the harness exercises the WRONG audit entry point
- The regression-harness calls **`audit_path_with_id`** (`src/core/code_audit/entry.rs`), which loads config via `discover_from_portable` and audits a path directly.
- The **CI Audit gate / `homeboy audit` CLI** (`src/commands/audit.rs::run`) calls **`run_main_audit_workflow`** with **`reference_paths`** (`resolve_audit_reference_paths`).

These are DIFFERENT code paths. `reference_paths` drive how the symbol-graph / cross-reference detectors (MissingMethod, NamingMismatch) and likely the test-path/fingerprint handling resolve against the whole codebase. The structural item-grammar migration perturbs something in the `run_main_audit_workflow` + reference-path path that `audit_path_with_id` simply doesn't exercise — which is why the toy fixture (and even the enriched one) can't reproduce it, but the real CLI audit fails every time.

### The actionable fix (prerequisite for #6855)
**The runtime audit regression harness must call `run_main_audit_workflow` (the same path CI uses), not `audit_path_with_id`.** Once it does, this exact +300 regression reproduces locally and becomes diagnosable. Until then, no `structural.rs`/symbol-graph-adjacent change can be safely validated locally.

This is the real fidelity gap (deeper than the fixture-content gap fixed in #6914). I'll redirect to upgrading the harness to use the CLI workflow entry point.

Migration itself is correct in isolation (parity + fixture harness pass); held pending the workflow-path harness.

chubes4:

Issue #6858, opened 2026-06-28
audit: provide a fast, incremental 'what-would-change' preview so contributors can validate baseline edits without a full rebuild


chubes4:
## Problem

The only way to know whether a homeboy.json baseline edit (pruning a stale/cleared row) is correct is to run the full differential Audit gate in CI. Locally, `homeboy audit` recompiles the whole binary (minutes + gigabytes of `target/`), which on a shared/constrained host is effectively unusable — during this session a worker stalled for ~15 minutes in an audit-recompile loop trying to self-verify prunes, and the practical workaround became "prune speculatively, let CI's differential Audit reject it if wrong."

That works (the differential gate is a good backstop, ref #6808/#6854), but it means every baseline edit costs a full CI round-trip, and contributors can't get fast local confidence.

## What would help

1. **A fast incremental audit mode** — `homeboy audit --changed-since <ref>` exists (#6383) but is local-only and still heavyweight. A cheap path that audits only the files in the diff and prints the **delta vs baseline** (new findings / cleared findings / net change) would let contributors confirm "this prune is genuine" in seconds.
2. **A dry-run baseline reconciler** — e.g. `homeboy audit --reconcile-baseline --dry-run` that reports which baselined rows no longer fire (safe to prune) and which proposed prunes would re-introduce findings (unsafe), without writing anything. This directly supports the burndown workflow.
3. **Make the audit binary cheaper to invoke** for this use case (precompiled/cached, or a `--lib`-only fast path), so self-verification doesn't require a from-scratch build.

## Why
Tightening the loop here makes the gate *easier to satisfy correctly* and reduces both CI churn and the temptation to guess. Pure tooling/ergonomics — no gate change.

PR #6917, opened 2026-06-28
test(audit): regression harness uses CI audit entry point (run_main_audit_workflow)

chubes4:
Fixes the deep fidelity gap behind #6896/#6906/#6915: the harness audited via audit_path_with_id while CI Audit uses run_main_audit_workflow + reference_paths — so reference-path/symbol-graph regressions could not reproduce locally. Switches the harness to run_main_audit_workflow (matching src/commands/audit.rs), re-captures the snapshot, keeps the skips_test_paths invariant + determinism test. Now a #6855-style structural/symbol-graph regression fails locally. Harness test + fixture only; no detector/config change.

### DIFF
--- a/tests/core/code_audit/audit_runtime_regression_test.rs
+++ b/tests/core/code_audit/audit_runtime_regression_test.rs
@@ -1,19 +1,46 @@
 //! Runtime audit regression harness.
 //!
-//! This test runs the **full audit pipeline** ([`audit_path_with_id`]) against a
-//! self-contained fixture component tree (`tests/fixtures/audit_runtime/`) and
-//! asserts that the produced finding set is byte-for-byte identical to a
-//! committed snapshot.
+//! This test runs the **real CI audit entry point**
+//! ([`run_main_audit_workflow`]) against a self-contained fixture component tree
+//! (`tests/fixtures/audit_runtime/`) and asserts that the produced finding set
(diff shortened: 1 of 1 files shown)


chubes4:
## Empirical finding: entry point was NOT the gap; the regression is REAL-CODEBASE-SPECIFIC

I tested this directly: cherry-picked the #6915 structural migration onto THIS branch (workflow-entry harness) and ran the harness — **it still PASSES (4 tests), even via `run_main_audit_workflow`.** So switching to the CI entry point, while strictly correct and worth keeping, does NOT reproduce the +300 regression on the synthetic fixture either.

**Conclusion:** the #6855 Phase 1 regression depends on the **actual homeboy repository's scale/content** (real cross-references, real file sizes hitting GodFile/HighItemCount thresholds, real `reference_paths`) — it cannot be reproduced by any synthetic fixture, however enriched. 

This PR is still a real improvement (the harness should audit via the same entry point CI uses — better fidelity for the class of regressions a fixture CAN catch), so it's worth merging on its own merits.

But the actual unblocker for #6855 is a different harness shape: **a test/CI step that audits the REAL homeboy repo tree and asserts zero net drift vs baseline** — essentially running the differential Audit gate locally/in-test. That's the only thing that reproduces real-codebase-specific regressions. Recommend that as the next step (it overlaps with #6858 'fast audit preview/reconcile').

Issue #6938, opened 2026-06-29
Dependency-aware release cascade: one command to release an upstream component and auto-bump+release its dependents


chubes4:
## Problem
Releasing an upstream component and propagating it to a dependent is a long, deterministic, error-prone hand-crank. Observed this session releasing `php-transformer` → `static-site-importer` (which pins it as a composer custom-package):

1. `homeboy release php-transformer` → new tag + commit SHA
2. manually rewrite the dependent's pin: composer custom-package `version`, `dist.url` (the `…/archive/refs/tags/<tag>.zip`), `source.reference`, `dist.reference`, and the `require` constraint
3. regen `composer.lock`
4. open PR, merge
5. (papercut) re-pin the component's `local_path` to absolute — it kept reverting to a relative value that can't resolve, and kept pointing at a STALE worktree instead of the primary
6. (papercut) refresh the primary, clear `vendor/` (GitDownloader/.git build issue)
7. release the dependent with a pile of flags: `--skip-checks --apply --bump patch --git-identity "…"` — `--bump patch` only because a lone dependency-bump commit is classified "nothing to release"

This should be **one command**: `homeboy release <component> --cascade`.

## Proposed design (keep core generic)
- **homeboy core — generic cascade orchestrator:** after releasing the upstream, consult the dependency graph (`homeboy deps`); for each dependent that declares a dependency on the released component, invoke a framework-specific "set dependency <id> to <version>@<sha>" hook, then release that dependent. No WP/composer specifics in core.
- **homeboy-extensions (wordpress) — composer-pin updater:** implement the hook for composer custom-packages: rewrite version/dist.url/source.reference/dist.reference/constraint + regen lock. Lives where PHP/composer knowledge belongs.

## Subsume these papercuts (root-cause, don't paper over)
- Component `local_path` stored relative → "cannot be resolved" at release. Normalize to absolute (or resolve against workspace root) on `component set`/`reconcile`.
- Component pinned to a stale worktree instead of the primary → release evaluated the wrong checkout (detached-HEAD refusal, unreachable tag). Prefer the primary / auto-point, or refuse with a clear fix.
- Pure dependency-bump release classified "nothing to release" → auto-apply a patch bump for a dependency-only change instead of requiring `--bump patch`.
- Lint baseline false-positive (#2012), vendor/.git build (#2010, done), dangling-tag recovery (#6933) — fold awareness in so `--cascade` doesn't trip on them.

## Acceptance
[...]

PR #7417, opened 2026-07-03
fix(release): normalize project attachment local_path so release and deploy agree (#7410)

### LINKED ISSUE #7410: release and deploy resolve different checkouts for the same component (two un-synced stores)
## Summary

For the same component id, `homeboy release <id>` and `homeboy deploy` (project component-attach) resolve **different git checkouts** because they read from two separate, un-synced stores. When those stores drift, a release can be cut from the wrong checkout (wrong `local_path` and/or wrong `version_targets`) with no warning, while deploy uses the correct one.

## Two stores, asymmetric resolution

- **Deploy / project scope** → `core::project::component::resolution::resolve_project_component()`:
  - Uses the project attachment's `local_path` (`~/.config/homeboy/projects/<id>/<id>.json` → `components[].local_path`).
  - Calls `discover_attached_component()`, which reads the **repo's own `homeboy.json`** at that path for `version_targets` etc.
  - The standalone `~/.config/homeboy/components/<id>.json` is consulted only as a **fallback** for `remote_path` / `extract_command` / `remote_url` (`apply_standalone_component_fallbacks`).

- **Release (positional component scope)** → `homeboy release <id>` → `core::release::context` → `component::resolve_effective(Some(id), None, None)` → falls through to `component::load(id)`:
  - Reads **only** `~/.config/homeboy/components/<id>.json` (the standalone store) for `local_path` **and** `version_targets`.
  - Never consults the project attachment or the repo's `homeboy.json`.

So `homeboy release <id>` and `homeboy deploy`/`homeboy release --project <proj>` can resolve different `local_path` values for the same component. (`resolve_component_ids` in `src/commands/release.rs` confirms: `--project` goes through project resolution, a positional id does not.)

## Concrete repro (observed)

Component `agents-api`:
- Project `studio-native-runtime` attachment: `local_path = /Users/chubes/Developer/agents-api-fanout` (branch `main`, has merged fanout work, tag `v0.5.0` reachable). Deploy is correct.
- Standalone `~/.config/homeboy/components/agents-api.json`: `local_path = /Users/chubes/Developer/agents-api-studioweb-release-main` (detached HEAD, only `v0.4.0` reachable). It ALSO carried an orphaned `version_target` (`tests/playground-ci/component/agents-api-docs-agent-driver.php`) that the repo's own `homeboy.json` had already dropped.
- Result: `homeboy release agents-api` failed with `Latest release tag v0.5.0 is not reachable from HEAD`, while `homeboy release agents-api --path <fanout>` planned `v0.5.0 → v0.5.1` cleanly.

The existence of the `homeboy release --path <PATH>` override ("Override local_path for version file lookup, single component only") is itself an acknowledgement that auto-resolution is known-unreliable.

## Why this is a design gap, not just stale config

The stale standalone config was the trigger, but the root design issue is that **release and deploy have no shared source of truth for a component's `local_path`/`version_targets`, and they never cross-check.** A component attached to a project can silently release from a different checkout than it deploys from.

## Suggested directions (for discussion)

1. When a component id resolves to exactly one project attachment, have positional `homeboy release <id>` prefer (or at least reconcile against) the project attachment's `local_path`, or warn on divergence.
2. Add a divergence check / `homeboy component reconcile` extension that flags when the standalone `local_path` differs from any project attachment's `local_path` for the same id. (`reconcile` today only detects a *missing* path, not a *wrong-but-existing* one.)
3. Prefer the repo's own `homeboy.json` `version_targets` in the release path the same way deploy already does, so orphaned/stale standalone `version_targets` can't diverge from the tracked repo config.

## Environment
- homeboy branch `ai-runtimes-provider-discovery`
- Files: `src/core/component/resolution.rs` (`resolve_effective_inner`, `load`), `src/core/project/component/resolution.rs` (`resolve_project_component`, `apply_standalone_component_fallbacks`), `src/commands/release.rs` (`resolve_component_ids`), `src/core/scope.rs`.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** Claude Opus 4.8 via Claude Code
- **Used for:** Root-causing the two-store resolver divergence and drafting this report.

chubes4:
Fixes #7410.

## Problem

`homeboy release <id>` could fail with

```
Invalid argument 'local_path': Component 'studio-native' has relative local_path 'studio-native' which cannot be resolved. Use absolute path
```

even after `homeboy component set studio-native --local-path <absolute>` reported success and wrote the absolute path to `~/.config/homeboy/components/studio-native.json`, and even though `homeboy deploy` resolved the same component to an absolute checkout. So `release`, `deploy`, and `component set` were resolving **different sources of truth** for one component's `local_path`.

## Root cause

Traced to `resolve_effective_inner` in `src/core/component/resolution.rs`. When `release <id>` runs from a directory that contains a sibling directory whose name matches the positional component id -- e.g. running `homeboy release studio-native` from `/Users/chubes/Developer`, where `studio-native/` exists -- `Path::new("studio-native").is_dir()` is true, so the id is treated as a **bare-directory target**:

```rust
if accept_bare_directory && id_path.is_dir() {
    if let Some(mut discovered) = try_discover_from_portable(id_path)? {
        discovered.local_path = id.to_string(); // <- raw RELATIVE "studio-native"
        ...
```

This short-circuits **before** the registry/inventory lookup, storing the raw relative value as `local_path`. `release`'s `validate_local_path` then rejects the relative path -- while `component set` (standalone store) and `deploy` (project attachment) both resolve the same component to an absolute checkout. That is the three-way divergence in #7410.

Confirmed with a reproduction (isolated `HOME`, standalone store holding the absolute path):

- **Before**, run from a dir containing a `studio-native/` subdir: `release studio-native --dry-run` -> `success: false`, `error: has relative local_path 'studio-native'`.
- **After**: `release studio-native --dry-run` -> `success: true`, resolved `local_path` is absolute, no `--path` needed.

## Fix

- Normalize the bare-directory `local_path` to an absolute, lexically-clean path (the same `normalize_component_local_path` normalization `component set` applies on write, #6938). `release`, `deploy`, and `component set` now resolve the same absolute checkout.
- Also normalize the project-attachment `local_path` at the shared resolution seam (`resolve_project_component_with_standalone_snapshot`) so a relative attachment can't diverge `deploy` (which silently accepted it) from `release`-via-inventory (which rejected it).
- Already-absolute, clean paths are unchanged (`normalize_local_path` is purely lexical -- no filesystem/symlink resolution), so existing behavior for registered absolute paths is untouched.

## Tests

Two real resolver assertions (no mocks):

- `resolve_effective_normalizes_relative_bare_directory_to_absolute` -- a relative bare-directory id resolves to an absolute `local_path` that `validate_local_path` accepts (the exact path `release` failed on before).
- `attachment_local_path_normalizes_matching_component_set` -- a project attachment resolves to the same absolute path `component set`'s `normalize_component_local_path` persists.

All 33 `component::resolution` tests pass. End-to-end verified against a release build: `release studio-native --skip-checks --dry-run` now resolves the absolute path and no longer errors, without `--path`.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** Claude Opus 4.8 (via Claude Code)
- **Used for:** Root-causing the resolver divergence in the Rust source (traced through inventory/bare-directory resolution with temporary instrumentation, since removed), implementing the normalization fix, and writing the regression tests. Chris reviewed and remains responsible for every line.

### DIFF
--- a/src/core/project/component/resolution.rs
+++ b/src/core/project/component/resolution.rs
@@ -61,7 +61,22 @@ pub fn resolve_project_component_with_standalone_snapshot(
     apply_standalone_component_fallbacks(&mut component, standalone_snapshot);
 
     let mut resolved = apply_component_overrides(&component, project);
-    resolved.local_path = attachment_local_path;
+    // Normalize the attachment `local_path` to an absolute path before it becomes
+    // the resolved component's machine-local checkout. Deploy resolves components
+    // directly through this function, while `release` reaches the same attachment
+    // via the composed inventory (project-attached components win on ID). Because
+    // `release` rejects a relative `local_path` outright (see
+    // `component::validate_local_path`) but deploy silently accepts one, a relative
+    // attachment value made the two commands resolve *different* checkouts for the
+    // same component id (#7410). Normalizing here — the single shared seam — makes
+    // `release`, `deploy`, and `component set` agree on one absolute path, matching
+    // the write-side normalization `component set` already applies (#6938). A tilde
+    // or absolute value is preserved as-is; a relative value is resolved against the
+    // current working directory, exactly as portable discovery already interprets it.
+    resolved.local_path = crate::core::component::normalize_component_local_path(
+        &attachment_local_path,
+    )
+    .unwrap_or(attachment_local_path);
 
     // Inherit project-level extensions when the component's homeboy.json doesn't
     // declare any. This handles clean tag clones from older releases where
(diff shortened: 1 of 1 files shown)


chubes4:
Heads up: this PR merged at commit `fb7b8a18b` (the project-attachment normalization), but the **primary root-cause fix for #7410 was not in the merged commit** — the branch was merged before the final force-pushed commit that added the bare-directory `local_path` normalization in `resolve_effective_inner`. Without it, `homeboy release <id>` still fails with "has relative local_path '<id>'" when run from a directory that has a sibling dir matching the component id (e.g. `homeboy release studio-native` from `/Users/.../Developer`). The missing half is shipped in #7422.

PR #7737, opened 2026-07-12
fix: block cook when executor require graph is unresolved on disk

### LINKED ISSUE #7736: agent-task CLI backends crash: shared runtime package 'agent-task-contracts' never materialized (MODULE_NOT_FOUND)
## Summary

Every CLI-runtime `agent-task` executor (`opencode`, `claude-code`, `codex`) crashes at startup with `MODULE_NOT_FOUND` because the shared runtime package **`agent-task-contracts`** was never materialized under `~/.config/homeboy/`, even though `runtime-agent-ci` (its sibling) was. This makes `homeboy agent-task cook` (and any CLI-backend agent-task) **completely non-functional** on this install — the executor process dies before the agent ever runs, producing empty stdout that homeboy classifies as `provider_error`.

Homeboy binary: `homeboy 0.281.20+123145ee082b` (current — matches latest release, not a stale binary).

## Repro

```
homeboy config set /agent_task/default_backend '"opencode"'
homeboy worktree create block-format-bridge --branch pilot --from main
homeboy agent-task cook \
  --to-worktree 'block-format-bridge@pilot' \
  --repo block-format-bridge \
  --backend opencode \
  --prompt 'trivial README edit' \
  --verify 'true' --no-finalize
```

Result: `status: provider_failure`, run ends in state `Failed`.

## Root cause (traced)

The executor wrapper requires the provider contract from a package that isn't on disk:

```
# ~/.config/homeboy/agent-runtimes/opencode/lib/opencode-agent-task-executor.js:17
} = require('../../../agent-task-contracts/agent-task-provider-contract');
```

`../../../` from `agent-runtimes/opencode/lib/` resolves to `~/.config/homeboy/agent-task-contracts/` — **which does not exist**:

```
$ ls ~/.config/homeboy/agent-task-contracts
ls: cannot access '.../agent-task-contracts': No such file or directory
```

The only present copy is a **compatibility shim** inside `runtime-agent-ci`, which itself re-points at the same missing package:

```
# ~/.config/homeboy/runtime-agent-ci/lib/agent-task-provider-contract.js
// Compatibility re-export. New code should import from agent-task-contracts.
module.exports = require('../../agent-task-contracts/agent-task-provider-contract');
```

So both the executors and the shim expect `~/.config/homeboy/agent-task-contracts/` to exist as a materialized sibling runtime package. It was never installed. `runtime-agent-ci` **was** installed correctly, which is why only the contracts package is missing.

## Scope — all three CLI backends affected

Same broken require in each executor wrapper:

```
agent-runtimes/opencode/lib/opencode-agent-task-executor.js:17
agent-runtimes/claude-code/lib/*.js:19
agent-runtimes/codex/lib/*.js:19
    } = require('../../../agent-task-contracts/agent-task-provider-contract');
```

## Raw stderr

```
Error: Cannot find module '../../../agent-task-contracts/agent-task-provider-contract'
Require stack:
- ~/.config/homeboy/agent-runtimes/opencode/lib/opencode-agent-task-executor.js
- ~/.config/homeboy/agent-runtimes/opencode/scripts/agent/homeboy-opencode-agent-task-executor.cjs
  code: 'MODULE_NOT_FOUND'
Node.js v25.7.0
```

Diagnostic class: `agent_task.provider_empty_stdout`
Run id: `agent-task-ed341028-b0db-424e-a165-986241430ad7` (captured via `homeboy agent-task diagnose`)

## Expected

Installing/syncing homeboy runtimes should materialize `agent-task-contracts` alongside `runtime-agent-ci` so the executors and the compat shim resolve. Either:
- the install/sync step that lays down `runtime-agent-ci` should also lay down `agent-task-contracts`, and/or
- `homeboy runtime refresh` / `agent-task doctor` should detect the missing shared package and remediate (today `agent-task doctor` reports **"Cook is ready"** despite this fatal gap — that's a second, related discoverability bug: readiness passes while the runtime is unrunnable).

## Notes for triage

- The `agent-task-provider-contract.js` in `runtime-agent-ci/lib/` is only 178 bytes and is purely a re-export shim — the real module has never been present, so this is a materialization gap, not a version skew.
- `agent-task doctor --runner local --backend opencode` returns a green "Cook is ready" verdict, which masks the problem — worth a preflight that actually resolves the executor's require graph.

chubes4:
## Summary

Closes Extra-Chill/homeboy#7736. `agent-task doctor` could report **"Cook is ready"** even when the selected provider executor could not actually load on disk. A stale or partial runtime install could pass declared provider-contract discovery and simple runner path checks, while every cook crashed before the agent ran with `MODULE_NOT_FOUND`.

## Fix

Adds an executor require-graph resolution probe for runner readiness. For selected Node-based provider executors, Homeboy invokes the executor entrypoint with `--provider-contract`, closed stdin, discarded stdout, bounded execution, and captured stderr. That loads the executor module graph without running an agent task.

- Healthy executor: reports `agent_task.provider_executor_resolution.<provider>` as ok.
- Broken require graph: runner doctor emits an error check and `agent-task doctor` returns blocked with blocker code `executor_require_graph_unresolved`.
- Unsupported/non-Node executor or missing command parts: probe is skipped as ok rather than inventing a failure.

The check is wired through both local and remote runner-doctor paths, and `agent-task doctor` passes its selected backend/selector into runner readiness so unrelated providers do not block the selected cook path.

## Architecture

Core remains runtime-agnostic: the probe keys off the resolved executable basename (`node`/`nodejs`) and the provider-declared `--provider-contract` behavior, not hard-coded backend names.

## Verification

- `cargo fmt --check`
- `cargo test executor_resolution_tests --lib`
- `cargo test commands::runner::doctor::tests::provider`
- `cargo test commands::agent_task::doctor::tests`

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** openai/gpt-5.5 via OpenCode
- **Used for:** Reviewed the PR, identified and committed the missing runner-doctor integration, updated PR metadata, and ran targeted verification under Chris's direction.

### DIFF
--- a/src/commands/agent_task/doctor.rs
+++ b/src/commands/agent_task/doctor.rs
@@ -100,6 +100,11 @@ fn provider_stage(args: &AgentTaskDoctorArgs) -> ProviderStage {
     let provider_ready;
     let mapping_value;
 
+    // When a concrete provider is selected, record its executor require-graph
+    // resolution so the verdict reflects on-disk runtime health, not just
+    // declared contract metadata (#7736).
+    let mut executor_resolution_value = Value::Null;
+
     match (selected_backend.as_deref(), backend_match) {
         (None, _) => {
             provider_ready = false;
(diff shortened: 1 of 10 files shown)


chubes4:
Filed the two VPS Rust dev-infra papercuts referenced in the CI note: #7738 (root-owned cargo registry cache) and #7739 (flaky scheduler_tests under parallel load).

PR #8244, opened 2026-07-15
fix: repair #8236 build break and complete workload command canonicalization

chubes4:
## Summary

Two things, both blocking or half-finished on `main`.

### Build break (main did not compile)

The #8236 command_contract split left the codebase with duplicate definitions:

- **`command_contract/lab_routing.rs`** was fully superseded by the new `command_contract/lab/placement.rs` (identical `impl Commands` methods — `lab_contract`, `portability_contract`, `lab_route_contract`, … — and the same helper fns) but was left in place and still `mod`-declared. This produced duplicate-definition errors (E0592) plus an import of constants that had moved out of `lab` (E0432). **Removed the orphaned file and its `mod` declaration.**
- **`output.rs`** imported `apply_lab_contract_to_descriptor` from the removed `lab_routing`. It's a pure contract→descriptor transform with no CLI dependency, so **inlined it in `output.rs`** next to the descriptor types.
- **`cli_surface.rs`** kept an `impl Placement` block, but `Placement` now lives in the `homeboy-cli-contract` crate (E0116 — inherent impl for a foreign type — plus duplicate methods). **Dropped the redundant impl**; the crate already defines both `allows_local_fallback` and `requests_lab`.

This unblocks the entire Test gate (same class of failure as #8179).

### Workload command canonicalization (#7972, completed)

Two runner-workload tests directly **contradicted** each other:
- `runner_workload_validation_rejects_dispatch_drift` expects a `test` dispatch to **validate** against a `trace` (Quality) workload.
- `runner_workload_validation_rejects_command_label_drift` expected the same dispatch to **fail**.

#7972 introduced command-identity canonicalization with the intent that same-family labels (`test`/`trace`, both `Quality`) resolve to one runner-workload identity, but never reconciled the older strict-drift test. 

Fix: the fallback validator now accepts a differing surface label when the **dispatched family matches the workload family** (falling back to strict label matching when the dispatched family is `Unknown` and therefore not canonicalizable). Updated `rejects_command_label_drift` to dispatch a genuinely un-canonicalizable label (`status`, Unknown family) so it still exercises real command-label drift rejection.

## Verification

- `cargo build --lib` and `cargo test --lib --no-run` — compile clean (were E0116/E0432/E0592).
- `core::runner::workload::tests` — **23 pass** (both formerly-contradictory tests now consistent).
- `command_contract` + `commands::infra::route` lab tests — 19 pass.


chubes4:
Superseded: the build-break portion was fixed independently by #8243. Reopening the workload canonicalization alone as a clean single-file change.

Issue #8400, opened 2026-07-16
homeboy-core test target does not compile after crate extraction (#8389): 740 errors from #[path]-included tests still using crate::core / crate::test_support


chubes4:
## Problem

After the homeboy-core crate extraction (#8389, merged into `main` as `9ff96feb4`), the extracted crate's **test target does not compile**. `cargo check -p homeboy-core --lib --tests` on current `main` fails with ~740 errors.

The `src/` path rewrites (`crate::core::*` → `crate::*`) landed, but the **`#[path = "..."]`-included test files under the top-level `tests/` tree** were not migrated with them, so they still reference the pre-extraction crate root.

## Error breakdown (from `cargo check -p homeboy-core --lib --tests`)

| Count | Error | Cause |
|---|---|---|
| 643 | `error[E0433]: failed to resolve: you might be missing crate 'core'` | `#[path]`-included `tests/core/**/*.rs` files still use `crate::core::…` |
| 93 | `error[E0432]: unresolved import` (89× `crate::test_support`, 3× `crate::core`, 1× `crate::commands`) | test helpers not remapped into the new crate |
| 4 | `couldn't read crates/homeboy-core/src/../../tests/fixtures/*.json` | `include_str!`/fixture depths not adjusted for the deeper crate path |

Example: `crates/homeboy-core/src/code_audit/../../../../tests/core/code_audit/report_test.rs:1` → `crate::core::…` (should be `crate::…`).

## Why this matters

#8389's description notes it rewrites `#[path]` / `include_str!` depths and defers the full suite to CI. That fixup was incomplete for the `tests/core/**` set, and because the suite was deferred, the breakage merged unverified. Any release/CI path that changed-scope-selects a homeboy-core test now compiles a broken test target.

Note: this is **distinct from #8386** — that issue's specific `E0425` lab_workspaces errors are resolved on current main; this is the broader crate-boundary test migration gap.

## Acceptance

- `cargo check -p homeboy-core --lib --tests` compiles clean on `main`.
- `#[path]`-included `tests/core/**` files use `crate::*` (not `crate::core::*`) and resolve `test_support`.
- `include_str!`/fixture paths resolve at the new crate depth.
- No production/test module-ownership regressions; no compatibility shims.

## Evidence
- Reproduced on `main` `9ff96feb4` via `cargo check -p homeboy-core --lib --tests` (740 errors).

## AI assistance
- **AI assistance:** Yes
[...]

PR #8414, opened 2026-07-16
fix(core): compile homeboy-core test target after crate extraction (#8400)

chubes4:
## Summary
The homeboy-core crate extraction (#8389) rewrote `src/` paths but left the `#[path]`-included test files under `tests/core/**` referencing the pre-extraction crate root. The extracted crate's test target therefore failed to compile (**740 errors** on `main`). This restores it.

## Changes
- **Mechanical path rewrite** across the 44 `tests/core/**` files the crate `#[path]`-includes: `crate::core::X` → `crate::X`, and `homeboy::X` → `crate::X`.
- **Shared test harness across the crate boundary:** move `test_support` into homeboy-core behind a `test-support` feature. The root binary enables it via `[dev-dependencies]` and re-exports `crate::test_support`, registering a **command-layer cache reset hook** so core never depends upward on the CLI `commands` layer (the one `entity_suggest` reset it needed). Root-used helpers promoted from `pub(crate)` to `pub`.
- **Fixture depths:** 4 `include_str!` paths corrected for the new crate depth (`../../tests` → `../../../tests`); `reset_config_cache_for_test` gated on `any(test, feature = "test-support")`.
- **`daemon_test.rs`:** `crate::command_contract::LabCommandContract` → `crate::lab_contract::LabCommandContract` (the type now lives in `homeboy-lab-contract`, re-exported by core as `lab_contract`).
- **Relocation:** the component-script test drives `crate::commands::test::run`, so it is a command-layer integration test and cannot live in core. Moved its `#[path]` include from `homeboy-core` to the root binary's `src/commands/test.rs`; promoted the two core fns it exercises (`run_component_scripts_with_env`, `source_path`) to `pub`.

## Verification
- `cargo check -p homeboy-core --lib --tests --features test-support` → **compiles clean** (740 → 0 errors; only pre-existing dead-code warnings).
- `cargo fmt` clean.
- Full test run deferred to CI per repo policy (the VPS cannot link the full suite).

## Scope note
The **root binary** test target has a *separate* set of ~34 pre-existing #8389 cross-crate visibility errors (`RUNNER_*_ENV`, `resource_policy_context_*`, `result_types`, etc.) in files this PR does **not** touch — those need their own follow-up and are out of scope here. This PR fixes exactly what #8400 tracks: the homeboy-core test target.

Refs #8400

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** Claude (via OpenCode), running as chubes-bot on the Extra Chill VPS
- **Used for:** Diagnosed the crate-boundary test breakage, implemented the path rewrites + shared-harness feature, and verified the homeboy-core test target compiles.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -91,6 +91,12 @@ windows-sys = { version = "0.61", features = ["Win32_Foundation", "Win32_Storage
 default = []
 slow-tests = []
 
+# Root-binary tests reuse the hermetic harness that now lives in homeboy-core.
+# Enabling the core crate's `test-support` feature for dev builds exposes
+# `homeboy_core::test_support` (re-exported as `crate::test_support`). (#8400)
+[dev-dependencies]
+homeboy-core = { version = "0.1.0", path = "crates/homeboy-core", features = ["test-support"] }
+
 [profile.dist]
 inherits = "release"
 lto = "thin"
--- a/crates/homeboy-core/Cargo.toml
+++ b/crates/homeboy-core/Cargo.toml
@@ -12,6 +12,12 @@ publish = false
 name = "homeboy_core"
 path = "src/lib.rs"
 
+[features]
+# Exposes the hermetic `test_support` harness to dependent crates (the root
+# binary) so their tests can reuse core's isolation helpers without duplicating
+# them. Enabled as a dev-dependency feature by the root crate. (#8400)
+test-support = []
+
 [dependencies]
 homeboy-cli-contract = { version = "0.1.0", path = "../homeboy-cli-contract" }
 homeboy-error = { version = "0.1.0", path = "../homeboy-error" }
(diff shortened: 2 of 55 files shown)


chubes4:
## Assessment: superseded — recommend close

The problem this PR fixes ("homeboy-core test target fails to compile after crate extraction, 740 errors") is **already resolved on current main**:

```
$ cargo check -p homeboy-core --tests
    Finished dev profile in 24s   # 0 errors
```

The `test_support` module + `test-support` Cargo feature this PR introduced already landed on main (via the `test-support-to-core` work), and `lab_contract` is re-exported (10 refs on main). This branch's `test_support.rs` now diverges from main's by 130 lines — it's a *competing older* implementation of a fix that's already in, not an additive change.

Since main's test target compiles clean, this PR's 55-file / 456-line diff is now redundant churn against a solved problem. **Recommend closing** unless there's a specific fixup here not on main — I couldn't find one.

chubes4:
Superseded by #8434 (merged), which restores homeboy-core test compilation via the same `test-support` cargo-feature approach this PR uses (feature-gated `test_support` + cache-reset helpers, `#[cfg(any(test, feature = \"test-support\"))]` gating, dev-dependency enabling it in cli/root).

The CI side is handled by homeboy-extensions #2271 (merged), which adds `--workspace` to full-scope `cargo test`/`nextest` runs so the member-crate tests actually execute (previously CI only tested the root package). Together those two land everything this PR set out to do.

Closing to avoid duplicate/conflicting changes. Thanks for the parallel work — the approaches converged.

PR #8497, opened 2026-07-16
fix(tunnel): restore core + tunnel test compilation after #8488 extraction

chubes4:
## Summary

#8488 extracted `tunnel` into `homeboy-tunnel` but **orphaned two test files in core** still referencing the removed `super::tunnel` — breaking the `homeboy-core` **test target** (`cargo check -p homeboy-core --tests` fails with E0432 on current main). #8488's own verification only checked `-p homeboy-tunnel --tests`, so core's break slipped through.

## Fix — relocate each test to where it belongs

**`tunnel_tests.rs` → moves into `homeboy-tunnel`.** It tests tunnel lifecycle/expose behavior *and* the crate-internal `preview_policy_allows` / `preview_artifact_for` (which are `pub(crate)`). Inside the crate it reaches them via `use crate::*`; its core fixtures (`Server`, `paths`, `test_support`, `ErrorCode`) become `homeboy_core::*` (tunnel already deps on core with `test-support`). **No API widening** — the alternative (dev-dep + promoting internals to `pub`) would have leaked tunnel internals.

**`preview_ingress_tests.rs` → stays in core.** It tests core's own `preview_ingress` module and only needs the public `native_preview_token_sha256`, now imported from `homeboy_tunnel`. Added `homeboy-tunnel` as a core **dev-dependency** for this — a dev-only cycle Cargo permits; it never enters the production graph.

Also removed a stray `#[cfg(test)]` that was left dangling over `pub mod update_check_cache` when `mod tunnel_tests;` was deleted.

## Verification

- `cargo check -p homeboy-core --tests` — **clean** (was E0432-broken on main)
- `cargo check -p homeboy-tunnel --tests` — clean
- 18 `tunnel_tests` + 12 `preview_ingress_tests` pass

Fixes the test-target regression from #8488.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -800,6 +800,7 @@ dependencies = [
  "homeboy-process",
  "homeboy-product-identity",
  "homeboy-redaction",
+ "homeboy-tunnel",
  "keyring",
  "libc",
  "regex",
--- a/crates/homeboy-core/Cargo.toml
+++ b/crates/homeboy-core/Cargo.toml
@@ -56,3 +56,11 @@ tempfile = "3.14"
 clap = { version = "4.5", features = ["derive", "string"] }
 toml = "1.0.3"
 sha2 = "0.10.9"
+
+[dev-dependencies]
+# Tunnel was extracted to homeboy-tunnel (#8488). Two core test targets
+# (tunnel_tests, preview_ingress_tests) still exercise tunnel behavior /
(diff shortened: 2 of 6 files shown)


chubes4:
Superseded by #8501, which landed the same core-test-compilation fix for the tunnel extraction. Closing. (The follow-on workspace-tests guard + a second orphan fix it surfaced are in a new PR.)

Issue #8425, opened 2026-07-16
homeboy-core has no dependency direction: extract leaves before domains (release/deploy blocked on engine/component cycles)


chubes4:
## Summary

The recent crate split extracted `homeboy-core` (486k LOC) and `homeboy-cli` (143k LOC), but **97% of the code still lives in one crate**. The remaining 10 crates are tiny leaves (error, paths, process, redaction, ~40–1,100 LOC each). The promised payoff — incremental compilation + lower peak build memory — isn't being delivered, because any change inside `homeboy-core` still recompiles all 486k LOC as one unit.

Root cause: **`homeboy-core` has no dependency direction.** Modules that should be the floor depend *upward* on feature modules, and domain modules depend *sideways* on each other. There are no leaves near the bottom, so no domain can be extracted without dragging ~all of core with it. This is why the split stalled at one blob and why the recent history is dominated by "break X↔Y cycle" / "decompose god-file" churn that never converges (36 of the last 100 commits are structural refactors).

## Evidence

Cyclic / wrong-direction edges found in `homeboy-core`:

- `engine` (12k LOC, should be the floor) depends **upward** on feature layers: `fixer` (22×), `refactor` (15×), `extension` (15×), `code_audit` (6×), `server`, `settings`, `project`.
- `component` depends **sideways/up** on `engine` (11×), `extension` (16×), and **`deploy` (2×)** — while `deploy` depends back on `component`. Cycle.
- `release ↔ deploy` bidirectional: `release → deploy::release_download` (7×), `deploy → release::version` (version read + tag prefix).

Because of this, extracting `homeboy-release` / `homeboy-deploy` as their own crates is **not currently possible** — not because they're entangled with each other (that's small/breakable), but because both bottom out on `engine`/`component`, which are cyclically fused to the whole feature layer.

## Highest-leverage first cut: relocate the mislocated GitHub primitive

`deploy/release_download.rs` (1,477 LOC) is **not deploy logic** — it's a GitHub client primitive (`GitHubRepo`, `parse_github_url`, `detect_remote_url`, download/sha helpers). It:

- imports **zero** `deploy` siblings (confirmed true leaf within deploy)
- only depends on `engine::temp`, `git`, `component::GithubHostConfig`, `error` (all below/beside it)
- is consumed by **27 files across git, release, triage, component, agent_task, scope** — nearly every domain

[...]

PR #8552, opened 2026-07-16
refactor: import audit config schema directly from homeboy-audit-contract (#8425)

chubes4:
## Summary

**Stage 3 of the `homeboy-audit` extraction.** After stage 1 (#8540) moved the audit config schema into `homeboy-audit-contract`, `code_audit` was still importing those types via `crate::component::{AuditConfig, DetectorProfileConfig, SourcePolicyRule, ...}` — bouncing through component's back-compat re-export instead of the crate where the types now live.

Repoint all ~24 such imports across **21 code_audit files** from `crate::component::<ConfigType>` → `homeboy_audit_contract::<ConfigType>` (and the `component::audit::*` namespace refs straight to the crate root). Pure import change; identical types, sourced from their real home.

## Result

`code_audit → component` real edges drop to only the **two genuine ones**: `Component` (the model) and `scope`. The entire config-schema coupling between audit and component is gone — both now depend on the shared contract, neither on the other for config types.

## Verification

- `cargo check -p homeboy-core --lib` — clean
- Depends on the pre-existing homeboy-cli fix (separate PR) for the full workspace guard to go green — that break is unrelated (a missing `ErrorCode::RuntimePromotionContended` match arm).

Refs #8425

### DIFF
--- a/crates/homeboy-core/src/code_audit/comment_hygiene.rs
+++ b/crates/homeboy-core/src/code_audit/comment_hygiene.rs
@@ -8,7 +8,7 @@ use super::conventions::{AuditFinding, Language};
 use super::detectors::upstream_workaround;
 use super::findings::{Finding, Severity};
 use super::fingerprint::FileFingerprint;
-use crate::component::DetectorProfileConfig;
+use homeboy_audit_contract::DetectorProfileConfig;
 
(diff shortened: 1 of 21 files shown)


chubes4:
Superseded by a combined PR that includes this config-type repoint plus the fingerprint+grammar type repoint (now that #8546 landed). Closing in favor of the combined change.

PR #8820, opened 2026-07-17
refactor: establish homeboy-observation-contract crate keystone

chubes4:
## Summary
Creates a new `homeboy-observation-contract` keystone crate and seeds it with the pure artifact-record type cluster: `ArtifactRecord` + `ArtifactViewerLink` (+ the `default_artifact_type` serde fn).

These are behavior-free serde types describing a recorded run artifact. They're carried by consumers like the **extension bench report layer**, which shouldn't have to depend on the observation store just for the record vocabulary.

## Why (keystone-first decoupling)
Same discipline as `homeboy-refactor-contract` (#8774): build the leaf contract below core, then drain type edges over time toward a future `homeboy-extension` extraction. `ArtifactRecord`'s only nested type is `ArtifactViewerLink` (moves with it); everything else is primitives — a clean, self-contained cluster.

## Effect
- New crate deps only `serde`/`serde_json` — pure leaf below core, no cycle.
- `homeboy-core` re-exports `crate::observation::{ArtifactRecord, ArtifactViewerLink}` so existing call sites are unchanged.
- Extension `bench/artifact.rs` (carries `Vec<ArtifactViewerLink>`) is **fully severed** from `crate::observation`.
- `bench/artifact_persistence.rs` takes `ArtifactRecord` from the contract too, keeping only its genuine `ActiveObservation` / finding-records behavior deps.
- Extension's `observation` edge drops from **13 files → 12**.

## Verification
- `cargo check -p homeboy-observation-contract --lib`, `-p homeboy-core --lib`, `-p homeboy-cli --lib` — 0 errors.
- `cargo check --workspace --tests --locked` (topology guard) — **passes**.
- `cargo test -p homeboy-core --lib observation::records` — **33 passed**.
- `git grep 'crate::observation'` in `extension/bench/artifact.rs` — **0 results**.

Refs #8425

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -847,6 +847,7 @@ dependencies = [
  "homeboy-lab-contract",
  "homeboy-lab-runner-contract",
  "homeboy-lifecycle-contract",
+ "homeboy-observation-contract",
  "homeboy-output",
  "homeboy-paths",
  "homeboy-process",
@@ -983,6 +984,14 @@ dependencies = [
 
 [[package]]
 name = "homeboy-lifecycle-contract"
+version = "0.1.0"
+dependencies = [
+ "serde",
+ "serde_json",
+]
+
+[[package]]
+name = "homeboy-observation-contract"
 version = "0.1.0"
 dependencies = [
  "serde",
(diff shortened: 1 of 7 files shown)


chubes4:
Superseded — ArtifactViewerLink was moved to homeboy-lifecycle-contract in a parallel cut that landed on main, making a separate observation-contract redundant. Redoing as: move ArtifactRecord to lifecycle-contract to join ArtifactViewerLink where it now lives.

PR #9104, opened 2026-07-20
feat(refactor): add 'collapse-defaults' subcommand — inverse of propagate (#9102)

### LINKED ISSUE #9102: refactor engine: add 'collapse struct defaults' capability (inverse of propagate) to automate ..Default::default() mass-adoption
## Context
We've been running a struct-construction simplification campaign (#9081, #9086, #9087, #9092, #9096): add `Default` to frequently-constructed structs and migrate call sites from fully-spelled field lists to `Struct { <meaningful fields>, ..Default::default() }`. The **primitive** (the `Default` impl + equivalence tests) is done per-struct, but the **mechanical mass-adoption** across ~45–100 sites per struct is still hand-done, one batch per PR.

The natural question: **can `homeboy refactor` do the mass-adoption itself?** I investigated. Answer: not today, but it's ~80% built.

## What exists and works
- `homeboy refactor propagate --struct-name X` already does the **inverse**: finds struct instantiations, detects *missing* fields, inserts them with inferred defaults. Verified working on this codebase — `propagate --struct-name RunRecord` scanned 128 files, found 102 instantiations, correctly identified default-valued fields.
- The AST-ish machinery lives in the Rust extension script `homeboy-extensions/rust/scripts/refactor/struct_fields.py`:
  - `parse_struct_fields()` — knows each field's name/type and `infer_default()` (Option→None, Vec→vec![], String→String::new(), bool→false, numerics→0, else Default::default()).
  - `find_struct_instantiations()` — brace-depth-aware literal finder that already handles nesting, skips struct defs / `-> Struct {` return types / shorthand init.
- Edits apply through the shared `EditOp` engine (`homeboy-refactor/src/propagate.rs`).

## What's missing (the capability gap)
There is **no** `collapse_struct_defaults` command (COMMANDS dispatch in `__main__.py` has 7 commands; none collapse). `homeboy refactor transform` is pure line/file regex and **cannot** do this safely:
- The identical line `metadata_json: Value::Null` appears **6×** as a removable default AND `metadata_json: serde_json::json!(...)` **34×** as a non-default that must be kept. A regex can't distinguish them.
- Safely collapsing requires: per-literal, compare each present field's *value* against the type's default; only when the removable set is non-empty, delete those field lines and insert `..Default::default()` exactly once before the closing brace — which needs the brace-depth tracking `find_struct_instantiations()` already has.

## Proposed capability: `collapse_struct_defaults` (inverse of propagate)
Add a sibling script + a `homeboy refactor collapse-defaults --struct-name X [--write]` subcommand that reuses the existing infrastructure:
1. Reuse `parse_struct_fields()` to get each field's default expression.
2. Reuse `find_struct_instantiations()` to locate literals (already brace-safe).
3. **New logic:** for each present field, textually compare its value to the field's default expr (`None`, `Vec::new()`/`vec![]`, `Value::Null`, `String::new()`/`"".to_string()`, `false`, `0`, and the struct's declared serde-default fn for constant defaults like `artifact_type = "file"`). Collapse only when ≥1 field is removable AND the literal doesn't already contain `..`.
4. Emit remove-line + insert-`..Default::default()` EditOps through the same apply engine.
5. Conservative by default: skip any literal with macro-valued fields it can't prove-default, skip if `..` already present, dry-run first.

### Safety / correctness
- The behavior is provably equivalence-preserving *when* the struct's `Default` matches its serde defaults (a property we already assert per-struct with byte-identical equivalence tests — see #9086/#9096). The tool should require the struct to have a `Default` impl and treat the field-default comparison textually (no eval).
- Ship with a golden-file test suite (before/after) covering: all-default trailing fields, interleaved defaults, non-default macro values that must be kept, already-has-`..` no-op, shorthand init.

## Payoff
- Turns each remaining struct's ~40–90 site migration from a hand PR into `collapse-defaults --struct-name X --write` + review.
- **Reflexive validation:** homeboy's own refactor engine improving homeboy's own codebase — the exact dogfooding that hardens the engine for external users.

## Acceptance
- [ ] `collapse_struct_defaults` script in `homeboy-extensions/rust/scripts/refactor/` with golden tests
- [ ] `homeboy refactor collapse-defaults` subcommand wired through the EditOp apply engine, dry-run by default, `--write` to apply
- [ ] Validated end-to-end on `AgentTaskOutcome` / `ArtifactRecord` (structs that already have `Default` + equivalence tests), producing the same diffs a human would

chubes4:
## Summary
Implements the **Rust side** of #9102 — `homeboy refactor collapse-defaults`, the inverse of `propagate`. Scans the codebase for instantiations of a named struct and collapses fields whose value equals the type default into a single trailing `..Default::default()`.

This is the automation for the struct-construction simplification campaign (#9081/#9086/#9087/#9092/#9096): the per-struct `Default` impl is added and equivalence-tested by hand, then this collapses the 40–90 call sites mechanically instead of hand-editing each.

**Pairs with Extra-Chill/homeboy-extensions#2308** (the `collapse_struct_defaults` analysis script). This PR is the discovery + apply + CLI wiring.

## What's here
- **New `collapse` module** (modeled on `propagate`): discovers the struct definition, sends each candidate file to the extension's `collapse_struct_defaults` script, applies the returned replace-range edits via a **bottom-up line splice** (earlier line numbers stay valid). Dry-run by default; `--write` applies after capturing an undo snapshot. Reuses propagate's `find_struct_definition`/`extract_struct_source` (now `pub(crate)`).
- **CLI**: `CollapseDefaults` clap variant + `RefactorOutput::CollapseDefaults` + `run_collapse_defaults` handler — all mirroring `propagate`.
- Uses the new `Error::invalid_argument`/`invalid_argument_for` constructors (#9081).

## Dry-run-first
This ships **dry-run by default**. Before trusting `--write` at scale, the generated diffs can be inspected. I validated it against a real un-collapsed `AgentTaskArtifact` site:
- collapsed **exactly the 3 default fields** (`url: None`, `sha256: None`, `metadata: Value::Null`)
- kept every non-default field **verbatim** (multi-line values intact)
- **left the schema-const field alone** — its `AGENT_TASK_ARTIFACT_SCHEMA.to_string()` expr isn't a recognized `String` default, so the tool conservatively keeps it (compiles, equivalent)
- **brace balance preserved** (0→0), line count −2 (3 fields → 1 spread)

## Tests
- 2 Rust unit tests for the apply/splice logic (single + multi-edit bottom-up, brace balance)
- + the extension's 9 golden tests (#2308) cover the analysis
- `cargo check -p homeboy-refactor -p homeboy-cli` clean, `cargo fmt` clean

## Reflexive win
This is homeboy's own refactor engine automating homeboy's own codebase debt — dogfooding that hardens the engine before external users touch it. Closes #9102 together with #2308.

### DIFF
--- a/crates/homeboy-cli/src/commands/refactor.rs
+++ b/crates/homeboy-cli/src/commands/refactor.rs
@@ -193,6 +193,34 @@ enum RefactorCommand {
         write_mode: WriteModeArgs,
     },
 
+    /// Collapse default-valued fields in struct instantiations into
+    /// `..Default::default()` (the inverse of `propagate`).
+    ///
+    /// Scans the codebase for instantiations of the named struct and, for each,
+    /// removes fields whose value equals the type's default (None, Vec::new(),
+    /// Value::Null, String::new(), false, 0, etc.), replacing them with a single
+    /// trailing `..Default::default()`. Conservative: skips literals that already
+    /// spread, contain an interspersed comment, or set an unknown-type field.
+    /// The struct must have a `Default` impl for the result to compile.
+    ///
+    /// Dry-run by default — pass `--write` to apply.
+    ///
+    /// Example: `refactor collapse-defaults --struct-name AgentTaskOutcome --component homeboy`
+    CollapseDefaults {
+        /// Name of the struct to collapse defaults for
+        #[arg(long, value_name = "NAME")]
+        struct_name: String,
+
+        /// File containing the struct definition (auto-detected if omitted)
+        #[arg(long, value_name = "FILE")]
+        definition: Option<String>,
+
+        #[command(flatten)]
+        target: RefactorTargetArgs,
+        #[command(flatten)]
+        write_mode: WriteModeArgs,
+    },
+
     /// Apply an ad-hoc pattern-based find/replace transform across a codebase
     ///
     /// Example: `refactor transform --find "old" --replace "new" --files "**/*.php" --component C`
@@ -386,6 +414,18 @@ pub fn run(args: RefactorArgs, _global: &crate::commands::GlobalArgs) -> CmdResu
             write_mode.write,
         ),
 
+        Some(RefactorCommand::CollapseDefaults {
+            struct_name,
+            definition,
+            target,
+            write_mode,
+        }) => operations_command::run_collapse_defaults(
+            &struct_name,
+            definition.as_deref(),
+            &target,
+            write_mode.write,
+        ),
+
         Some(RefactorCommand::Transform {
             find,
             replace,
@@ -528,6 +568,13 @@ pub enum RefactorOutput {
     Propagate {
         #[serde(flatten)]
         result: refactor::PropagateResult,
+        dry_run: bool,
+    },
+
+    #[serde(rename = "refactor.collapse_defaults")]
+    CollapseDefaults {
+        #[serde(flatten)]
+        result: refactor::CollapseResult,
         dry_run: bool,
     },
 
--- a/crates/homeboy-cli/src/commands/refactor/operations_command.rs
+++ b/crates/homeboy-cli/src/commands/refactor/operations_command.rs
@@ -85,6 +85,18 @@ pub(super) fn run_propagate(
     })
 }
 
+pub(super) fn run_collapse_defaults(
+    struct_name: &str,
+    definition_file: Option<&str>,
+    target: &RefactorTargetArgs,
+    write: bool,
+) -> CmdResult<RefactorOutput> {
+    let targets = target.resolve_targets()?;
+    run_across_targets("collapse-defaults", targets, |component_id, path| {
+        run_collapse_defaults_single(struct_name, definition_file, component_id, path, write)
+    })
+}
+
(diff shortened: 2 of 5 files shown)


chubes4:
Updated: made the language dispatch **generic** instead of Rust-pinned.

The first cut copied `propagate`'s hardcoded `find_extension_for_file_ext("rs")` — which pins collapse to Rust. That's now fixed: `collapse` enumerates every installed refactor-capable extension and, for each, scans the file extensions it handles and routes those files to that extension's `collapse_struct_defaults` command.

**Any language whose refactor extension implements the command now participates with no change to `collapse.rs`** — the Rust extension is just the first implementer. The analysis (default-value recognition, literal collapsing) correctly stays in the per-language extension script; `..Default::default()` is a Rust idiom today, but the dispatch no longer assumes it.

This makes `collapse` a true multi-language template like `transform`/`rename`, rather than inheriting `propagate`'s Rust-pinning. Verified: `homeboy-refactor` + `homeboy-cli` `--lib` clean, collapse apply tests still pass.

Issue #9206, opened 2026-07-20
Bind controller job identity before Lab snapshot preacceptance


chubes4:
## Problem

`homeboy agent-task cook` can fail during Lab handoff before provider execution because runner snapshot validation compares a real runner job ID with an empty controller job ID.

Observed error:

```text
pre-provider failure in phase `lab_handoff_preacceptance` classified as `invalid_input`:
Invalid argument 'runner_job_id': runner snapshot job <uuid> does not match controller job 
```

The trailing controller job value is empty. The run then becomes terminal `pre_execution_failure`, consumes zero provider executions, produces no candidate, and requires an unrelated control-plane repair before the requested repository work can start.

## Environment

- Controller: Homeboy `0.297.1+dfa79afb6263`
- Connected Lab daemon/job binary: Homeboy `0.297.1+d5b6dec23c7d`
- Runner: connected direct-SSH Lab runner
- Workflow: `agent-task cook` with an existing clean managed worktree, explicit `--runner`, and OpenCode backend

## Reproduction shape

1. Connect a direct-SSH Lab runner.
2. Create or select a clean managed worktree.
3. Run `homeboy agent-task cook` with `--runner <runner>`, `--to-worktree <handle>`, and an existing-worktree `--cwd`/`--workspace`.
4. Observe Lab snapshot/materialization complete and `agent-task run-plan` get submitted to the runner.
5. Observe `lab_handoff_preacceptance` reject the snapshot because the controller-side expected job ID is empty.

## Expected behavior

The accepted runner job identity should be persisted before snapshot validation and compared with the snapshot job identity. If no controller job identity is available yet, the workflow should bind the accepted runner job deterministically rather than compare against an empty string.

Validation should still reject a genuinely different non-empty job identity.

## Acceptance criteria

- Lab cook persists or derives the authoritative controller/runner job identity before preacceptance validation.
- A snapshot for the accepted job passes handoff when the controller field was initially absent.
- A snapshot carrying a different non-empty job ID still fails closed.
- Regression coverage exercises the real Lab handoff state transition rather than only a helper shape.
- The failure remains pre-provider and typed when identity evidence genuinely conflicts.

PR #9254, opened 2026-07-20
fix(agents): dedupe colliding bind_pending_lab_handoff_snapshot definitions

chubes4:
## Summary
- Fixes a **compile break on `main`**: `E0428: the name bind_pending_lab_handoff_snapshot is defined multiple times`, which broke the `Build` and `Windows Build` gates.
- Two PRs merged close together each added a function with that exact name to `lifecycle_ops.rs` — green individually, colliding once both landed:
  - **#9241** ("bind Lab snapshot job identity") — `-> Result<()>`, rebuilds the record via the durable, `LabHandoffLock`-guarded `record_detached_lab_run`. Called from `reconcile_transport_proxy_snapshot`.
  - **#9238** ("Bind Lab snapshot identity before validation", #9206) — `-> ()`, mutates `lab_handoff` in place (`pending_handoff.accepted(..)`) + metadata, no store lock/persistence. Called from `reconcile_runner_job_snapshot`.

## Resolution
Keep **#9241's** durable, locked implementation as the single definition — it is strictly more robust, and both PRs' *mismatch* tests actually rely on `record_detached_lab_run`'s fail-closed conflict rejection. Details:
- Removed #9238's inline `()` duplicate.
- #9238's call site in `reconcile_runner_job_snapshot` now uses `?` (its 3 production callers get the binding — the broadest correct coverage).
- #9241's call in `reconcile_transport_proxy_snapshot` is retained (that path also flows through `reconcile_runner_job_snapshot`; the repeat call is idempotent via the early `runner_job_id().is_some()` guard).
- Verified both PRs' guards converge: their binding tests all set a matching `target_runner_id`, and their mismatch tests trip the `runner_job_id().is_some()` early return, so #9241's stricter `target_runner_id` check satisfies both suites.

## Testing
All **four** binding tests from both PRs pass unchanged:
- `runner_snapshot_binds_pending_lab_handoff_before_validation` (#9241)
- `runner_snapshot_rejects_conflicting_bound_lab_job_identity` (#9241)
- `preacceptance_snapshot_binds_the_accepted_daemon_job_before_validation` (#9238)
- `preacceptance_snapshot_rejects_a_different_bound_daemon_job` (#9238)

`cargo check -p homeboy-agents --lib` clean (E0428 gone); `cargo fmt` clean.

### Note on an unrelated failure
`handoff_and_proxy::cook_lab_handoff_controller_reads_ignore_runner_plan_projection` fails in this module, but it is **pre-existing and unrelated**: it exercises `status()` reading a deleted controller plan file (expects `InternalIoError`), a code path this diff does not touch (the diff is only the dedup: one `?` + removing the duplicate fn). It could not surface on `main` before because `main` did not compile due to this very E0428 collision. Flagging separately rather than folding an unrelated fix into this one.

### DIFF
--- a/crates/homeboy-agents/src/agent_task_lifecycle/lifecycle_ops.rs
+++ b/crates/homeboy-agents/src/agent_task_lifecycle/lifecycle_ops.rs
@@ -2148,7 +2148,7 @@ pub(crate) fn reconcile_runner_job_snapshot(
             preserve_terminal_runner_identity(record, &event)?;
         }
     }
-    bind_pending_lab_handoff_snapshot(record, snapshot);
+    bind_pending_lab_handoff_snapshot(record, snapshot)?;
     validate_runner_job_snapshot(record, snapshot)?;
     let mut reconciled = record.clone();
     reconciled.record_runner_reachable();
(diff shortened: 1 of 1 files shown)


chubes4:
Superseded by #9251 ("fix(agents): remove duplicate Lab snapshot binder"), which already landed on main and resolved the E0428 collision the same way — keeping #9241's durable `Result<()>` `record_detached_lab_run` implementation. Closing as redundant.

PR #9862, opened 2026-07-24
Revert #9857 to restore green main (reverse-cook controller regression)

chubes4:
## Summary
Reverts the two commits merged in #9857 (`7a57f9731 fix(daemon): diagnose replacement startup publication` and `98bd2818e fix(lab): bound workspace snapshot scans`) to get `main` back to green. Both reverts apply cleanly and restore the exact file state of the last green release (#9851 / v0.310.3 baseline).

## Why
`main` has been red since #9857. The failing test is the full-suite integration test:

```
tests/reverse_cook_queue_acceptance.rs
detached_cook_accepts_reverse_capacity_queue_and_worker_completes_once ... FAILED
panicked at tests/reverse_cook_queue_acceptance.rs:332:13: controller did not project terminal broker result
```

The detached reverse-cook parent run hangs in `running` until the 30s reconcile timeout because the controller daemon never projects the terminal broker result (`stale_reason: runner_disconnected`, `runner_job_status: null`, looping `runner is not connected to a daemon` evidence-refresh warnings).

## Root-cause isolation (CI archaeology)
- **#9851** (branch `fix/9849-private-attachment-staging`): its own PR CI ran the **full** suite and this test **passed** (`... ok`).
- **#9857** (same branch, re-merged with 2 additional commits): its own PR CI already showed this test **`FAILED`** (405s timeout) — but the run was reported green and merged anyway (the full-suite failure was not enforced on that run).
- Everything merged afterward (#9852, #9859) inherited the break. #9859 (`recover terminal Lab aggregates`) was a red herring — a focused unit test confirmed its aggregate-recovery logic correctly recovers the reverse-cook embedded lifecycle event; the break is in the controller-daemon reconcile/spawn path, not aggregate projection.

The two new commits in #9857 are the delta between green (#9851) and red (#9857):
- `98bd2818e` (bound workspace snapshot scans) is **SSH-runner only**; the fixture uses a **Local** runner, so it is not the trigger (reverted here for a clean full-PR revert; it is harmless to the fixture either way).
- `7a57f9731` (daemon replacement startup diagnostics) touches the controller daemon spawn/replacement path that the detached-staging controller relies on — the actual regression.

## Follow-up
The #9857 work (startup-publication diagnostics + SSH snapshot-scan bounding) should be reopened on a fresh branch and re-landed **with the `reverse_cook_queue_acceptance` integration test enforced** so the reverse-cook controller-daemon path can't regress silently again.

## AI assistance
Investigated and reverted with OpenCode (chubes-bot). Root cause isolated by downloading and diffing the per-PR CI observation artifacts for #9851 vs #9857. Chris Huber remains responsible for the change.

### DIFF
--- a/crates/homeboy-cli/src/commands/daemon.rs
+++ b/crates/homeboy-cli/src/commands/daemon.rs
@@ -108,9 +108,6 @@ enum DaemonCommand {
         /// Local bind address. Defaults to an OS-selected loopback port.
         #[arg(long, default_value = daemon::DEFAULT_ADDR)]
         addr: String,
-        /// Startup ownership token forwarded by the daemon supervisor.
-        #[arg(long, hide = true)]
-        startup_token: Option<String>,
     },
     /// Supervise one daemon child and persist its termination evidence.
     #[command(hide = true)]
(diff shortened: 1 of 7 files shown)


chubes4:
Filed #9865 to track the safe reland requested here. Today's candidate-adoption recovery reproduced the reverted opaque failure: child exited 0, observed_state null, stderr null, then startup-token timeout. #9865 separates bounded startup diagnostics from workspace scan changes and requires reverse_cook_queue_acceptance plus the full workspace suite.

Issue #9964, opened 2026-07-24
release: bootstrapped changelog is untracked, so release aborts on its own working_tree gate


chubes4:
## Summary

The first `homeboy release` for a component that has no changelog **creates** `docs/CHANGELOG.md` (the `preflight.changelog_bootstrap` step), but leaves it **untracked**. The release then fails its own `working_tree` cleanliness gate because of the file it just wrote — so the very first release of any new component always aborts.

## Repro

New component, no `docs/CHANGELOG.md` yet, clean tree:

```
$ git status --short          # clean
$ homeboy release --project h44-lacrosse --skip-checks --apply
```

Result:

```
Invalid argument 'working_tree': Uncommitted changes detected
details: { "files": ["plugins/h44-forms/docs/"], "hint": "Commit changes or stash before release" }
```

`plugins/h44-forms/docs/` is the changelog directory the release created moments earlier.

Hit this on two components in a row (theme, then a plugin). Workaround each time: manually `git add` + commit the bootstrapped changelog, push, then re-run the release. For the third component I pre-created `docs/CHANGELOG.md` by hand before releasing to avoid the abort entirely.

## Impact

Every component's *first* release fails, with an error that points at a file the tool itself generated. It reads like the operator left the tree dirty, which is misleading. Also leaves partial state behind (the untracked changelog) that the operator has to clean up or commit manually.

## Expected

One of:
- Stage/commit the bootstrapped changelog as part of the release (it's release-owned content, and the release already creates version-bump commits), or
- Exclude the just-bootstrapped changelog path from the `working_tree` check, or
- Bootstrap the changelog *after* the working-tree gate rather than before it.

## Environment

- homeboy `0.306.0+d33eb3b8b69d`
- Monorepo with per-component `homeboy.json`, each declaring `"changelog_target": "docs/CHANGELOG.md"`

Issue #9963, opened 2026-07-24
deploy: deploying a release tag ships pre-version-bump content (live version != tag version)


chubes4:
## Summary

`homeboy deploy` of a freshly-created release tag shipped files whose version strings were **one bump behind the tag**. Deploying tag `X` should be byte-identical to tag `X`.

## Repro (WordPress project, 3 components)

1. `homeboy release --project h44-lacrosse --skip-checks --apply`
   - Succeeded: `{'failed': 0, 'released': 3, 'skipped': 0, 'total': 3}`
   - Created tags `h44-lacrosse-theme-v0.2.2`, `h44-forms-v0.2.1`, `h44-core-v0.1.1`
2. `homeboy deploy -p h44-lacrosse h44-lacrosse-theme h44-forms h44-core`
   - Reported success: `status: succeeded`, `deployed_ref = h44-core-v0.1.1 (main)`, `status = deployed` for all three.
3. Inspected the remote install:

| Component | Tag | Version actually deployed |
|---|---|---|
| h44-lacrosse-theme | v0.2.2 | **0.2.1** |
| h44-forms | v0.2.1 | **0.2.0** |
| h44-core | v0.1.1 | **0.1.0** |

## The tags themselves are correct

Verified the tag contents are the bumped versions, so this is not a bad tag:

```
$ git show h44-core-v0.1.1:plugins/h44-core/h44-core.php | grep -m1 Version:
 * Version:           0.1.1        # tag is correct

$ ssh box 'grep -m1 Version: .../plugins/h44-core/h44-core.php'
 * Version:           0.1.0        # what deploy actually shipped
```

Local `HEAD` also matched the tags (`0.2.2` / `0.1.1`), so the discrepancy is purely between "what the tag contains" and "what deploy uploaded".

## Impact

Silent and easy to miss — deploy reports `succeeded` with the right `deployed_ref`, so nothing signals that the remote is running older code than the tag claims. On a WordPress install this also means asset cache-busting (`?ver=`) stays on the previous version, so browsers keep serving stale CSS/JS after a release.

## Suspected cause

Looks like the deploy artifact is built/cached before the release's version-bump commit is applied, so the packaged content predates the bump. A plain `rsync` of the local checkout (which matched the tags) immediately corrected all three.

## Expected

Deploying a tag uploads exactly that tag's content; version strings on the remote match the tag. If a cached artifact is reused, it should be invalidated when the underlying ref/content changes.

## Environment

[...]

PR #9983, opened 2026-07-24
Ignore composer bookkeeping churn from preflight.dependencies (#9965)

### LINKED ISSUE #9965: release: preflight.dependencies dirties tracked composer files, blocking the release it's preparing
## Summary

`preflight.dependencies` runs `composer install`, which regenerates `vendor/composer/installed.json` and `vendor/composer/installed.php` with differing content. Those files are tracked in the repo, so the release then fails its own working-tree gate — the preflight step blocks the release it is preparing.

## Repro

Component with a committed `vendor/` (composer autoloader shipped in the plugin), clean tree before the run:

```
$ homeboy release --project h44-lacrosse --skip-checks --apply
```

Result:

```
Invalid argument 'working_tree': preflight.dependencies dirtied tracked files:
  plugins/h44-forms/vendor/composer/installed.json,
  plugins/h44-forms/vendor/composer/installed.php

hints:
  - Commit intentional generated asset changes before retrying the release
  - Fix the build/test/package step so it is deterministic and leaves tracked files unchanged
  - Restore generated output with `git restore <file>` before rerunning if the change was not intentional
```

The diagnostics are good and clearly anticipate this class of problem — but here the non-determinism comes from homeboy's *own* preflight step, not from anything the operator did.

## Workaround

Untracked those two files and gitignored them (they are pure generated artifacts, recreated by `composer install`):

```
git rm --cached plugins/h44-forms/vendor/composer/installed.json \
                plugins/h44-forms/vendor/composer/installed.php
echo "vendor/composer/installed.json" >> plugins/h44-forms/.gitignore
echo "vendor/composer/installed.php"  >> plugins/h44-forms/.gitignore
```

That made releases deterministic. But WordPress plugins commonly ship a committed `vendor/` (so the plugin works without a build step on the host), and `installed.json`/`installed.php` are the two files composer rewrites non-deterministically — so this is likely to bite any PHP component that vendors its autoloader.

## Suggested

- Treat composer's `vendor/composer/installed.*` churn as expected noise from `preflight.dependencies` and exclude it from the dirty-tracked-files check, or
- Skip `composer install` during preflight when `vendor/` is already present and the lock file is unchanged, or
- Document the gitignore recommendation for vendored PHP components.

## Environment

- homeboy `0.306.0+d33eb3b8b69d`
- Composer 2.10.2, PHP 8.5 (remote) / 8.3 (local)
- WordPress plugin component with committed `vendor/` and `"extensions": { "wordpress": {} }`

chubes4:
Fixes #9965. Third of the release trilogy (#9963 / #9964 / #9965); #9964 is merged.

## Problem
`preflight.dependencies` runs `composer install`, which regenerates `vendor/composer/installed.json` and `installed.php` with differing content **every run**. Components that vendor their autoloader — common for WordPress plugins that must work without a build step on the host — track those files, so the step dirtied the tree and tripped the release's own dirty-tracked-files guard:

```
Invalid argument 'working_tree': preflight.dependencies dirtied tracked files:
  plugins/h44-forms/vendor/composer/installed.json,
  plugins/h44-forms/vendor/composer/installed.php
```

The preflight blocked the release it was preparing — and the non-determinism came from Homeboy's own step, not from anything the operator did.

## Root cause
`release_step_unexpected_dirty_files` applied an allow-list **only** for `release.prepare`; every other guarded step (`preflight.dependencies`, `preflight.lint`, `preflight.test`, `package`, …) reported its dirty files verbatim.

## Fix
Filter Composer's regenerated bookkeeping files (`vendor/composer/installed.json`, `vendor/composer/installed.php`) out of the `preflight.dependencies` result — that churn is expected noise from Homeboy's own step.

Deliberately scoped:
- Only `preflight.dependencies` (the step that actually runs `composer install`) gets the exclusion; other guarded steps still report those paths.
- Only the two files Composer rewrites non-deterministically; the rest of `vendor/` is still guarded.
- Real source changes introduced by the step are still reported.

This implements the issue's first suggestion (treat the churn as expected noise) rather than skipping `composer install` or pushing a gitignore recommendation onto every vendored PHP component.

## Tests
- `preflight.dependencies` ignores composer bookkeeping while still flagging a real source change.
- Another guarded step (`preflight.test`) still reports those same paths.

## Verification
- `cargo fmt --all --check` clean
- `cargo check -p homeboy-release --lib --tests` exit 0

Diff: +94, 1 file.

### DIFF
--- a/crates/homeboy-release/src/release/execution_dispatch.rs
+++ b/crates/homeboy-release/src/release/execution_dispatch.rs
@@ -688,6 +688,19 @@ fn release_step_unexpected_dirty_files(
     context: &ReleaseExecutionContext,
     files: Vec<String>,
 ) -> Result<Vec<String>> {
+    if step.kind == "preflight.dependencies" {
+        // `composer install` rewrites its own bookkeeping files with differing
+        // content every run. Components that vendor their autoloader (common for
+        // WordPress plugins that must work without a build step) track those
+        // files, so this preflight would otherwise dirty the tree and block the
+        // release it is preparing (#9965). This churn is expected noise from
+        // Homeboy's own step, not an operator-introduced change.
+        return Ok(files
+            .into_iter()
+            .filter(|file| !is_composer_regenerated_bookkeeping_file(file))
+            .collect());
+    }
+
     if step.kind != "release.prepare" {
         return Ok(files);
     }
@@ -697,6 +710,23 @@ fn release_step_unexpected_dirty_files(
         .into_iter()
         .filter(|file| !allowed.iter().any(|allowed| file == allowed))
         .collect())
+}
+
+/// Whether a path is one of Composer's non-deterministically regenerated
+/// bookkeeping files under `vendor/composer/`.
+///
+/// `composer install` rewrites `installed.json` and `installed.php` with
+/// differing content (ordering/timestamps/paths) even when the lock file has not
+/// changed, so their churn is not a meaningful tracked-file modification.
+fn is_composer_regenerated_bookkeeping_file(path: &str) -> bool {
+    const COMPOSER_REGENERATED: &[&str] = &[
+        "vendor/composer/installed.json",
+        "vendor/composer/installed.php",
+    ];
+    let normalized = path.replace('\\', "/");
+    COMPOSER_REGENERATED
+        .iter()
+        .any(|suffix| normalized == *suffix || normalized.ends_with(&format!("/{suffix}")))
 }
 
 fn release_prepare_allowed_dirty_files(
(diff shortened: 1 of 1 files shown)


chubes4:
Closing: this hardcodes Composer/PHP ecosystem knowledge (`vendor/composer/installed.json`) into `homeboy-release`, which violates the core-agnostic-source boundary (there is even a `core_boundary_leak:core-agnostic-source` audit convention that flags exactly this class of ecosystem literal in core).

The correct seam is the existing declaration-driven mechanism: components already declare generated paths in their own `homeboy.json` (see `declared_build_artifact_paths`), and the dirty-tracked-files guard filters declared artifacts. Reworking #9965 to be ecosystem-agnostic — a component (or the WordPress/PHP extension) declares the volatile paths; core just honors the declaration.

Issue #10058, opened 2026-07-25
deploy --check / fleet check report up_to_date while remote file content differs (version-only drift detection)


chubes4:
## Summary

Component drift detection compares **version strings only**. When a deployed component's files are modified on the remote without the version header changing, `deploy --check` (and by extension `fleet check`) reports `component_status: up_to_date` even though local and remote content genuinely differ.

This gives false confidence exactly where it is most dangerous: the operator believes local and remote agree, so the next deploy silently overwrites remote-only work.

## Deterministic repro

Component: a WordPress theme deployed to a project's server.

```bash
# 1. Baseline: local and remote agree
homeboy deploy <project> <component> --check
#   component_status = up_to_date
#   local_version = 0.2.2 / remote_version = 0.2.2

# 2. Change remote CONTENT without touching the version header
ssh <server> 'echo "<?php // drift-probe" >> <remote_path>/404.php'

# 3. Check again
homeboy deploy <project> <component> --check
#   component_status = up_to_date      <-- WRONG
#   local_version = 0.2.2 / remote_version = 0.2.2

# 4. Ground truth
ssh <server> 'md5sum <remote_path>/404.php'   # 563984bb11f60eb4fc66c5989d7980aa
md5sum <local_path>/404.php                    # 433b20c548a8950b5f0e1810197b7c5a
```

Checksums differ; `--check` still says `up_to_date`.

## How this bit us in production

On a managed WordPress install, an AI agent was asked to update site copy and edited two theme templates **directly on the live server** (`archive-h44_event.php`, `archive-h44_team.php`). Real content authored by the site owner.

`homeboy deploy --check` reported `up_to_date`, because the agent had no reason to bump a version header. The changes existed **only** on the remote and were not in version control. The next `homeboy deploy` (or any rsync) would have silently destroyed them, with nothing in homeboy's output to warn the operator.

We only caught it by hand-diffing file mtimes and checksums.

## Expected

Drift detection should be **content-aware**, not version-aware. Some options:

- Compare a content hash (per-file or a manifest digest) between `local_path` and `remote_path`, in addition to the version comparison.
- Introduce a distinct status such as `content_drift` / `remote_modified` so it is distinguishable from both `up_to_date` and `needs_update`.
- Surface which files differ, so an operator can review before deploying over them.

At minimum, `--check` should not report `up_to_date` when the deployed bytes do not match local.

## Why it matters beyond this case

[...]

PR #10076, opened 2026-07-25
feat(cli): add component harvest

### LINKED ISSUE #10060: Feature: homeboy harvest — recover drift from deployed sites (inverse of deploy)
## Problem

`homeboy deploy` pushes a component's code **to** a server. Nothing pulls it **back**. On a managed install where something other than deploy writes to the site, that work exists only on the remote and is destroyed by the next deploy — silently.

Not hypothetical. On 2026-07-25 an AI agent running on a client's own VPS was asked by the site owner to update site copy, and edited two theme templates **directly on the live server**:

- `archive-h44_team.php` — real team age groups (10U, 12U, 14U, High School) plus "Teams will be announced after the combine."
- `archive-h44_event.php` — "Fall events coming soon!"

Genuine owner-authored content, existing nowhere but the remote. `homeboy deploy --check` reported `component_status: up_to_date` (see #10058 — drift detection is version-only), so the operator had no signal. The changes were recovered by hand-diffing mtimes and checksums, one deploy away from being lost.

### Why "make the agent use git" isn't the answer here

It's the right answer on a developer's machine. It's the wrong answer for a **managed install owned by a non-technical person**. Requiring the site owner to hold a GitHub account and the agent to hold push credentials on a box the operator doesn't fully trust is friction that won't survive contact with reality. On this install the agent's workspace was empty and it had no GitHub auth, so editing live files was its *only* available action.

Live editing should be an **accepted, supported path** for managed installs, with the operator's tooling responsible for capturing it.

## Prior art

`harvest` is already an established homeboy verb — lab-runner harvests work products back from a remote cook worker: `harvest_uncommitted_patch`, `harvest_committed_patch_with_metadata`, `harvest_preflight`/`harvest_preflight_outcome`, `harvest_context`, `harvest_transport`, plus provenance checks.

Same operation, different source:

| | lab-runner harvest | proposed site harvest |
|---|---|---|
| Source | remote cook worker workspace | deployed component `remote_path` |
| Mechanism | git patch from a worktree | content diff vs `local_path` |
| Trigger | after a cook completes | on demand / scheduled |

This should **extend the existing verb** and reuse patch/provenance machinery where it fits, not introduce a new concept.

## Proposal

```
homeboy harvest <project> [components...]
homeboy harvest --fleet <fleet>
```

The inverse of `deploy`. Per component:

1. Resolve `local_path` ↔ `remote_path` from the existing project/component model (no new config — deploy already uses it).
2. Compare **content**, not versions (depends on #10058).
3. Materialize remote-only changes into `local_path`.
4. Commit with provenance identifying the change came from the remote, not a developer.
5. Optionally push / open a PR.

### Flags

| Flag | Behavior |
|---|---|
| `--check` | Report drift, write nothing. Safe to schedule. |
| `--dry-run` | Show the patch that would be applied. |
| `--apply` | Write, commit, and with `--push` push. |
| `--author` | Attribution for the commit (e.g. the agent identity). |
| `--exclude` | Paths to ignore (`vendor/`, `node_modules/`, build output). |

### Attribution

The commit author should reflect **who changed the remote**, with the harvesting operator as committer. Done by hand today as `git -c user.name='h44-bot' … commit`. Harvest should make that the default so history honestly records that the agent authored the content and the operator recovered it.

## Conflict handling

Harvest must **never silently pick a winner**. In order of preference:

1. **Refuse and report** the conflicting files — safest default.
2. **Harvest to a branch** (`harvest/<project>/<date>`) so a human resolves via PR — good default for scheduled runs.
3. `--force-remote` / `--force-local` — explicit, never implicit.

Related guard rail: **`deploy` should refuse (or loudly warn) when it would overwrite un-harvested remote drift.** That alone would have prevented this incident.

## Scope

Harvest **the components homeboy already manages** — theme and first-party plugins — not the whole site. WordPress core and third-party plugins are versioned upstream and would bury real changes in vendor noise; uploads are media, not code. The component model already draws the right boundary.

Sensible default excludes: `vendor/`, `node_modules/`, build artifacts, `*.log`, and anything in the component's `.gitignore`.

## Scheduling and credentials

**Pull model.** Harvest runs from the operator's machine, SSHing into the project's server exactly as `deploy` already does.

This matters: **no GitHub credentials ever land on the client's box.** On this project the operator explicitly refused to place a token there — the box runs an autonomous agent, and the operator's GitHub account has access to unrelated organizations. A push model would need a deploy key per client machine and widen the blast radius for no real benefit.

A daily `--check` plus on-demand `--apply` is likely the right operational default.

## Open questions

1. Should `deploy` hard-refuse on un-harvested drift, or warn? (Refusing is safer but could block urgent fixes.)
2. Reuse lab-runner's patch/provenance machinery, or is the file-diff path different enough to warrant its own?
3. How should binary/media files be treated if they appear under a component?
4. Should `--check` results feed `homeboy triage`, so drift appears in the existing attention report rather than needing separate cron?

## Environment

- homeboy `0.306.0+d33eb3b8b69d`
- Project: WordPress site, 3 components (theme + 2 plugins), monorepo with per-component `homeboy.json`
- Related: #10058 (version-only drift detection makes this invisible today)

chubes4:
## Summary

- add `homeboy harvest` as the pull-side inverse of deploy for managed components
- compare remote and local component bytes with configurable component-owned exclusions
- support non-mutating check/dry-run reports and conflict-safe single-component apply commits
- preserve remote authorship and `Harvested-from` provenance in Git history

## Safety

- core remains framework and product agnostic
- apply refuses dirty local worktrees and multi-component mutations
- Git staging is scoped to the harvested component
- binary additions/modifications, remote deletions, and recursive excludes are covered
- deploy drift protection remains owned by the in-place manifest from #10058 / #10065

## Verification

- `cargo fmt --all -- --check`
- `rustfmt --edition 2021 --check crates/homeboy-cli/src/commands/harvest.rs`
- `cargo check --workspace`
- `cargo test -p homeboy-core content_diff::tests`
- `cargo test -p homeboy-core harvest::tests`
- `cargo test --workspace` compiled the workspace and passed harvest coverage, but remains red on unchanged test `runner_status_includes_lab_diagnostics`: expected fixture runner `lab-local`, observed `null`; reproduced independently with `cargo test -p homeboy --test output_errors runner_status_includes_lab_diagnostics -- --exact --nocapture`

Closes #10060

## AI assistance

OpenCode with `openai/gpt-5.6-terra` drafted the initial implementation and tests. OpenCode with `openai/gpt-5.6-sol` reviewed and corrected transaction boundaries, scoped Git staging, exclusion semantics, documentation, and verification. Chris Huber remains responsible for review and every line.


chubes4 (commented):
Reviewed as the requester of #10060, against a real managed WordPress project (3 components on a remote VPS). The shape is right and matches what I asked for — `--check` exiting 2 on drift is exactly the hook a scheduled run needs, and `--apply` with `--author` + `Harvested-from` provenance is precisely the manual work I have been doing by hand.

One correctness issue worth fixing before merge.

## Path resolution diverges from deploy

Harvest computes the remote root as:

```rust
let remote_path = crate::join_remote_path(Some(&base_path), &component.remote_path)?;
```

Deploy computes it as:

```rust
let install_dir = resolve_effective_remote_path(project, component, base_path)?;
```

`resolve_effective_remote_path` (crates/homeboy-release/src/deploy/path_roots.rs) additionally honors managed path-root rules, parent-relative managed paths, and absolute-path handling. Harvest reuses none of it — I grepped the diff and the only `strip_prefix` hits are ordinary `str::strip_prefix` calls, so `project.path_roots` is never consulted.

That matters because the WordPress extension declares:

```json
{"path_prefix": "wp-content/", "root": "wp_content", "strip_prefix": true, "detect_command": "wp eval 'echo WP_CONTENT_DIR;'"}
```

So for any component whose `remote_path` begins with `wp-content/`, deploy **strips that prefix** and joins the remainder onto the detected `WP_CONTENT_DIR`, while harvest concatenates the unstripped path onto `base_path`.

These coincide only when `WP_CONTENT_DIR == <base_path>/wp-content`. They diverge for any install with a relocated content directory — Bedrock (`web/app`), a custom `WP_CONTENT_DIR`, or a moved `wp-content`. The `detect_command` in that rule exists precisely because the location is meant to be detected at runtime rather than assumed.

On my own project the two agree (default layout), and harvest would work today. But the failure mode elsewhere is bad: harvest either errors on a nonexistent remote root, or silently reads a stale parallel directory and reports it as drift against content deploy never wrote there.

Since harvest is billed as "the pull-side inverse of deploy," I would argue it should resolve the remote root through the *same* function by construction — any divergence between the two is a correctness bug by definition, and a duplicated join will drift again as path rules evolve.

Suggested fix: promote `resolve_effective_remote_path` into core as a shared primitive and have both deploy and harvest call it. It is already core-shaped (`Project` + `Component` + `base_path`), and it does not compromise the PR's framework-agnostic goal — the rules come from the extension contract, core only evaluates them.

## Smaller note

The empty-`remote_path` guard at harvest.rs:380 looks unreachable in practice: `resolve_project_component` already layers manifest to attachment to standalone and then calls `resolve_remote_path()` for extension auto-resolution, so the field arrives populated. Harmless as a last-resort guard, but worth confirming it is not masking the auto-resolution path in a test fixture that sets `remote_path` explicitly.

## What I verified

- `homeboy deploy h44-lacrosse h44-core --check` resolves `/var/www/h44lacrosse.com/wp-content/plugins/h44-core`
- Component manifest carries `remote_path: wp-content/plugins/h44-core`; the project attachment carries none
- Project `path_roots.wp_content` is set, and is ignored by the harvest path computation

chubes4:
Correctness follow-up pushed in `ab6ea2a76`:

- Moved deploy effective remote-path and `path_roots` detection into `homeboy-core::project` as the shared deploy/harvest primitive.
- Harvest now resolves detected managed roots before reading remote content.
- Deploy retains `[deploy]` status output; harvest uses `[harvest]`.
- Added a harvest regression exercising an extension-detected relocated managed root.

Passing locally:
- `cargo fmt --all -- --check`
- `cargo check --workspace`
- `cargo test -p homeboy-core project::effective_remote_path::tests` (12 passed)
- `cargo test -p homeboy-core harvest::tests` (2 passed)
- `git diff --check`

A local broad `cargo test -p homeboy-release` run reached 651 passing tests and 7 unrelated environment/fixture failures (missing isolated extensions/build fixtures and filesystem group assumptions). The changed deploy caller compiles under the workspace check; CI is running the clean hosted suite.

PR #10432, opened 2026-07-27
fix(lab-runner): keep snapshot materialization POSIX-sh portable

### LINKED ISSUE #10399: fix snapshot materialization shell portability in workspace prune tests
## Problem

Current main `ce45b23b9106323be537669132cce15caa6c0bf3` fails ten Lab workspace prune tests because local snapshot materialization executes a Bash-shaped command through `sh`, which rejects it with `Syntax error: "(" unexpected`.

## Evidence

Release run: https://github.com/Extra-Chill/homeboy/actions/runs/30235333698

Affected tests include `prune_workspaces_previews_orphans_without_deleting_by_default`, `prune_workspaces_apply_removes_only_metadata_backed_orphans`, TTL lifecycle pruning, drain-until-empty, and bounded mixed-entry scans.

## Expected contract

- Snapshot materialization commands use the declared shell contract or remain POSIX-sh portable.
- Workspace prune behavior tests execute through the same shell boundary as production.
- The full affected prune group passes on Linux without weakening lifecycle, ownership, continuation, or deletion assertions.

## AI assistance

OpenAI GPT-5.6 Sol via OpenCode inspected the current-main Release failure and drafted this tracker. Chris Huber reviewed and owns it.

chubes4:
## RELEASE-BLOCKING

No release has shipped since **v0.319.2** (Jul 26 22:39). Every release run since has failed the Test gate on ten workspace-prune tests, so multiple merged fixes — including a fanout repair the operator is waiting on — have never reached users. This PR clears that gate.

`Closes #10399`

## The offending construct

Scratch admission wrapped the snapshot command with a `TMPDIR=<dir>` **assignment prefix**, but `snapshot_archive_command` always emits a leading `(...)` subshell. POSIX allows assignment prefixes only on *simple* commands, so the whole program is a parse error before a byte is transferred:

```
$ dash -c 'TMPDIR=/tmp/x (cd /tmp && echo hi)'
dash: 1: Syntax error: "(" unexpected
```

`sync_workspace` passes `Some(scratch)` **unconditionally**, so every local materialization took this path — which is why ten prune tests all died with the same `sh: 1: Syntax error: "(" unexpected`.

Worth recording: this was **not** a bash-vs-dash divergence. Bash rejects the construct too. The tests never caught it because nothing exercised the assembled command except a full materialization.

## What this changes

The prefix itself was already corrected. What remained — and what this PR fixes — is the reason no test could see it.

- **`snapshot_materialization_command`** is now the single construction point for the program `materialize_snapshot_piped` hands to `sh -c`. The archive pipeline and scratch scoping were composed inline in the runner, so each part was individually well-formed while the *composition* was unobservable. Tests now assert against the byte-for-byte string production runs, across exclude shapes, both runner targets, and scratch admitted or not.
- **Validate under `dash` when present**, not just `sh -n`. Where `/bin/sh` is bash, its parser accepts bash-only syntax dash rejects — process substitution, the `function` keyword, array assignments — so the old check could pass locally and fail on the runner. Measured:

  | construct | `dash -n` | `bash -n` |
  |---|---|---|
  | `TMPDIR=x (cd ...)` | REJECT | REJECT |
  | `cat <(echo hi)` | REJECT | accept |
  | `function f() { :; }` | REJECT | accept |
  | `a+=(1)` | REJECT | accept |

  The comment is explicit that a parse check cannot catch `[[ ... ]]` or `((...))`, which parse as ordinary words under dash and only fail at run time.
- **`snapshot_install_command` and `snapshot_overlay_install_command`** are held to the same contract; they cross the same boundary.

Fix direction is POSIX portability, not switching the executor to bash — bash is not guaranteed on an SSH runner, and the SSH path (`client.execute`) has no control over the remote login shell.

## Verification

Per this host's resource rules, no `cargo build`/`test`/`check`/`clippy` was run — only `cargo fmt`. Commands were reconstructed faithfully and syntax-checked directly:

- Reproduced the exact CI error from the old shape under `dash -n`.
- Confirmed the current archive shapes (plain + root-anchored excludes), with and without scratch scoping, and both install commands (including space-containing paths) all parse cleanly.
- Audited the lab-runner shell literals for `<(`, `>(`, `[[`, `function `, `mapfile`, `readarray` — no other live non-POSIX construct.

**CI passing is the acceptance criterion here.** Full build and test are left to CI.

### DIFF
--- a/crates/homeboy-lab-runner/src/workspace/snapshot.rs
+++ b/crates/homeboy-lab-runner/src/workspace/snapshot.rs
@@ -1703,11 +1703,30 @@ fn materialize_snapshot_piped(
     action: &str,
     scratch: Option<&Path>,
 ) -> Result<()> {
-    let mut command = snapshot_archive_command(local_path, target_command, excludes);
-    if let Some(scratch) = scratch {
-        command = scratch_scoped_command(&command, scratch);
-    }
+    let command = snapshot_materialization_command(local_path, target_command, excludes, scratch);
     run_shell_command(&command, action)
+}
+
+/// Build the exact shell program a snapshot materialization hands to `sh -c`.
+///
+/// This is the single construction point for that program so a test can assert
+/// against the byte-for-byte string production runs. Composing the archive
+/// pipeline and scratch scoping inline in the caller left the assembled command
+/// unobservable: `scratch_scoped_command` was individually well-formed, yet the
+/// only thing that ever exercised the *composition* was a full materialization,
+/// so a shell-syntax regression reached `sh -c` with no test able to see it and
+/// surfaced as ten unrelated workspace-prune behavior failures (#10399).
+pub(super) fn snapshot_materialization_command(
+    local_path: &Path,
+    target_command: &str,
+    excludes: &[String],
+    scratch: Option<&Path>,
+) -> String {
+    let command = snapshot_archive_command(local_path, target_command, excludes);
+    match scratch {
+        Some(scratch) => scratch_scoped_command(&command, scratch),
+        None => command,
+    }
 }
 
 /// Scope one snapshot command to an admitted controller scratch filesystem.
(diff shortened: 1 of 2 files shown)


chubes4:
### Expect a red Test gate on this PR — it is not from this change

This branch is cut from a `main` that already fails 9 workspace-prune tests for an unrelated reason, tracked as #10426 and introduced by 0b17eb15d (#10425, *bound workspace prune page wall time*).

That is a **different** failure from the one this PR is about. The `sh: 1: Syntax error: "(" unexpected` breakage is gone from `main` — the latest release run contains zero `Syntax error` occurrences. The current failures are `assertion left == right failed, left: 0, right: 1`: prune returns no candidates because `classify_local_candidate` now downgrades liveness to `"unknown"` whenever `bounded_directory_size` fails, and only `"inactive"` entries become candidates. Full mechanism in https://github.com/Extra-Chill/homeboy/issues/10426#issuecomment-5092535649.

So the release needs **both**: #10426 fixed, and this PR to keep the shell boundary from regressing again.

Compile-level gates on this PR are green — Rustfmt, Windows Compile, and Workspace Tests Compile all pass — which is the meaningful signal available here, since the new tests are the thing that must build and run.

Issue #8340, opened 2026-07-15
Fail changed-scope test gates when source changes select zero tests


chubes4:
## Problem

The release Test gate can report success without executing any tests when a Rust source file changes.

In [Release run 29438700149](https://github.com/Extra-Chill/homeboy/actions/runs/29438700149), the Test job passed in six seconds for `4ca328bf2084b03680da25767e013fa98993e002`. Its uploaded `homeboy-observations-homeboy-review-test-gate-test-Linux` artifact records:

```json
{
  "command": "homeboy test homeboy --skip-lint --changed-since=f3565cbf4b6ed96fee780f6fb440ea740cc1a19c",
  "status": "pass",
  "test_scope": {
    "changed_since": "f3565cbf4b6ed96fee780f6fb440ea740cc1a19c",
    "mode": "changed",
    "selected_count": 0
  }
}
```

The compared commit changed `src/core/code_audit/walker.rs` and added assertions in that file, so zero selection is not meaningful release evidence.

## Expected contract

- A changed-scope test gate that selects zero tests fails closed when relevant production or test source changed.
- The result explains why no tests were selected and reports the impacted changed files.
- Explicitly documentation-only or otherwise configured no-test scopes may pass with a typed reason.
- Deterministic coverage proves both fail-closed source changes and allowed no-test scopes.

## Relationship

Follow-up to #8271 and the release reliability work under #8010.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** OpenCode with openai/gpt-5.6-sol
- **Used for:** Inspected the release workflow artifact, identified the zero-test false-positive gate, and drafted this tracker. Chris owns the report.

PR #10450, opened 2026-07-27
fix(test-scope): recognize monorepo member sources in changed-scope drift

### LINKED ISSUE #10449: Changed-scope test gate is blind to Cargo workspace member crates (selects no tests, never fails closed)
## Problem

The changed-scope test gate is blind to Cargo workspace member crates. Since homeboy moved its code into `crates/*`, **changes to essentially the entire codebase select no tests from the changed crate and do not trip the `#8340` fail-closed guard** — the gate reports green having run none of the relevant tests.

## Root cause

The Rust extension declares `drift.source_dirs: ["src"]`. `glob_patterns` turns that into a single root-anchored pattern:

```
src/**/*.rs
```

`matches_any_pattern` uses `glob::Pattern::matches_path`, which is anchored. So:

```
src/lib.rs                                          -> matches
crates/homeboy-lab-runner/src/lab/offload/hydration.rs -> NO MATCH
crates/homeboy-core/src/daemon/stop.rs                 -> NO MATCH
crates/contracts/homeboy-audit-contract/src/finding.rs -> NO MATCH
```

Only the thin root package (`src/lib.rs`, `src/main.rs`, `src/bin/`, `src/extensions/`) is visible. Everything else reads as "not source".

Two independent protections then fail together:

1. **Selection.** `detect_drift_for_changed_files` builds `prod_files` with `matches_any_pattern(f, &opts.source_patterns)`. For a workspace-member change that list is empty, so it returns early with zero drifted tests.
2. **Fail-closed guard.** `source_changes_without_tests` is built from `source_relevant_changed_files`, which uses the same patterns. A workspace-member change never lands there, so the `#8340` "selected zero tests but source changed" guard — the thing that is supposed to fail closed — never fires.

## Evidence

PR #10200 (`76c8c99a8` "fix(lab): reuse immutable prepared source views") changed `crates/homeboy-lab-runner/src/lab/offload/hydration.rs` and merged green.

In its Test job log, `hydration.rs` appears **only in compiler warnings** — no `homeboy_lab_runner` test binary ever ran. The gate's own verdict:

```
##[notice]Differential gate accepted review test: current=1 base=1
```

Both sides' single failure was the same unrelated pre-existing test (`aggregate_repo_artifact_next_action_runs_outside_a_checkout`), so the differential passed.

That commit shipped two regressions, both covered by unit tests **in the crate it changed**, neither of which was executed:

- `hydration_skips_when_linked_extensions_do_not_provide_dependencies` (fails `RunnerNotFound: unused-runner`)
- `hydration_runs_for_materialized_workspace_jobs` (fails `Invalid argument 'workspace_root'`)

Both were still failing on `main` a full day later and were only found by hand while fixing an unrelated red release (fixed in #10431).

## Expected contract

- A change under any monorepo member subtree (`crates/<member>/src/**`, `packages/<pkg>/src/**`, ...) is recognized as source.
- Such a change either selects the tests that reference it, or fails closed via `#8340`.
- Root-package sources keep working; docs/config changes stay non-source.
- The rule stays language-agnostic — driven by configured directory names, not build-system knowledge.

## Note on `baseline_red`

`baseline_red` (warning-only in `enforce-final-status.sh`) is a *separate* softening worth reviewing on its own, but it is **not** what let these regressions through. The count-based differential and the scope gap above are.

chubes4:
Fixes #10449.

You asked why the differential Test gate didn't block the regressions. It wasn't `baseline_red` — I was wrong about that in the earlier thread. The real cause is worse.

## The gate cannot see `crates/**`

The Rust extension declares `drift.source_dirs: ["src"]`, and `glob_patterns` turned that into one root-anchored pattern, `src/**/*.rs`. `glob::Pattern::matches_path` is anchored, so:

| path | matches `src/**/*.rs` |
|---|---|
| `src/lib.rs` | yes |
| `crates/homeboy-core/src/daemon/stop.rs` | **no** |
| `crates/homeboy-lab-runner/src/lab/offload/hydration.rs` | **no** |

Since homeboy moved into `crates/*`, only the thin root package is visible to the scope engine. Two protections fail together, which is what makes it silent rather than noisy:

1. **Selection** — `detect_drift_for_changed_files` filters `prod_files` through the source patterns, gets an empty list, and returns early with zero drifted tests.
2. **Fail-closed** — `source_changes_without_tests` uses the same patterns, so the change never reaches the `#8340` guard that is supposed to fail closed on "source changed, zero tests selected".

So the gate ran no tests for the changed crate *and* reported that as legitimate.

## Proof from the actual merge

PR #10200 (`76c8c99a8`) changed `hydration.rs` and merged green. In its Test job log, `hydration.rs` appears **only in compiler warnings** — no `homeboy_lab_runner` test binary ran at all. The gate's verdict:

```
##[notice]Differential gate accepted review test: current=1 base=1
```

Both sides' one failure was the same unrelated pre-existing test, so the differential passed. That commit shipped the two regressions I fixed in #10431 — and both are covered by unit tests **inside the crate it edited**, which simply never ran. They stayed red on `main` for a day and were found by hand, not by CI.

## The fix

Emit an any-depth pattern alongside the anchored one for each configured directory (`src/**/*.rs` **and** `**/src/**/*.rs`).

Deliberately not Cargo-aware: it is driven entirely by the configured directory names, so it encodes no build-system knowledge (respecting the `core-agnostic-source` policy) and fixes every `<member>/src` monorepo layout — Rust crates, JS packages, PHP packages — rather than special-casing Cargo.

## Verification

Both new tests fail without the pattern change and pass with it — I flipped the fix out and back to confirm neither is vacuous:

| test | covers |
|---|---|
| `workspace_member_sources_are_source_relevant` | relevance / fail-closed half |
| `drift_selects_tests_for_workspace_member_sources` | selection half — asserts the referencing test is actually chosen |

Plus `root_package_sources_remain_source_relevant` and `documentation_and_config_remain_non_source` to pin that this widens coverage without dragging docs/config in.

```
cargo test -p homeboy-extension --lib test::     127 passed, 0 failed
cargo test -p homeboy-extension --lib            605 passed, 10 failed
```

Those 10 are pre-existing and environment-dependent on my host (toolchain path, provider discoverability, lab snapshot evidence) — identical set on unmodified `main`, which I verified by stashing. Net effect of this PR locally is 601→605 passed, same 10.

I also updated the two tests that pinned the exact pattern list, since the emitted set is the thing that changed.

## Tradeoff to weigh

This makes the gate stricter, and that is the point — but it is a real behavior change for **every** component, not just homeboy. Expect some PRs that previously passed on an empty selection to now either select tests or fail closed with the `#8340` message. That is the safe direction, and the Rust extension already widens to the full suite when a derived scope executes zero tests, so I don't expect a wall of red. Worth your call on timing, since the blast radius is fleet-wide.

Two follow-ups I did **not** touch:

- **`baseline_red` is warning-only** in `enforce-final-status.sh` — never sets `FAILED`. Separate softening, worth reviewing on its own merits.
- **The differential compares failure *counts*, not identities.** A PR that fixes one pre-existing failure and introduces a different one nets zero and passes.

### DIFF
--- a/crates/homeboy-extension/src/test/drift.rs
+++ b/crates/homeboy-extension/src/test/drift.rs
@@ -64,16 +64,30 @@ impl DriftOptions {
     }
 }
 
+/// Build match patterns for one set of configured directories.
+///
+/// Each directory yields two patterns: one anchored at the component root and
+/// one that matches the same directory at any depth. A monorepo keeps its real
+/// code in per-member subtrees — `crates/<member>/src`, `packages/<pkg>/src`,
+/// and so on — so a root-anchored `src/**` alone matches only the thin root
+/// package. Everything else silently reads as "not source", which disables both
+/// drift-based test selection and the `#8340` zero-selection guard that is
+/// supposed to fail closed. Matching the directory name at any depth keeps this
+/// language-agnostic: it is driven entirely by the configured directory names,
+/// with no build-system knowledge.
 fn glob_patterns(dirs: &[String], extensions: &[String]) -> Vec<String> {
     dirs.iter()
         .flat_map(|dir| {
-            extensions.iter().map(move |extension| {
+            extensions.iter().flat_map(move |extension| {
                 let extension = extension.trim_start_matches('.');
                 if dir.is_empty() || dir == "." {
-                    format!("**/*.{}", extension)
-                } else {
-                    format!("{}/**/*.{}", dir.trim_end_matches('/'), extension)
+                    return vec![format!("**/*.{}", extension)];
                 }
+                let dir = dir.trim_end_matches('/');
+                vec![
+                    format!("{dir}/**/*.{extension}"),
+                    format!("**/{dir}/**/*.{extension}"),
+                ]
             })
         })
         .collect()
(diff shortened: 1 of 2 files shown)


chubes4:
## CI evidence from this PR's own Test run

This PR changes how scope is selected, so its own Test job is the experiment. All six checks pass, and the scope demonstrably widened:

| | test binaries executed | Test job duration |
|---|---|---|
| PR #10200 (the merge that shipped the regressions) | **7** | 18m50s |
| this PR | **23** | 40m47s |

Before, the selection reached only `tests/agent_task_promotion_provider.rs`, `agent_task_provider_flag_diagnostic.rs`, `agent_tool_dispatch.rs`, `cleanup_next_actions.rs`. Now it reaches 20 integration targets including `reverse_cook_queue_acceptance.rs`, `cook_lab_handoff_test.rs`, `release_workflow_test.rs`, and `deps_test.rs` — despite this PR touching *fewer* files than #10200 did.

The gate is green, so this widening did not produce a wall of red.

## Honest limitation — this is a partial fix

Crate-level **inline** unit tests still do not run for a changed crate source. `homeboy_lab_runner --lib` / `homeboy_extension --lib` do not appear in this run either.

Cause: `DriftOptions::from_config` drops `TestDriftConfig.inline_tests` entirely — it is never stored on the struct, so nothing downstream can act on it. Drift maps a changed source only to *separate* test files that reference its symbols; a `#[cfg(test)] mod tests` in the same file is never selected.

That is exactly where the two `76c8c99a8` regressions lived, so this PR alone would **not** have caught them — it would have caught them only via the `#8340` fail-closed guard if selection had come back empty, which it doesn't here.

The routing machinery to fix this already exists (`rust_cargo_changed_routing_emits_inline_filter`, `..._maps_repository_relative_inline_test_to_owning_package`); the missing link is selecting the changed source file itself as a test file when `inline_tests` is true. Filed separately rather than bundled here, since it is a second behavior change with its own blast radius.

Issue #10326, opened 2026-07-27
Epic: tech-debt burndown from the b937def60 investigation (57 findings)


chubes4:
Tracking epic for a full-surface investigation of `homeboy@b937def60` and `homeboy-extensions@v3.34.4`. Every finding below carries verified `file:line` evidence in its own issue. Scope measured: **43 workspace crates** (`crates/*` 24 + `crates/contracts/*` 18 + root), 831k LOC Rust, ~600 live CLI commands.

Filed: **48 here + 9 in homeboy-extensions**. Existing related work is cross-referenced per issue rather than duplicated.

---

## Wave 0 — unblock. Must land alone, first.

Everything downstream is unmeasurable until this lands, and widening audit scope against a dead baseline detonates every PR.

- #10278 — **audit baseline is 98.9% dead paths** (448/453 fingerprints reference `src/core/...` which no longer exists) and `core-agnostic-source` scans 4 of 43 crates
- #10279 — policy misses the toolchain-output and org/product term classes *(blocked by #10278)*

## Wave 1 — P0 correctness. All independent, parallelizable.

- #10280 — **reviewer-URL SSRF guard is compiled out**; a 683-LOC orphan directory forked the validator and the live copy accepts `https://127.0.0.1/…`
- #10281 — `resolve_artifact` iterates a `HashMap`, so multi-extension components get a **different deploy artifact per process run**
- #10282 — **regression against shipped #7787/#7804**: `--notification-route` is dropped across `thread::scope` for every fanout child
- #10283 — `fanout cook-batch` silently runs every cook on the controller; `--placement lab` errors while `--runner` works
- #10284 — directory artifacts have no rollback and the artifact root is never reconciled filesystem-side
- #10285 — `artifacts` table has no `run_id` index, and retention is an N+1 that reopens the store per run
- #10286 — `worktree cleanup` is destructive by default while all 8 siblings are `--apply`-gated
- #10287 — `merge_lab_runner_workload_capability_preflight` is a no-op; possible silent capability-gating gap *(investigate before acting)*
- #10288 — `runtime controller-prune` ignores the configured retention window
- #10289 — `stack apply` aborts the cherry-pick, then says to resolve it manually
- #10290 — `deploy --check` and `harvest --check` use two tree-hashers with different exclusion policies

## Wave 2 — pure deletion and mechanical consolidation.

- #10291 — delete the dead deterministic-loop framework (**1,473 LOC, zero production callers**)
[...]

Issue #10298, opened 2026-07-27
Normalize #[path] smuggling: 8 sites hide production code from directory-based tooling


chubes4:
## Problem

88 `#[path]` sites exist. ~80 are legitimate — mounting `tests/core/**` / `tests/commands/**` under `#[cfg(test)]`. The remainder mount **production** source from another directory:

| Site | Mounts | Verdict |
|---|---|---|
| `homeboy-agents/src/agent_task_lifecycle/mod.rs:27-28` | `../lifecycle_store.rs` (445 LOC) | Smuggling — parent-directory file pulled into a subdirectory module, then aliased `use lifecycle_store as store;` (`:30`). Not declared in `lib.rs` at all. |
| `homeboy-lab-runner/src/connection.rs:44-45` | `connection_daemon.rs` | Smuggling — file sits at `src/` while `src/connection/` also exists; the module's files are split across two directories. |
| `homeboy-lab-runner/src/connection.rs:51-52` | `connection_stop_transport_recovery.rs` | Same split. |
| `homeboy-code-audit/src/detectors/test_topology.rs:14-15` | `../test_quality.rs` | Smuggling — **not** `cfg(test)`-gated. |
| `homeboy-core/src/observation/records.rs:6` | `finding_records.rs` | Resolves to `observation/finding_records.rs` while its five siblings (`:8-12`) resolve to `observation/records/*.rs`. Compiles; traps the next reader. |
| `homeboy-review/src/review/render.rs:30` | `render_audit.rs` | Smuggling (sibling). |
| `homeboy-core/src/deps.rs:7-8` | `deps_provider.rs` (1,642 LOC) → `mod provider` | Smuggling + rename. Crate-root file remounted as `deps::provider`; the only `homeboy-core/src/` file absent from `lib.rs`. |
| `homeboy-core/src/defaults.rs:5` | `defaults/builtins.rs` | **Redundant** — exactly equivalent to plain `mod builtins;`. Its two siblings at `:7-8` prove it. |
| `homeboy-core/src/engine/temp/implementation.rs:484` | `implementation/storage.rs` | **Redundant** — equivalent to plain `mod storage;`. |

## Why it matters

[...]

PR #10554, opened 2026-07-28
refactor(cli): delete GlobalArgs and the dead ops descriptor spec column

### LINKED ISSUE #10322: Delete GlobalArgs: an empty struct threaded through 68 handler signatures
## Problem

```rust
// crates/homeboy-cli/src/commands/mod.rs:77
pub struct GlobalArgs {}
```

Zero fields. Threaded as `_global: &GlobalArgs` (underscore-prefixed = provably unused) through **68 handler signatures**, plus ~30 forwarding sites (`adapter.rs:12,59,73`, `json_output/{mod,ops,quality,workspace}.rs`, `output_runtime.rs:459,537`) and 100+ `&GlobalArgs {}` constructions in tests. It is baked into the descriptor macro's handler type (`command_contract/descriptors.rs:12-26`) and into `JsonCommandExecutor` (`adapter.rs:12`).

Related dead ceremony in the same file: `merge_json_sources` (`mod.rs:141-170`) binds `let base = …; Ok(base)`, and the `spec: Option<&str>` branch is the only real work.

## The real finding underneath

**Three global values are passed via `AtomicBool`/process-global setters while an empty struct is passed by reference:** `set_skip_deps_hydration` (`commands/mod.rs:74-85`), `set_artifact_root_override` (`cli_runtime.rs:353`), and the notification route thread-local. If a genuine global-context bag is wanted later, those are its fields.

## Fix

Delete the type and the parameter. Mechanical `sed`-scale, gated only on the descriptor-macro consolidation landing first so the signature is edited in one macro row instead of 15.

~200 LOC.

Related: #7516.

### LINKED ISSUE #10323: Two macros duplicate 15 command spec expressions verbatim, and three dispatch tables exist only to unreachable!() on each other
## Problem — two halves of one consolidation

**1. `ops_command_descriptors!` and `ops_command_spec!` duplicate 15 rows.**
`crates/homeboy-cli/src/command_contract/descriptors.rs` defines the same 15 command rows twice with **byte-identical `$spec` expressions** — `:12-26` (`ops_command_descriptors`, taking module/Variant/Args/SPEC/handler) and `:40-54` (`ops_command_spec`, taking module → SPEC). E.g. `:18` vs `:46`:
```rust
command_spec_with_safety("deploy", CommandJsonFamily::Ops, operator_safety(Some("--dry-run"), DEPLOY_DANGEROUS_FLAGS))
```
The doc at `:31-37` states the reason: `spec.rs` can't see `crate::commands`. Consequence: any safety-metadata edit must be made in two places or the manifest and dispatcher disagree, and nothing enforces the pairing.

*Fix:* give `ops_command_descriptors!` a second consumer arm emitting only the `$spec` field, expanded from `spec.rs` — the `crate::commands` paths appear only in the `$args`/`$handler` positions, which a spec-only consumer drops. Delete `ops_command_spec!`. **-17 LOC.**

**2. Three near-identical dispatch tables enforced by runtime panics.**
- `json_output/workspace.rs:9-30` — 17 arms, each `Commands::X(args) => map(x::run(args, global))`, then `:28 _ => unreachable!("command routed to wrong JSON output family")`
- `json_output/quality.rs:6-14` — 4 arms, same shape, same panic at `:12`
- `json_output/ops.rs:13-24` — 15 arms via macro, same panic at `:18`
- selected by `spec.dispatch_family()` at `json_output/mod.rs:314-321`

The only thing "family" buys the dispatcher is choosing which of three identical `variant → handler` tables to consult, enforced by three runtime panics **instead of the type system**. Two more `unreachable!` sites in the same routing at `contract_output_routing.rs:110-112`.

*Fix:* extend the unified macro into a single `command_descriptors!` covering all 40 variants and generate **one exhaustive dispatch match**. Exhaustiveness becomes compiler-checked and all five `unreachable!` sites become impossible states. `json_family` stays on `CommandSpec` for the manifest/output contract but stops being a dispatch key. **-60 LOC, -5 latent panics.**

Part 2 sequences behind the duplicate-enum collapse.

Related: #7516, #7456.

chubes4:
Closes #10322. Closes #10323 (part 1; part 2 deliberately deferred — see below).

Three commits, each independently reviewable:

1. `refactor(cli): delete the fieldless GlobalArgs handler parameter` — #10322
2. `refactor(cli): drop the dead spec column from ops_command_descriptors` — #10323 part 1
3. `style(cli): rustfmt the GlobalArgs edits cargo fmt cannot reach` — fallout, plus a tooling finding

One PR rather than two because commits 1 and 2 both edit `commands/mod.rs` and `json_output/ops.rs`; splitting them would guarantee a rebase conflict between my own branches.

---

## #10322 — `GlobalArgs` is inert. Proof.

The issue asked whether deleting it changes the accepted argv surface or the rendered `--help`. **It cannot**, and here is the evidence rather than an assertion:

| check | result |
|---|---|
| definition | `pub struct GlobalArgs {}` — zero fields, **zero attributes**, no `#[derive]` at all |
| derives `clap::Args`? | no — so no `FromArgMatches`/`CommandFactory`/`Args` impl exists |
| `impl` blocks | `grep -rn "impl .*GlobalArgs"` → none |
| `#[command(flatten)]` | 148 flatten sites in the crate, **zero** name `GlobalArgs` |
| field of any struct | `grep -rn ": *GlobalArgs"` → only `fn` parameters, never a field |
| referenced from `cli_surface::Cli` | no |
| how it was constructed | `let global = GlobalArgs {};` in `cli_runtime::run_matches`, *after* `ArgMatches` — built from nothing, so it demonstrably carried no argv-derived state |

Clap only knows about types reachable from `Cli`'s `Parser`/`Subcommand`/`Args` derives. `GlobalArgs` implements none of them and appears nowhere in that tree, so it contributes **zero bytes** to rendered help. The real process-global CLI state lives on `Cli` as fourteen `global = true` clap arguments.

Because rendered help *is* a wire protocol here — `homeboy-lab-runner::negotiate_leaseless_recovery_contract` parses a remote binary's `Options:` block to negotiate the lease-less recovery capability handshake (the thing #10536 nearly broke) — this PR adds `cli_surface::tests::root_global_flag_surface_is_pinned`, which pins the exact set of globally-propagated long flags so any future change to that surface is a deliberate, reviewed protocol change.

### Scope: it was not 68 signatures, it was 98

| kind | count |
|---|---|
| function definitions taking `&GlobalArgs` | **98**, across 67 files |
| function-type declarations | 3 (`JsonCommandExecutor`, `BoundCommandAdapter::run` box, `ReviewStageDescriptor::run` fn-pointer field) |
| call sites passing `&GlobalArgs {}` / forwarding `global` | ~140 |

The issue's 68 is close to the count of *underscore-prefixed* `_global:` occurrences (69 lines); it misses the ~29 non-underscore `global:` parameters that actually forwarded the reference down a dispatch chain.

## #10323 — the issue's proposed fix is backwards, and the duplication is *dead*

The issue proposes: *"give `ops_command_descriptors!` a second consumer arm emitting only the `$spec` field, expanded from `spec.rs` — Delete `ops_command_spec!`."*

Two problems.

**1. It is not possible as stated.** `ops_command_descriptors!` expands to a single `$consumer! { row, row, ... }` block. `spec.rs` splices the fifteen ops rows into a hand-ordered `COMMANDS: &[CommandSpec]` array at **fifteen non-contiguous positions** (lines 1099, 1100, 1137, 1142, 1147–1150, 1164, 1165, 1178, 1194, 1309, 1314, 1315), interleaved with non-ops entries. A block-emitting consumer cannot be spliced into fifteen separate slots. `ops_command_spec!`'s per-name arms exist precisely because of that.

**2. The direction is inverted — the descriptor macro's `$spec` column is the dead copy.** `ops_command_descriptors!` has exactly two expansion sites in the entire repo:

- `commands::register_ops_command_modules` → `$(pub mod $module;)*` — uses `$module` only
- `json_output::ops::registered_ops_dispatch` → `$(Commands::$variant(args) => map($handler(args)),)*` — uses `$variant` and `$handler` only

**Neither consumer ever references `$args` or `$spec`.** macro_rules metavariables are inert token trees until substituted, so those fifteen `CommandSpec` expressions expanded to *no code whatsoever*. They were pure maintenance load — the "edit it in two places or the manifest and dispatcher disagree" hazard, with nothing gained.

So the fix is deletion, not a second consumer arm: each row is now `(module, Variant, handler)`, and the ops `CommandSpec` lives in exactly one place (`ops_command_spec!`). Bonus: the rows now fit on readable lines, which matters because **rustfmt does not format `macro_rules!` bodies** — the old rows were 200+ chars and permanently outside the formatter's reach.

Nothing is lost by dropping the co-location, because the pairing was never enforced by co-location anyway. It is enforced by `cli_surface::tests::command_registry_manifest_and_docs_metadata_align`, which asserts a *bijection* between the clap subcommand set and `COMMAND_SPECS` — a stronger guard covering all 40 commands, not just the ops 15. This PR adds `every_ops_descriptor_is_registered_in_the_ops_json_family` as a targeted second guard.

Also: the issue says there are "two more `unreachable!` sites in the same routing at `contract_output_routing.rs:110-112`". There is one (line 111).

### #10323 part 2 (three dispatch tables → one exhaustive match): deferred, with a reason

I costed this out. The union really is exhaustive — 3 adapter-routed (`Fleet`/`Observe`/`Contract`) + `Deps` + 15 ops + 4 quality + 17 workspace = 40 = every `Commands` variant — so the refactor would be almost entirely compiler-verified (a mis-wired arm is a type error, since every variant carries a distinct Args type). It is genuinely attractive: it deletes three `unreachable!()` panics and makes exhaustiveness a compiler obligation.

**But it cannot be done well right now, and the reason is mechanical.** `crate::ops_command_descriptors!(consumer)` expands to a whole expression, not to a fragment of a `match`, so macro-generated arms cannot be interleaved with literal arms. To get one exhaustive match you must move all 25 hand-written arms *inside* a `macro_rules!` consumer body — and rustfmt does not format `macro_rules!` bodies. That trades three `unreachable!()` for 25 production dispatch arms permanently exiled from the formatter, which this PR demonstrates is not hypothetical (see below). The alternative — keeping the arms formatted — forfeits the exhaustiveness that was the entire point.

The issue already sequences part 2 behind the duplicate-enum collapse; this is the concrete reason that sequencing is right. Doing it properly needs the descriptor macro replaced by something that can emit match arms (a proc macro or a `Commands`-derived table), which is #10326's territory.

## Commit 3 — a tooling finding worth its own issue

**`cargo fmt --all` does not visit the fifteen ops command modules.** They are declared by `$(pub mod $module;)*` inside `register_ops_command_modules!`, and rustfmt does not expand macros, so `commands/{ssh,server,db,file,logs,triage,deploy,harvest,daemon,schedule,status,git,self_cmd,api,upgrade}.rs`, everything beneath them, and the `#[cfg(test)] #[path = "../../tests/…"]` modules they pull in are all invisible to the formatter.

Verified empirically: appending `fn __fmt_probe(){let x=1;}` to `daemon.rs` survives `cargo fmt --all` completely unformatted, while standalone `rustfmt --check daemon.rs` reports 11 diff blocks.

Those files therefore carry pre-existing drift on `main` today (daemon.rs 11 blocks, deploy.rs 14, status/mod.rs 5, ssh.rs 3, triage.rs 3, logs.rs 2, schedule.rs 2, upgrade.rs 1, tests/commands/deploy_test.rs 5). **This PR does not sweep that up.** Commit 3 only formats the call sites the `GlobalArgs` removal reflowed, verified by comparing per-file `rustfmt --check` block counts against a clean `origin/main` worktree — every file is `<=` main, and `tests/commands/deploy_test.rs` goes 5 → 0 because my edit reflows those exact calls anyway.

This is a close cousin of #10298 (`#[path]` smuggling hides production code from directory structure): here macro-declared modules hide production code from the formatter. I did not touch the eight `#[path]` sites; noting the overlap for #10298.

## What I could not verify

Per repo policy I did not run `cargo build`, `cargo test`, `cargo check`, or `cargo clippy` — CI is the gate. What I *did* verify locally:

- **Syntax**: `rustfmt --check` parses all 93 modified files without error (control-tested: rustfmt does report `error:` on a deliberately broken file).
- **No dangling references**: `grep -rnw global` and `grep -rn GlobalArgs` across the repo return zero hits outside three unrelated string fixtures in `homeboy-code-audit/src/import_matching.rs` (which test generic import-string parsing, not the type).
- **No arity mismatch is possible**: the removed parameter was *always* spelled `global`; with zero remaining occurrences of that identifier, no call site can still be passing it. Every higher-order use (3 fn-type declarations, the `TypedCommandAdapter::json_only` test closure) was updated by hand.
- **No new import warnings**: every rewritten `use` list was checked for remaining-name usage (the one flagged case, `runs/mod.rs`'s `CmdResult`, is re-exported to child modules via `super::` — a false positive).

Not verified: type checking, borrow checking, and actual test outcomes. In particular the two new tests are written blind. `root_global_flag_surface_is_pinned` filters clap's generated `help`/`version` out of the global set defensively, in case clap marks them global.


chubes4:
### CI status and what the Test gate actually reported

| gate | result |
|---|---|
| Rustfmt | pass |
| Workspace Tests Compile | pass |
| Windows Compile | pass |
| Lint (clippy) | pass |
| Audit | pass |
| Test | **timed out** — see below |

**The Test gate did not report a test failure. It was killed on a wall clock.** From the run's observation artifact (`runs.json` → `metadata_json.child_supervision`):

```
status:              interrupted
exit_code:           124
cancellation_reason: timeout
started_at:          2026-07-28T07:05:17Z
finished_at:         2026-07-28T07:30:17Z
stderr_tail:         Homeboy command timed out after 1500000ms;
                     terminated child process group before returning failure evidence
changed_since:       null      <- whole suite, not a scoped subset
test_counts:         null
```

25 minutes, whole suite, SIGKILL'd child process group. The nine `... FAILED` lines visible in the truncated stdout tail before the kill are all in **`homeboy-agents`**:

```
agent_task_dispatch_plan::tests::{dispatch_merges_global_settings_into_executor_config,
                                  dispatch_plan_accepts_component_contracts_from_provider_config,
                                  dispatch_plan_preflight_fails_on_unresolved_runtime_dependency}
agent_task_finalization::tests::production_validator_accepts_only_the_exact_promoted_recovery_commit
agent_task_gate::baseline_tests::{bounded_baseline_gate_is_cancelled,
                                  bounded_baseline_gate_reaps_background_descendants_before_reader_join}
agent_task_gate::tests::existing_gate_commands_are_automatic_toolchain_requirements
agent_task_lifecycle::activity_provider::tests::{probe_by_id_resolves_one_record_without_scanning_or_writing,
                                                 show_activity_resolves_an_agent_task_id_through_the_indexed_probe}
```

**This branch cannot affect any of them.** It touches exactly two crates — `homeboy-cli` and `homeboy-code-audit` — plus repo-root `tests/` files that are `#[path]`-included only into those two. `crates/homeboy-agents/Cargo.toml` depends on `homeboy-core`, `homeboy-extension`, `homeboy-error`, `homeboy-engine-primitives`, and the contract crates; it does not depend on `homeboy-cli` or `homeboy-code-audit`. There is no edge from anything I changed to those tests.

The failing set is also exactly the shape you would expect from a process-supervision suite dying alongside a killed process group: bounded-gate cancellation, background-descendant reaping, PID/lease probes, and several tests the log flags as `has been running for over 60 seconds`.

I have re-run the failed job. If it times out again, the suite simply does not fit in the 1500s budget for a 94-file diff (the gate falls back to `changed_since: null`, i.e. the full workspace) — that is a gate-budget question, not a signal about this change. The compile-bearing gates that *do* cover this branch — Workspace Tests Compile, Windows Compile, and clippy — are all green, which is what actually validates a mechanical parameter removal.

chubes4:
**Update: all six gates green.** The re-run of the Test gate passed in 40m07s — confirming the earlier failure was the 1500s child-supervision timeout (`exit_code: 124`, `cancellation_reason: timeout`) on a full-suite run, not a regression. Rustfmt / Workspace Tests Compile / Windows Compile / Lint / Audit / Test all pass; `mergeStateStatus: CLEAN`.

Worth noting for the gate owner: a 94-file diff makes the Test gate fall back to `changed_since: null` (whole workspace), and the whole workspace lands within a few minutes of the 25-minute budget — 40m07s on the passing run means the job-level budget and the child-supervision budget are not aligned. Large mechanical PRs will keep tripping this nondeterministically.

Issue #10315, opened 2026-07-27
activity watch and runs watch are the same command implemented twice


chubes4:
## Problem

Two independent poll loops with two copies of the same constant and identical output field sets:

| | `activity watch` | `runs watch` |
|---|---|---|
| arg | `<ID>` — "Activity id, **observation run id**, agent-task run id, or runner job id" | `<RUN_ID>` — observation run id |
| flags | `--timeout`, `--interval` [2s], `--notify` | `--timeout` [5m], `--interval` [2s], `--notify`, `--forever` |

- `crates/homeboy-cli/src/commands/activity.rs:16` — `const TIMEOUT_EXIT_CODE: i32 = 124;` · loop at `:204-279`
- `crates/homeboy-cli/src/commands/runs/watch.rs:33` — `const TIMEOUT_EXIT_CODE: i32 = 124;` · loop at `:117-260`

Same output fields (`terminal`, `timed_out`, `waited_secs`, `poll_count`, `notify`, `_homeboy_actionable`): `activity.rs:77-97` vs `runs/watch.rs:57-76`.

The activity module even points users at the other one: `activity.rs:370` — `watch_command: format!("homeboy runs watch {run_id}")`.

Two more watch loops exist with a third flag vocabulary: `commands/triage.rs:84-90` (`--timeout`/`--poll-interval`) → `crates/homeboy-triage/src/watch.rs:174`, and `commands/runner/jobs.rs:130,177`.

## Fix

`runs/watch.rs::run_watch_loop` (`:145-186`) is **already generic** over poller/sleep/clock and unit-tested (`:334-392`). Make `activity::watch` construct an `ActivityPoller: RunPoller` and call it; delete `activity.rs:204-279`. Align `triage`'s `--poll-interval` → `--interval`, keeping `--poll-interval` as a `visible_alias` (the pattern already used at `deploy.rs:53`).

~90 LOC.

## Related gap worth folding in

Neither watch accepts more than one id, so orchestrating N tasks means N watcher processes or a hand-rolled poll loop. Extend `ActivityWatchArgs` to accept repeated ids and a `--batch <fanout_id>` selector resolving through `agent_task_batch::status`; `ActivityCollector` (`core/src/activity/collector.rs`) already dedupes multi-source items.

Related: #10267, #6781.

chubes4:
Corrections from #10587, which implements this.

[...]

Issue #10312, opened 2026-07-27
Promote core::scope::Scope to a clap arg group — every command reinvents entity selection


chubes4:
## Problem

`homeboy_core::scope::Scope` **already is** the single entity primitive:
```rust
// crates/homeboy-core/src/scope.rs:17-29
pub enum Scope { Component(String), Project(String), Fleet(String), Rig(String), Workspace, Path { path, component_id } }
// :65-74  pub enum ScopeKind { ... }   // same six
```

The CLI uses it in **three** places, all as `Scope::Project(..)` only (`status/mod.rs:263`, `release/mod.rs:873`, `build.rs:111,148`). Everywhere else the same six-way choice is re-declared.

Most damning: `triage` declares it **twice inside one command** — once as subcommands (`crates/homeboy-cli/src/commands/triage.rs:93-116`) and once as five mutually-exclusive flags with `conflicts_with_all` against each other (`:138-156`).

At least four incompatible component-selection shapes coexist:
| Shape | Site |
|---|---|
| `component: Option<String>` + `--path` | `commands/utils/args.rs:280-288` |
| `-c/--component` (Append) + `--components` (delimited) + `--path` | `commands/refactor.rs:293-306` |
| positional `target_id` + `component_ids: Vec` + `--all` + `--json` | `commands/build.rs:19-26` |
| `target_id` + `component_ids` + `-p` + `-c` + `--json` + `--all` + `-f` + `-s` | `commands/deploy.rs:25-66` |

Because there is no shared shape, **the Lab offload path re-parses argv as strings** to normalize it — `commands/infra/route.rs:1817-1868 strip_component_target_args()`, matching on the literals `"-c"`, `"--component"`, `"--components"`, `"--components="`.

## Fix

Promote `Scope` to a `#[derive(Args)] ScopeArgs` in `commands/utils/args.rs` (next to the existing `PositionalComponentArgs`/`BaselineArgs`/`SettingArgs` groups) with `resolve() -> Scope`. Flatten into `triage`, `deploy`, `build`, `refactor`, `review`, `cleanup`, `status`, `release`.

Delete `RefactorTargetArgs`, `triage`'s duplicate scope flags, and `strip_component_target_args` (`route.rs` can construct `Scope::Path{..}` directly).

**No new contract** — `Scope`/`ScopeKind` are already public, serialized, and documented.

~450 LOC. Unblocks the `route.rs` argv-surgery cleanup.

Related: #7456, #6779.

[...]

PR #10587, opened 2026-07-28
refactor(cli): one watch loop, and pin the ScopeArgs boundary

chubes4:
Refs #10312, #10315.

## Verdict up front

- **#10315: `activity watch` and `runs watch` are NOT the same command.** They poll different stores, return different domain types, emit different output schemas, and differ deliberately in reconcile behaviour. What *was* duplicated is the poll loop, the `124` timeout constant, and the terminal/timeout conclusion shape. That is extracted; both commands, both names, and both wire shapes survive intact.
- **#10315's prescribed fix is not implementable as written** — see below.
- **#10312: all five remaining commands stay deferred**, with two of #10510's reasons upgraded to stronger ones. The reasons are now executable tests instead of prose in a merged PR body, plus a structural pin that stops a *new* hand-rolled six-way selector from appearing.
- **One live defect fixed** (the one #10538 surfaced): `activity watch --notify` silently ignored `--notification-route`.

---

## #10315 — evidence that they are not duplicates

| | `activity watch` | `runs watch` |
|---|---|---|
| resolver | `activity::resolve_activity` — a projection across the observation store, the agent-task store, runner sessions, and daemon jobs (`activity.rs:125-132` matches on `item.id`, `refs.run_id`, `refs.agent_task_run_id`, `refs.runner_job_id`) | `runs_service::require_run` — observation store, mirrored runner runs, run labels (`runs_service/run_lookup.rs:5-18`) |
| item type | `ActivityItem` / `ActivityState` (`= RunLifecycleStatus`) | `RunRecord` / `RunStatus` |
| output | `activity.watch` — `schema`, `id`, `state`, `item`, `next_actions` | `runs.watch` — `run_id`, `status`, `run` |
| reconciles on poll | **no**, deliberately | **yes**, deliberately |

The decisive one: **`homeboy runs watch <runner-job-id>` cannot work.** `require_run` never consults the runner-session, agent-task, or daemon-job sources, so it returns `missing_run_error`. `activity watch` is the only surface that can watch those ids. This is why `activity_run_ref` points observation runs at `homeboy runs watch` (`activity.rs:375`) while `activity_job_ref` points runner jobs at `homeboy activity watch` (`activity.rs:385`) — the issue read that as "the module points users at the other one", but it is a division of labour, not an accident.

The non-reconciling read is not incidental either: `activity_reads_do_not_reconcile_stale_running_records` asserts that a `watch` over a stale-owner run leaves the record byte-identical. Collapsing activity onto `StorePoller` would have broken that test and changed observable behaviour.

### The prescribed fix could not be taken literally

> Make `activity::watch` construct an `ActivityPoller: RunPoller` and call it.

`RunPoller::poll` returned `homeboy::core::Result<RunRecord>`. An activity item frequently has **no `RunRecord` behind it at all** (agent-task runs, runner jobs, controller jobs). So `ActivityPoller` could not implement `RunPoller`.

What is implementable — and what this PR does — is to make the loop generic over the polled item:

```rust
pub trait WatchPoller {
    type Item;
    fn poll(&self, id: &str) -> homeboy::core::Result<Self::Item>;
    fn is_terminal(&self, item: &Self::Item) -> bool;
}
```

`crates/homeboy-cli/src/commands/utils/watch.rs` owns the loop, `WatchConfig`, `WatchConclusion`, `WatchResult<T>`, and the single `TIMEOUT_EXIT_CODE = 124`. `runs/watch.rs::StorePoller` and `activity.rs::ActivityPoller` supply only what is genuinely theirs. `runs/watch.rs` loses `RunPoller`, its private `WatchConfig`/`WatchConclusion`/`WatchResult`/`run_watch_loop`, and its copy of the constant; `activity.rs` loses its hand-rolled loop and its copy of the constant.

### A divergence the issue did not mention: the two duration parsers are not interchangeable

`runs::common::parse_duration` **rejects zero** (`"duration amount must be greater than zero"`, `common.rs:105-112`). `activity::parse_duration` **accepts it** — and activity's own test depends on that, calling `watch(... timeout: "0s", interval: "0s" ...)`.

So the parsers are deliberately left alone. Unifying on the runs parser would break an existing test and forbid `activity watch --interval 0s`; unifying on the activity parser would let `runs watch --interval 0s` spin-poll and would leak into the `--since` parsing that `runs gh-actions`/`query`/`drift` share. The loop is shared; the parsers are not, and the reason is now in a comment.

### Live defect fixed

Confirmed the #10538 finding. `activity.rs` built its notification as:

```rust
NotifyEvent { ..., transport: None, route: None, payload: None }.with_payload(payload)
```

`notify::dispatch` falls back to `config.notifications.default_transport` when `event.transport` is `None`, and emits `--route` only when `event.route` is `Some`. So `homeboy activity watch <id> --notify --notification-transport X --notification-route Y` delivered to the *configured default* transport with **no route**. `--notification-route` was silently discarded by this command alone.

Fixed by reading `notification_route::current()` — the thread-local `cli_runtime.rs:562` binds for the whole invocation from `--notification-transport`/`--notification-route` or `HOMEBOY_NOTIFICATION_*`.

`runs watch` is deliberately left as-is: it reads `NotificationRoute::from_metadata(&run.metadata_json)`, i.e. the route `ObservationStore::start_run` stamped onto the record at creation (`observation/store/runs.rs:86-87`). That is the stronger, run-scoped contract. `activity watch` cannot use it — `ActivityItem` is a cross-store projection with no metadata channel — so the caller's explicit route is the available and requested answer. Noted in the code.

### `triage --poll-interval` → `--interval`

Aligned as the issue asked, with `visible_alias = "poll-interval"` (the `deploy.rs:53` pattern). Purely additive: nothing that parsed before stops parsing. Tests cover both spellings, the unchanged `60s` default, and that they are one argument rather than two.

### Two "other watch loops" the issue listed — one is, one isn't

- **`triage --watch`** (`homeboy-triage/src/watch.rs:174`) is a real poll loop, but over the GitHub REST API against `--until <STATE>` targets with optional `--auto-merge`. It shares no store, no item type, and no exit-code contract. Only the flag *vocabulary* was worth aligning, which is what this PR did.
- **`runner job logs --follow`** (`runner/jobs.rs:130,177`) is **not a watch loop of this shape** and the issue is wrong to list it as one. It is a log-streaming follow loop with transport reconnection (`MAX_RECONNECTS`), cursor emission for exactly-once resume, and a cooperative shutdown handler. It has **no timeout at all**, takes `--poll-ms` in milliseconds, and has no terminal-state exit-code contract. Converging it would change user-visible behaviour, not remove duplication.

### Not folded in

The "related gap" (multi-id / `--batch` watching) is feature work with a new resolver dependency on `agent_task_batch::status`, not consolidation. Left for its own issue.

---

## #10312 — per-command status for the remaining five

`#10510` migrated `triage landing`, `build`, `status`. All five remaining are **confirmed deferred**. Two reasons are stronger than the ones #10510 gave.

| Command | Status | Why |
|---|---|---|
| **`deploy`** | deferred (confirmed) | Its selectors **combine** rather than exclude: `deploy --project P --component C` is a documented recipe (`DEPLOY_RECIPES`, `deploy.rs:17`), and the shared group makes exactly that combination a parse error. `--component` is `Option<Vec<String>>`, a multi-value target. Test: `deploy_combines_selectors_the_shared_group_makes_exclusive`. |
| **`refactor`** | deferred (confirmed, **stronger reason**) | #10510 said "multi-entity target", which is true (`-c` is `Append`, `--components` is delimited). The harder blocker is mechanical: `refactor` **already owns the `component` and `path` arg ids** through its flattened `PositionalComponentArgs`, and `ScopeArgs` declares both. Flattening would be a duplicate-id panic, not a semantic debate. Tests: `refactor_targets_many_components_at_once`, `refactor_already_owns_the_arg_ids_the_shared_group_needs`. |
| **`release`** | deferred (confirmed, **stronger reason**) | #10510 cited `--project` composing with `--outdated`. The real blockers are (a) its positional is `components: Vec<String>` — `release comp-a comp-b` is a *multi*-component target, so the group's singular `--component` would be a second, conflicting spelling of the same thing; and (b) `--project` carries the short `-p`, which the deliberately long-only group cannot supply, so migrating would silently delete a published short flag. Its `--path` is also a per-component checkout override (`path_override`, `release/mod.rs:418`) that *composes* with the positional, not a standalone scope. Test: `release_project_selector_keeps_a_short_flag_the_group_cannot_supply`. |
| **`cleanup artifacts`** | deferred (confirmed) | No entity selection exists. `--path: PathBuf` selects a *checkout root* for managed-worktree discovery; `--self` targets the source checkout that built the binary. Flattening would advertise five flags with nothing implementing them and change `--path`'s type. Test: `cleanup_artifacts_has_no_entity_selection_to_migrate`. |
| **`review`** | deferred (confirmed) | Already delegates to `PositionalComponentArgs`. There was never a six-way switch here to remove. Test: `review_selects_a_component_positionally`. |

### The structural finding

**The duplication #10312 was filed about is gone.** A full-tree walk finds exactly one command node that declares `--fleet` without the rest of the shared group: `deploy`. `hand_rolled_entity_selection_is_confined_to_deploy` asserts that offender list **by exact value**, so adding a new hand-rolled six-way selector — or migrating `deploy` — fails the test and forces a reviewed decision. `migrated_commands_carry_the_whole_shared_group` pins the other direction for `triage landing` / `build` / `status`.

### `strip_component_target_args`

Unchanged, and still not deletable — #10510 is right and the issue is wrong. `route.rs` must hand a *rewritten argv* to a remote runner and clap cannot re-serialize a parsed `Commands` back to argv. (It has also drifted from the issue's cited location: `route.rs:2140`, not `:1817`.)

### Other single-`--project` selectors are not in scope

`runner exec`, `runner dispatch`, `extension list`, and `release changes` each declare a lone `--project`. None is a six-way switch; giving them fleet/rig/workspace paths is feature work, not consolidation.

---

## `--help` handshake: explicitly NOT broken

**One CLI surface change in this PR: `triage`'s `--poll-interval` gains the primary spelling `--interval` plus a `visible_alias`.**

- `negotiate_leaseless_recovery_contract` (`connection.rs:784-813`) parses **only** `homeboy daemon reconcile-leaseless-orphans --help`, and `declared_long_options` (`:815-837`) accepts only bare long options in the `Options:` block.
- `triage`'s `--timeout`/`--interval` are `global = true` on **`TriageArgs`**, not on `Cli`. They propagate to `triage`'s subcommands and nowhere else. They cannot appear in `daemon reconcile-leaseless-orphans --help`.
- `root_global_flag_surface_is_pinned` (#10554) lists the 14 root globals; neither `timeout` nor `interval`/`poll-interval` is among them, and this PR adds, removes, and renames **zero** root globals — so that test is untouched.
- No `daemon` argument, doc comment, or `--confirm-*` flag is touched.

Everything else in this PR is internal: a new non-command module, a poller trait, a notification-route read, and tests.

## Generated CLI docs: no regeneration needed

`reference_docs.rs::documented_options` (`:286-294`) filters out `arg.is_global_set()`. Every argument on `TriageArgs` is `global = true`, which is why `docs/reference/cli/commands/triage.md` has **no Options table for `homeboy triage` at all** today. The renamed flag is therefore invisible to the generated tree, and `cli_surface::reference_docs`' byte-match test is unaffected. Verified by inspection of the generated file; no other command surface changed.

`docs/commands/triage.md` (hand-written narrative) **was** updated, since it showed `--poll-interval` in a recipe.

## Corrections to the issues

**#10315:**
1. "the same command implemented twice" / "identical output field sets" — **wrong**. Six fields overlap (`terminal`, `timed_out`, `waited_secs`, `poll_count`, `notify`, `_homeboy_actionable`); `runs.watch` additionally has `run_id`/`status`/`run` and `activity.watch` has `schema`/`id`/`state`/`item`/`next_actions`. Different types, different resolvers, different stores.
2. "Make `activity::watch` construct an `ActivityPoller: RunPoller`" — **not implementable**; `RunPoller` yielded `RunRecord`.
3. `runner job logs --follow` listed as a comparable watch loop — **wrong shape**; see above.
4. Unmentioned: the two `parse_duration`s disagree on zero, and activity's own test depends on it.
5. Line refs drifted: `activity.rs:16` → `:21`; loop `:204-279` → `:209-284`; output fields `:77-97` → `:82-102`; `:370` → `:375`. `runs/watch.rs:33` and `:57-76` were correct.

**#10312:**
1. "Delete … `strip_component_target_args` (`route.rs` can construct `Scope::Path{..}` directly)" — **impossible**, as #10510 established. Also now at `:2140`.
2. "Flatten into … `refactor`, `review`, `cleanup`, `release`" — **four of those cannot take it**, and `refactor`'s blocker is a hard clap arg-id collision, not a judgement call.
3. Neither issue was already fixed on main; `git log` shows #10510 landed the partial `ScopeArgs` work and nothing since has touched either surface.

## What I could not verify without cargo

Per this host's constraints (16 GB shared; `cargo build`/`test`/`check`/`clippy` OOM the supervisor), **only `cargo fmt` was run** — and it was verified to actually reach the new files with a deliberate mis-format probe, per the #10565 lesson. Everything below is CI's to confirm:

- That it compiles. The higher-risk spots I reasoned through by hand: the `WatchResult<RunRecord>` partial move in `build_output` (reads `waited`/`poll_count` after `item` moves), temporary-lifetime binding of `Cli::command()` in the new structural tests, and `Vec<String> == Vec<&str>` / tuple-`PartialEq` inference in the new assertions.
- That `hand_rolled_entity_selection_is_confined_to_deploy` yields exactly `["homeboy deploy"]`. Established by grep, not by running it: the only fields producing `--fleet` in the entire workspace are `ScopeArgs::fleet` and `DeployArgs::fleet`, there is no `long = "fleet"` override anywhere, and every `derive(Args)` in the workspace lives in `homeboy-cli`.
- That the new parse assertions match clap's real behaviour, in particular `release -p growth` and `refactor rename --components one,two` resolving past their optional-subcommand ambiguity.
- That `watch_notification_delivers_to_the_bound_cli_route` reaches `missing_transport` under `with_isolated_home` rather than finding an installed transport.
- That `cargo fmt`'s macro blind spot (#10565) does not apply here — neither new module is declared inside a `macro_rules!` body, and the probe confirmed rustfmt reaches both.

### DIFF
--- a/crates/homeboy-cli/src/commands/activity.rs
+++ b/crates/homeboy-cli/src/commands/activity.rs
@@ -9,6 +9,7 @@ use homeboy::core::activity::{
 use homeboy::core::notification_payload::{
     NotifyAction, NotifyAttachment, NotifyEventKind, NotifyPayload, NotifySubject,
 };
+use homeboy::core::notification_route::{self, NotificationRoute};
 use homeboy::core::notify::{self, NotifyEvent, NotifyOutcome};
 use homeboy::core::{Error, Result};
 
@@ -16,9 +17,8 @@ use super::utils::response::{
     CommandActionableMetadata, CommandAgentTaskRef, CommandJobRef, CommandNextAction,
     CommandNextActionKind, CommandResultRefs, CommandRunRef,
 };
+use super::utils::watch::{watch_loop, WatchConfig, WatchPoller, WatchResult, TIMEOUT_EXIT_CODE};
 use super::CmdResult;
-
-const TIMEOUT_EXIT_CODE: i32 = 124;
 
 #[derive(Args, Clone)]
 pub struct ActivityArgs {
@@ -206,50 +206,61 @@ fn show(id: &str) -> CmdResult<ActivityOutput> {
     ))
 }
 
+/// Production poller: resolve one activity item across every activity source.
+///
+/// Deliberately **non**-reconciling, which is the load-bearing behavioural
+/// difference from `runs watch`'s `StorePoller`. An activity read must not
+/// mutate an observation record — `activity_reads_do_not_reconcile_stale_running_records`
+/// pins that — so the two commands share the poll loop (`utils::watch`) but not
+/// the poller. This is why #10315's prescribed fix ("construct an
+/// `ActivityPoller: RunPoller`") could not be taken literally: `RunPoller`
+/// yielded a `RunRecord`, and an activity item is a projection over four stores
+/// that frequently has no `RunRecord` behind it at all.
+struct ActivityPoller;
+
+impl WatchPoller for ActivityPoller {
+    type Item = ActivityItem;
+
+    fn poll(&self, id: &str) -> Result<ActivityItem> {
+        activity::resolve_activity(id)
+    }
+
+    fn is_terminal(&self, item: &ActivityItem) -> bool {
+        !activity::is_active(item.state)
+    }
+}
+
 fn watch(args: ActivityWatchArgs) -> CmdResult<ActivityOutput> {
     let interval = parse_duration(&args.interval)?;
     let timeout = args.timeout.as_deref().map(parse_duration).transpose()?;
-    let started = Instant::now();
-    let mut poll_count = 0;
+    let config = WatchConfig { interval, timeout };
 
-    loop {
-        let item = activity::resolve_activity(&args.id)?;
-        poll_count += 1;
-        eprintln!(
-            "activity {}: {:?} (poll {})",
-            args.id, item.state, poll_count
-        );
-        if !activity::is_active(item.state) {
-            return Ok(watch_output(
-                args,
-                item,
-                false,
-                started.elapsed(),
-                poll_count,
-            ));
-        }
-        if let Some(timeout) = timeout {
-            if started.elapsed() >= timeout {
-                return Ok(watch_output(
-                    args,
-                    item,
-                    true,
-                    started.elapsed(),
-                    poll_count,
-                ));
-            }
-        }
-        std::thread::sleep(interval);
-    }
+    let started = Instant::now();
+    let id = args.id.clone();
+    let result = watch_loop(
+        &ActivityPoller,
+        &args.id,
+        &config,
+        std::thread::sleep,
+        || started.elapsed(),
+        |item, poll_count| emit_progress(&id, item, poll_count),
+    )?;
+
+    Ok(watch_output(args, result))
+}
+
+fn emit_progress(id: &str, item: &ActivityItem, poll_count: u64) {
+    eprintln!("activity {}: {:?} (poll {})", id, item.state, poll_count);
 }
 
 fn watch_output(
     args: ActivityWatchArgs,
-    item: ActivityItem,
-    timed_out: bool,
-    waited: Duration,
-    poll_count: u64,
+    result: WatchResult<ActivityItem>,
 ) -> (ActivityOutput, i32) {
+    let timed_out = result.timed_out();
+    let waited = result.waited;
+    let poll_count = result.poll_count;
+    let item = result.item;
     let notify = maybe_notify(&args, &item, timed_out);
     let exit_code = if timed_out {
         TIMEOUT_EXIT_CODE
@@ -269,7 +280,7 @@ fn watch_output(
             schema: activity::ACTIVITY_REPORT_SCHEMA,
             command: "activity.watch",
             id: args.id,
-            state: item.state.clone(),
+            state: item.state,
             terminal: !timed_out,
             timed_out,
             waited_secs: waited.as_secs(),
@@ -413,6 +424,35 @@ fn maybe_notify(
     if !args.notify {
         return None;
     }
+    // The invocation's route, bound once by `cli_runtime` from
+    // `--notification-transport`/`--notification-route` or the matching
+    // environment. This used to be dropped: the event was built with
+    // `transport: None, route: None`, so `activity watch --notify
+    // --notification-route ...` silently delivered to the configured default
+    // transport with no route at all (#10315, found by #10538).
+    //
+    // Unlike `runs watch`, there is no run-scoped fallback available here:
+    // `ActivityItem` is a cross-store projection with no metadata channel, so
+    // it cannot carry the route `ObservationStore::start_run` stamped onto the
+    // record. Honouring the caller's explicit route is the available — and the
+    // requested — answer.
+    let route = notification_route::current();
+    Some(notify::dispatch(&activity_notify_event(
+        item,
+        timed_out,
+        route.as_ref(),
+    )))
+}
+
+/// Build the completion notification for a watched activity item.
+///
+/// Split out from [`maybe_notify`] so the routing decision is unit-testable
+/// without an installed transport to dispatch through.
+fn activity_notify_event(
+    item: &ActivityItem,
+    timed_out: bool,
+    route: Option<&NotificationRoute>,
+) -> NotifyEvent {
     let status = if timed_out {
         "timed_out".to_string()
     } else {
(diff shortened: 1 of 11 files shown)


chubes4:
## Second commit: unrelated generated-doc drift

The `CLI Reference Docs` gate failed on the first push. **None of it came from this branch** — I downloaded the gate's own `cli-reference-docs` artifact and diffed it: `activity.md` and `triage.md` are byte-identical, and the only stale files are `index.md`, `runner.md`, `runs.md`, `self.md`.

The drift is from **#10562**, which deleted `homeboy runs retention` and re-worded the retention flag help on `runs artifact cleanup-persisted`, `self cleanup-runtime-tmp`, `runner slots prune`, and `runner workspace prune`. It landed alongside **#10563** (which introduced the generated tree) without regenerating it, so the gate has been red on every `crates/homeboy-cli/**` PR since — it is also failing on `fix-10564-runner-downloads-predicate`.

Applied verbatim from the artifact. This incidentally confirms the claim in the PR body: **the `triage --interval` rename needed no regeneration**, because both triage watch args are `global = true` and `documented_options` excludes globals by design.

PR #10710, opened 2026-07-29
fix(release): attach verified immutable main checkout

### LINKED ISSUE #10702: Release dry-run suppresses main pushes from immutable detached checkout
## Problem

The rolling Release workflow checks out the immutable push SHA (`ref: ${{ github.sha }}`), which correctly prevents a moving-branch race but leaves Git in detached-HEAD state. `homeboy release` derives the current branch from Git, sees `HEAD` instead of `main`, and deliberately skips the release.

Observed on Release run https://github.com/Extra-Chill/homeboy/actions/runs/30410024691 at `53eb7f3d66c2e77469d0618412a8d37cc090c3a4`:

- Dry-run annotation: `Not on main (current: HEAD) - skipping release`
- Workflow decision: `No releasable commits at HEAD 53eb7f3d`
- `v0.321.1..53eb7f3d` contains many releasable `fix:` and `feat:` commits, including #10530.
- Build, quality, prepare, and publish jobs are therefore all skipped.

## Required contract

Preserve immutable SHA checkout while providing explicit, auditable release-branch identity to planning. A detached CI checkout of the exact `main` event SHA must be treated as `main` only when the workflow supplies and verifies that identity; arbitrary detached local checkouts must continue to fail closed. Add deterministic coverage for the detached immutable-main case and the untrusted detached case.

## Impact

No Homeboy release has been published since `v0.321.1`; SSI #746 remains blocked because merged sidecar parsing repair #10530 is unreleased.

chubes4:
## Summary

- verify the push event branch and SHA against the detached checkout before release planning
- attach the verified immutable commit to local `main` so released Homeboy versions recognize the default branch
- preserve fail-closed behavior for non-main and mismatched-SHA identities

Closes #10702.

## Verification

- `cargo fmt --check`
- `bash -n .github/attach-release-branch.sh`
- `cargo test --test release_workflow_test` (32 passed)
- `cargo test -p homeboy-release` exercised 764 tests; 759 passed and 5 existing environment/baseline-sensitive tests failed outside this workflow-only diff

## AI assistance

OpenCode with `openai/gpt-5.6-terra` drafted the initial investigation candidate. OpenCode with `openai/gpt-5.6-sol` identified the bootstrap flaw, reduced the change to the workflow-owned repair, added behavioral coverage, and ran verification. Chris Huber remains responsible for the final change.

### DIFF
--- a/.github/attach-release-branch.sh
+++ b/.github/attach-release-branch.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+EVENT_BRANCH="${1:?event branch is required}"
+EVENT_SHA="${2:?event SHA is required}"
+HEAD_SHA="$(git rev-parse HEAD)"
+
+if [ "${EVENT_BRANCH}" != "main" ] || [ "${HEAD_SHA}" != "${EVENT_SHA}" ]; then
+  echo "Refusing release identity branch=${EVENT_BRANCH} event=${EVENT_SHA} checkout=${HEAD_SHA}" >&2
+  exit 1
+fi
+
+git switch --force-create "${EVENT_BRANCH}" "${EVENT_SHA}"
+test "$(git branch --show-current)" = "${EVENT_BRANCH}"
+test "$(git rev-parse HEAD)" = "${EVENT_SHA}"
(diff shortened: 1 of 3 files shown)


chubes4:
Superseded by #10706, which landed the same owning workflow repair with stronger measured-verdict coverage and is now exercising the rolling release on main.

Issue #10659, opened 2026-07-28
detached_cook_accepts_reverse_capacity_queue_and_worker_completes_once fails in 59 of 62 completed runs; every one reported green


chubes4:
## Summary

`detached_cook_accepts_reverse_capacity_queue_and_worker_completes_once` (`tests/reverse_cook_queue_acceptance.rs:46`, root `homeboy` package integration test) failed in **59 of 62** phases in which it ran to completion, across 07-28T04:34Z → 07-28T13:34Z. Every one of those CI runs was green, because the failure reproduces identically on the baseline and the differential gate rewrites `current == base` to `pass` (#10657).

## Failure

`tests/reverse_cook_queue_acceptance.rs:371`:

```
thread 'detached_cook_accepts_reverse_capacity_queue_and_worker_completes_once' panicked at tests/reverse_cook_queue_acceptance.rs:371:13:
controller did not project terminal broker result
```

That is the 10 s poll loop waiting for the controller to project the broker's terminal result after the worker exits. The captured `daemon stderr` in the same block repeats:

```
Warning: could not refresh mirrored Lab runner evidence for `agent-task-…-attempt-1-…`:
Invalid argument 'result.observation_run_details': reverse runner terminal result is not a
valid typed observation detail contract: missing f…
```

so the terminal result is being produced but rejected by the typed observation-detail contract, and the controller consequently never projects it. That looks like the actual bug rather than a slow-poll problem.

Example: run [30362303448](https://github.com/Extra-Chill/homeboy/actions/runs/30362303448), Test job log lines 15426 (candidate) and 25814 (baseline) — both phases, same panic.

## Load inversion

The outcome inverts with suite size, which is worth flagging because it will confuse whoever reproduces it:

| phase kind | observations | FAILED | ok |
|---|---|---|---|
| completed (small scope, 17–25 binaries) | 62 | **59** | 3 |
| killed at 1500 s (large scope, 27–29 binaries) | 43 | 0 | 43 |

Both happen *inside the same run*: run 30362979347's candidate phase completed and FAILED while its baseline phase was killed and printed `ok`. So this is not a commit difference. When the suite is large the test takes **431.81 s** and passes; when the suite is small it panics on the 10 s deadline. Something in the reverse-broker path is waiting on wall-clock or on contention that only materialises under a busy machine.

The 431 s figure is also the slow-test problem in #10655 — same test, different defect. This issue is about it *failing*; #10655 is about it *costing 28 % of the budget*.

[...]

Issue #10655, opened 2026-07-28
Make slow tests a first-class finding, not something a human has to read 729 KB of logs to notice


chubes4:
## The gap

`TestCounts` (`crates/contracts/homeboy-extension-contract/src/test_result.rs:6`) is:

```rust
pub struct TestCounts { total, passed, failed, skipped }
```

There is **no duration field**, and nothing in the workspace parses per-test timings. `grep -rn` for `finished in` / `... ok` outside test files returns only `ci_failure_log_triage.rs` matching on `test result: failed` — a pass/fail string, never a time. Homeboy can tell you a test **failed**. It cannot tell you a test is **slow**.

## Why this is a primitive, not a nice-to-have

Slowness is a test defect in the same way failure is. It has an owner, a location, a trend, and a cost — and it silently consumes the budget that everything else needs.

Concrete evidence from today: the CI Test gate began timing out at its 1500s inner budget. Diagnosing it took a full cook session of manual log reading. The finding:

- `reverse_cook_queue_acceptance.rs::detached_cook_accepts_reverse_capacity_queue_and_worker_completes_once` runs **424 seconds** — sole test in its binary, **~28% of the entire budget**.
- Next slowest are 75s and 73s. So one test costs more than the next five combined.

Nothing surfaced that. No gate, no report, no trend. A human had to read 729 KB of retained stdout to find it.

Worse, the timeout **destroys the evidence that would explain it**: when the child is killed it never writes its sidecar, so counts come back `total: 0` — which reads as "no tests ran" and hides genuine failures in the same breath (see #10639).

## What I want

Treat duration as a first-class test fact:

1. **Capture it.** Extend the test-result contract with per-test and per-binary duration. Additive only — `homeboy` and `homeboy-extensions` ship independently, so older consumers must keep deserializing.
2. **Report it.** `review test` should surface the slowest N tests the way it surfaces failures, so a regression is visible the run it appears in.
3. **Flag it.** A threshold (absolute, or share-of-budget) that marks a test as a finding. Share-of-budget is probably the more honest signal — a 400s test is fine in a 2-hour suite and pathological in a 25-minute one.
4. **Trend it.** Homeboy already persists run artifacts and observations; a duration that doubles between runs is exactly the kind of thing this system exists to notice.

## Prior art in this repo to follow

[...]

Issue #10237, opened 2026-07-26
Cook failure output references missing recovery context and hides continuation


chubes4:
## Problem

Cook terminal summaries instruct the operator to use `failure_context`, but the response omits that field and every executable recovery action. The durable candidate is real, yet the normal result does not say whether to run `cook-continue`, `review`, `promote`, `reconcile`, or wait for an operation lease.

## Evidence

Cook `agent-task-f5382842-04da-47ea-96bb-ba5b4957f877` stopped with:

```json
{
  "status": "policy_failure",
  "stop_reason": "promotion provider response was rejected. The successful candidate remains durable; use failure_context to inspect or continue the Cook.",
  "finalization": null,
  "terminal_phase": null
}
```

The compact and full status established:

- provider succeeded
- one non-empty patch exists
- candidate state is `patch_available`
- promotion and gates were not attempted
- a `promote:<cook-id>` operation claim remained `running`

But the Cook response had no `failure_context`, operation/lease information, or `next_actions`. Rerunning `cook --run-id <id>` produced another `durable_failure` that again referenced absent `failure_context`; the actual continuation verb, `cook-continue`, was discoverable only by reading `agent-task --help`.

## Expected

Every non-successful Cook terminal result is self-recovering from the returned envelope. It identifies the failed phase, exact reason, candidate state, blocking operation, lease deadline/owner liveness, and one safe executable next action.

## Acceptance criteria

- `failure_context` is present whenever `stop_reason` references it.
- Compact output includes phase, reason code, durable candidate ref, blocking claim if any, and primary next command.
- `next_actions` distinguish inspect, continue, reconcile, retry-provider, and abandon operations.
- `cook --run-id` either clearly attaches/continues or rejects with the canonical `cook-continue` command; it never creates an ambiguous second lifecycle.
- Full output exposes redacted provider validation evidence and operation history.
- Contract tests cover provider failure, promotion rejection, gate failure, stale claim, finalization failure, and recoverable candidate states.
- Human output and JSON derive from the same recovery model.

## Related

- #8357
- #8363
- #9817

## AI assistance

- **AI assistance:** Yes
- **Tool:** OpenCode with `openai/gpt-5.6-sol`
- **Used for:** Followed the failed Cook through status, review, promote, resume, and command discovery and documented the missing recovery contract with Chris Huber.

chubes4:
[...]

Issue #10657, opened 2026-07-28
Differential gate rewrites an equal-failure candidate to `pass`, so a test red on main produces a fully green run


chubes4:
## Summary

The differential gate does not only downgrade a pre-existing failure to `baseline_red` (a `::warning::` and a :warning: in the PR comment). When the candidate and the baseline report the **same** failure count, it rewrites the result to **`pass`** — a `::notice::` and a :white_check_mark:. A test that is red on `main` therefore produces a fully green CI run with no warning anywhere, and stays red indefinitely because nothing ever reports it.

This is a distinct defect from `baseline_red`. `baseline_red` is visible and defensible. `pass` is neither.

## Where

`Extra-Chill/homeboy-action`, deployed tag `v2` = `66fb512` (v2.8.25); `.github/workflows/ci.yml:148` pins `ref: v2`.

`scripts/core/apply-differential-gate.py` at `v2`:

```python
207:        if current <= base:
208:            adjusted[command] = "pass"
209:            print(
210:                f"::notice::Differential gate accepted {command}: current={current} base={base}",
```

`enforce-final-status.sh` then has nothing to act on: it errors on `fail` (`:54`) and `timeout` (`:58`), warns on `baseline_red` (`:62`) and `inconclusive` (`:65`), and has no branch at all for a `pass` that was manufactured from two failures.

The same branch exists on `homeboy-action@main` at `apply-differential-gate.py:235-236`; #305 did not change it.

## Evidence — a green run over a real failure

Run [30380078587](https://github.com/Extra-Chill/homeboy/actions/runs/30380078587), Test job `90345508296`, conclusion **success**:

| phase | line | output |
|---|---|---|
| candidate | 15524 | `test read_only_cli_commands_complete_while_runtime_promotion_is_held ... FAILED` |
| candidate | 15540 | `test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.33s` |
| baseline | 25950 | `test read_only_cli_commands_complete_while_runtime_promotion_is_held ... FAILED` |
| baseline | 25966 | `test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.13s` |
| gate | 26154 | `##[notice]Differential gate accepted review test: current=1 base=1` |

[...]

PR #10728, opened 2026-07-29
fix(lab): stop re-hashing the immutable controller runtime pin (#10659)

chubes4:
## Why

`detached_cook_accepts_reverse_capacity_queue_and_worker_completes_once` is red on `main` and is the slowest binary in the suite. Two intertwined defects, tracked separately here.

**Runtime.** `homeboy agent-task cook` seals itself into an immutable controller runtime pin and re-execs it. Publishing and validating that pin SHA-256s a multi-hundred-MB *debug* binary repeatedly — `pin_executable` hashes the source, `publish_pin` hashes the staged copy or the existing destination, and `validate_pin` hashes the destination again; `admit_current_for` then repeats the whole sequence in the re-exec'd child. The repo already measured this cost once and worked around it for the daemon (`HOMEBOY_TEST_DAEMON_BINARY_SHA`, "~20s per daemon-state write"); the controller-runtime path has no equivalent.

**Diagnosis.** The only completed observation of this test on `main` ([run 30408709906](https://github.com/Extra-Chill/homeboy/actions/runs/30408709906), job 90442105594) panics at the detached Cook CLI assertion with a `durable_failure` summary report. The summary Cook view exposes no `failure_context` (#10237), so the panic message names no cause at all.

## What

Commit 1 (this push) makes the failure self-describing before changing behaviour:

- a wall-clock phase ledger rendered into every panic, so a CI log alone shows where the budget goes
- `agent-task status --full` plus the controller daemon's stderr hydrated when the Cook CLI fails
- the terminal-projection poll bounded on **observations** as well as wall clock

That last one is not a widened timeout. The loop spawns a whole `homeboy` subprocess per poll; when one observation costs more than the 10s deadline, the deadline expires inside the first iteration and the controller is declared stalled having been asked exactly once. The deadline was measuring the probe, not the controller.

Further commits follow once CI reports the ledger.

Refs #10659, #10655, #10237

### DIFF
--- a/tests/reverse_cook_queue_acceptance.rs
+++ b/tests/reverse_cook_queue_acceptance.rs
@@ -85,6 +85,48 @@ impl Drop for ReverseSessionHeartbeat {
     }
 }
 
+/// Wall-clock ledger for the acceptance run.
+///
+/// This test is the slowest binary in the suite and its deadlines are wall
+/// clock, so a bare panic message says nothing about which phase consumed the
+/// budget. Record each boundary and render the ledger into every panic so a CI
+/// log is sufficient evidence — the machine that reproduces it is not
+/// available to whoever reads the failure.
+struct PhaseLedger {
+    started: Instant,
+    previous: Instant,
+    phases: Vec<(&'static str, Duration)>,
+}
+
+impl PhaseLedger {
+    fn new() -> Self {
+        let now = Instant::now();
+        Self {
+            started: now,
+            previous: now,
+            phases: Vec::new(),
+        }
+    }
+
+    fn mark(&mut self, phase: &'static str) {
+        let now = Instant::now();
+        self.phases.push((phase, now.duration_since(self.previous)));
+        self.previous = now;
+    }
+
+    fn render(&self) -> String {
+        let mut rendered = String::from("phase timings (seconds):");
+        for (phase, elapsed) in &self.phases {
+            rendered.push_str(&format!("\n  {phase}: {:.2}", elapsed.as_secs_f64()));
+        }
+        rendered.push_str(&format!(
+            "\n  TOTAL: {:.2}",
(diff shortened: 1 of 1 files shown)


chubes4:
## First CI evidence: the whole test is one CLI invocation

Test job [90459667599](https://github.com/Extra-Chill/homeboy/actions/runs/30415063936/job/90459667599). The phase ledger from the candidate phase:

```
phase timings (seconds):
  hermetic_context: 0.00
  git_fixtures: 0.03
  daemon_ready: 0.26
  server_and_runner_configured: 0.35
  detached_cook_cli: 75.93
  TOTAL: 76.92
```

**75.93 of 76.92 seconds is inside the single `homeboy agent-task cook` subprocess.** Everything else in the fixture — hermetic context, git fixtures, daemon startup, `server create`, `runner enable` — costs 0.64s combined. There is no slow polling, no sleep, and no waiting on the reverse worker: the test never gets that far.

### The gate reported green on a red test again

Both phases failed identically (candidate `:381`, baseline `:307` — same assertion, my line numbers moved), so the differential gate rewrote `current == base` to pass (#10657). `homeboy / Test` is green on this PR and the test is red in both phases. Worth stating plainly because it is the third time today.

### The failure

Cook exits `durable_failure` with `attempts: []` after 75.93s, and `agent-task status <latest_run_id> --full` reports:

```
Invalid argument 'run_id': agent-task run record not found
```

The durable run record was never written. `submit_plan_with_runtime_admission` persists the record *before* controller admission, so the failure is upstream of that — after recipe materialization (which is what qualifies it for `durable_failure`) but before the run record exists. `daemon stderr` is empty: the Cook never reached the controller daemon at all.

The compacted Cook view drops `failure_context`, so the reason code is still unnamed. Pushed `--full` on the fixture's cook invocation to get it.

Issue #10733, opened 2026-07-29
Release recovery cannot complete assets after a draft becomes published mid-run


chubes4:
## Problem

Tagged recovery assumes the matching release remains a draft for the entire rebuild/finalization window. If it becomes published while recovery is building artifacts, the finalizer refuses the same release instead of reconciling its verified partial inventory.

## Evidence

Recovery run: https://github.com/Extra-Chill/homeboy/actions/runs/30414880669
Tag/commit: `v0.322.0` / `3a94fad401418550e76d378bcc9ced33632da36c`
Published release: https://github.com/Extra-Chill/homeboy/releases/tag/v0.322.0

- Recovery accepted the exact existing tag and rebuilt all four platform matrices.
- The authoritative plan expected 13 assets.
- The release transitioned from draft to published at `2026-07-29T02:14:40Z` while recovery was still running.
- At finalization, the published release contained 7 valid assets.
- `validate_draft_adoption` rejected it with `draft adoption requires the matching unpublished GitHub Release`.
- Publication verification failed because the host job failed.
- Missing planned assets are `dist-manifest.json`, both aarch64 archives/sidecars, and the x86_64 macOS archive/sidecar.

## Required contract

Recovery must reconcile a matching release that changes from draft to published after planning. Verify tag/target identity and every pre-existing asset by name, size, and digest; upload only missing authoritative assets; reject conflicting bytes, duplicates, or an unrelated release; and finish successfully without trying to republish an already-published release.

## Acceptance criteria

- Deterministic coverage changes release metadata from draft to published between probe and finalization.
- Existing matching assets are reused and missing assets are uploaded.
- Conflicting existing assets fail closed with exact names/digests.
- Final verification requires all 13 planned assets and a published release.
- Existing draft adoption and already-complete published-release idempotency remain intact.
- Recovery run 30414880669 and exact asset inventory remain linked as reviewer-facing evidence.

## Related

- #10519 tracks release recovery bootstrap and end-to-end proof.
- #10611 added complete-draft adoption.

## AI assistance

OpenCode with `openai/gpt-5.6-sol` traced the release timeline, compared planned and remote inventories, and drafted this report. Chris Huber reviewed and owns the issue.

Issue #10519, opened 2026-07-28
Release recovery executes the stranded tag binary and cannot bootstrap publisher fixes


chubes4:
## Problem

`workflow_dispatch` recovery for an existing release tag executes Homeboy from the stranded tag itself. A publisher fix merged to `main` therefore cannot repair an older draft release: recovery repeatedly runs the defective binary that created the draft.

The workflow also rebuilt every platform artifact before reaching the known-broken publisher, turning each failed recovery into a 40-minute cycle.

## Evidence

Target: `v0.321.0` at `192964f61a7e`.

- Initial release run: https://github.com/Extra-Chill/homeboy/actions/runs/30288253027
- Explicit recovery run: https://github.com/Extra-Chill/homeboy/actions/runs/30294528021
- Owning fix: https://github.com/Extra-Chill/homeboy/pull/10480
- Both runs failed with `gh api release metadata exited with status 1`.
- The recovery workflow checked out `v0.321.0` for the host job and invoked `homeboy-action` with `source: .`, so it built and ran `0.321.0+192964f61a7e` without #10480.
- The draft already contained all 13 uploaded assets with valid GitHub SHA-256 digests.

Recovery only succeeded after manually separating the roles:

1. Build the control binary from current `main` containing #10480.
2. Keep the release source checkout at exact tag `v0.321.0`.
3. Restore the original run artifacts whose hashes matched the existing draft.
4. Run `homeboy release --head --from-artifacts ... --apply` with the repaired control binary.

That is the correct owning-layer path, but the workflow should perform it directly.

## Expected contract

Release recovery separates the trusted recovery runtime from the immutable release target. The current control plane may repair publication for an older tag without changing that tag or its artifact bytes.

## Acceptance criteria

- Recovery uses a current trusted Homeboy binary/workflow implementation while targeting the exact tagged source tree.
- Output records control-binary identity, target tag/SHA, and artifact provenance separately.
- If the originating run artifacts remain available, recovery reuses them instead of rebuilding all platforms.
- If the draft already has every authoritative asset, recovery verifies and publishes without rebuilding or re-uploading.
- Artifact identity remains fail-closed by name, size, and digest.
- A regression test proves a publisher bug fixed after tag creation can recover the older draft.

## Related

- #10441 tracks orphaned/stranded release state.
- #10480 fixed draft lookup, but the tagged recovery binary could not consume that fix.

## AI assistance

- Model: OpenAI GPT-5.6 Sol
[...]

Issue #10477, opened 2026-07-27
review test runs only the root package's targets on a workspace component — member crate lib tests never execute in CI


chubes4:
## Problem

`homeboy review test` on a Cargo **workspace** component executes only the **root package's** test targets in CI. Every member crate's `--lib` unit tests are compiled but never run.

This is the outermost reason `hydration::tests::*` never executed for `76c8c99a8`, and it is independent of the two scope-selection bugs already fixed (#10449 / #10450 and #10465 / #10468).

## Evidence

Test job for PR #10468 (`homeboy 0.320.0+9a4eeffed9e6`, rust extension `ed2ce7a5`, `SCOPE_MODE: changed`):

Every executed target belongs to the root package — 3 unittest targets plus 20 integration targets:

```
Running unittests src/lib.rs               (deps/homeboy-ea3485611ca8eafd)
Running unittests src/bin/bench-audit-self.rs
Running unittests src/main.rs
Running tests/agent_task_promotion_provider.rs
... 20 integration targets total
```

No `deps/homeboy_lab_runner-*`, `deps/homeboy_core-*`, `deps/homeboy_extension-*` — i.e. no member-crate lib tests, across **both** differential runs (current and base).

Member crates *are* compiled as test targets in the same job (13 `` `<crate>` (lib test) `` clippy warnings), so this is not a build gap — those binaries are built and then not executed.

Same 23-target set observed before #10450, after #10450, and after #10468. The scope-selection fixes are correct at the unit level but have not changed which targets CI runs.

## Where it likely is

`rust/scripts/test-runner.sh` only adds `--workspace` when the scope kind is `workspace` or `full`:

```sh
if [ "$SCOPE_KIND" = "workspace" ] || [ "$SCOPE_KIND" = "full" ]; then
    case " $(... .args[] ...) " in
        *" -p "* | *" --package "* | *" --workspace "* | *" --exclude "*) ;;
        *) TEST_ARGS+=(--workspace) ;;
    esac
fi
```

`rust_test_scope_json` defaults the kind to `workspace`, so a default run should get `--workspace`. The observed run did not, and emitted no `HOMEBOY_TEST_SCOPE_MESSAGE`, so the effective `SCOPE_KIND` under `SCOPE_MODE: changed` needs to be established directly — I was not able to pin it from the job log alone.

## Expected contract

- A changed-scope run on a workspace component executes the affected member crates' lib tests.
- A full/default run executes every member crate's tests.
[...]

Issue #10687, opened 2026-07-28
Full test gate times out in handoff lifecycle suite


chubes4:
## Summary

The Homeboy PR `Test` gate exceeds its 2,700-second command timeout while running `agent_task_lifecycle::tests::handoff_and_proxy`, preventing structured test counts and making unrelated PRs fail after roughly 1h40m.

## Evidence

Observed on PR #10678, run https://github.com/Extra-Chill/homeboy/actions/runs/30399605219:

- Rustfmt: pass
- Lint: pass
- Audit: pass
- Windows Compile: pass
- Workspace Tests Compile: pass
- Test: timeout, exit 124
- Candidate and baseline `homeboy review test` runs both timed out.

The final output reported:

```
test phase timed out after 2700s before reporting test counts, suite incomplete
```

Before timeout, unrelated lifecycle failures appeared in:

- `controller_proxy_is_queued_before_handoff_then_binds_runner_child`
- `preacceptance_snapshot_binds_planned_runner_job_before_validation`

Many tests in the same module ran for more than 60 seconds. The PR's modified focused suites passed independently: 37 lint workflow tests and 24 release quality tests.

## Expected

The full suite must complete within its configured budget, or the long-running lifecycle suite must be partitioned so quality gates provide bounded, actionable signal. Candidate/baseline differential execution should not consume the entire gate budget twice and return zero structured counts.

chubes4:
Fresh evidence: this blocked a release on `main` today.

**Run:** [30449012761](https://github.com/Extra-Chill/homeboy/actions/runs/30449012761) — Release for `9b1915ae1` (#10763).
**Result:** `Test` job failed, `Release Quality Policy` failed, and every downstream release job was skipped. Build, Audit and Lint were all green.

From `homeboy-observations-homeboy-review-test-gate-test-Linux/runs.json`:

```json
{
  "command": "homeboy test homeboy --skip-lint --changed-since=cd88f444d4ec52dbbd62fec36ab370e4721efd48",
  "status": "error",
  "metadata_json": {
    "exit_code": 124,
    "failure_count": 0,
    "observation_status": "interrupted",
    "summary": { "exit_code": 124, "failed": 0, "passed": 0, "skipped": 0, "total": 0 }
  }
}
```

[...]

PR #10836, opened 2026-07-29
test(release): make deploy guard baselines independent of host state

### LINKED ISSUE #10126: test: release crate baselines depend on host git and group state
## Summary

A serial `cargo test -p homeboy-release --lib` run on the production coding host has three unrelated baseline failures after all #6853-focused tests pass:

- `deploy::orchestration::tests::default_release_guard_rejects_commits_after_component_prefixed_tag` unexpectedly returns `Ok(())`.
- `deploy::orchestration::tests::default_tagged_release_guard_still_blocks_unreleased_head` unexpectedly returns `Ok(())`.
- `deploy::safety_and_artifact::tests::test_directory_artifact_normalizes_group_write_and_setgid` assumes the deployed directory group differs from the process group; both are gid 0 when tests run as root.

## Reproduction

`cargo test -p homeboy-release --lib -- --test-threads=1`

The failures reproduce outside the #6853 paths. Focused release package/cleanup, deployment recovery, planning, publish, and workflow contract suites pass. Tests should fully isolate git configuration/tag policy and avoid requiring a non-root group identity.

chubes4:
Closes #10126.

## What the failures actually were

The issue attributed all three to "host git and group state." Two of them have a different and more consequential cause.

`detect_tag_gap` resolves the latest component tag through the **release provider registry**, and that registry is only populated by the CLI at startup (`cli_runtime.rs:153`). In a `homeboy-release` lib test, `NoopProvider::latest_component_tag` returns `None`, so `detect_tag_gap` returns `None`, no gap is ever found, and every deploy guard passes vacuously.

That explains the reported symptom precisely — the two negative tests "unexpectedly return `Ok(())`" because the guard *cannot* refuse without a provider.

**It also means the positive test was passing for the wrong reason.** `default_release_guard_accepts_component_prefixed_tag_at_head` asserts `Ok(())` from a guard that had no way to fail. It has never verified anything. Registering the provider makes all three meaningful; all three now pass.

The setgid test is a genuine host-state dependency, just not the one described. `deploy_artifact` resolves permission modes through `defaults::load_defaults()`, which reads the host's Homeboy config — so an operator-configured `permissions.remote` decided whether the assertion held. It is now wrapped in `with_isolated_home` so it reads built-in defaults. (It passes as root on this host either way, so the reporter's failure was almost certainly config, not gid.)

## Verification

`cargo fmt --check`; `cargo check --workspace --tests` exit 0.

Serial `cargo test -p homeboy-release --lib -- --test-threads=1`:

| | failures |
|---|---|
| unmodified `main` | **12** |
| this branch | **10** |

## ⚠️ 10 pre-existing failures remain on main — no gate reports them

All verified red on unmodified `main` with my changes stashed. They fail in **parallel runs too**, and at least one fails **in complete isolation**, so this is not test pollution:

```
release::deployment::tests::release_recover_resumes_only_failed_project_with_published_source_projection
release::execution_plan::tests::release_tag_versions_sort_by_semver_precedence
release::executor::github_release::gh_cli::tests::bounded_command_closes_inherited_descendant_pipes
release::executor::github_release::gh_cli::tests::bounded_command_preserves_nonzero_empty_stderr
release::executor::github_release::gh_cli::tests::bounded_command_reports_timeout
release::executor::github_release::gh_cli::tests::bounded_command_retains_large_release_metadata_but_diagnostics_stay_compact
release::executor::github_release::gh_cli::tests::bounded_command_times_out_with_continuous_output
release::planning_quality::tests::extension_release_lint_blocks_extension_bootstrap_failure
release::planning_quality::tests::extension_release_lint_blocks_malformed_missing_and_producer_error_evidence
release::planning_quality::tests::extension_release_lint_rejects_invalid_explicit_lint_ownership
```

I diagnosed one of them while here. **`release_tag_versions_sort_by_semver_precedence`** fails in isolation at `execution_plan.rs:360`:

```
assertion `left == right` failed
  left: Greater
 right: Equal
```

The assertion is `v("v1.2.3+build.5").cmp(&v("v1.2.3")) == Ordering::Equal`, and the doc comment above it states *"build metadata is ignored for precedence (semver §10-11)"*. That is true of semver **precedence**, but `semver::Version` implements `Ord` consistent with its `Eq`, which includes build metadata — so `cmp` returns `Greater`. Introduced by `706627296` ("parse versions with semver instead of splitting on dots"), which correctly replaced hand-rolled parsing but inherited an incorrect assumption about the crate's `Ord`.

Fixing it means deciding whether release tag selection should follow spec precedence (strip build before comparing) or the crate's total order. That has release-selection consequences, so I left it rather than changing ordering semantics inside a test-isolation PR.

**These five `gh_cli::bounded_command_*` failures are worth attention given the release-pipeline work in #10519 and #10733** — they cover exactly the bounded-download and timeout paths that pipeline depends on.

This is #10477 in the flesh: member-crate lib tests are compiled but never executed, so ten red tests in `homeboy-release` have been invisible to CI.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) via chubes-bot

### DIFF
--- a/crates/homeboy-release/src/deploy/orchestration/mod.rs
+++ b/crates/homeboy-release/src/deploy/orchestration/mod.rs
@@ -844,6 +844,21 @@ mod tests {
         }
     }
 
+    /// Install the real release provider for this test binary.
+    ///
+    /// `detect_tag_gap` resolves the latest component tag through the release
+    /// provider registry, and that registry is only populated by the CLI at
+    /// startup. In a `homeboy-release` lib test the `NoopProvider` returns
+    /// `None`, so no tag gap is ever detected and every deploy guard passes
+    /// vacuously — the negative assertions below could not fail for the right
+    /// reason, and the positive one passed for the wrong one (#10126).
+    ///
+    /// Registration overwrites a mutex slot, so calling this from each test is
+    /// safe under any thread count or ordering.
+    fn with_release_provider() {
+        crate::release::provider_impl::register();
+    }
+
     fn init_repo_with_tag_gap(path: &Path) {
         let run = |args: &[&str]| {
             std::process::Command::new("git")
@@ -1442,6 +1457,7 @@ mod tests {
 
     #[test]
     fn default_tagged_release_guard_still_blocks_unreleased_head() {
+        with_release_provider();
         let dir = TempDir::new().expect("temp dir");
         init_repo_with_tag_gap(dir.path());
 
@@ -1459,6 +1475,7 @@ mod tests {
 
     #[test]
     fn default_release_guard_accepts_component_prefixed_tag_at_head() {
+        with_release_provider();
         let dir = TempDir::new().expect("temp dir");
         let root = dir.path();
         let plugin_path = root.join("plugins/host/studio-native");
@@ -1489,6 +1506,7 @@ mod tests {
 
     #[test]
     fn default_release_guard_rejects_commits_after_component_prefixed_tag() {
+        with_release_provider();
         let dir = TempDir::new().expect("temp dir");
         let root = dir.path();
         let plugin_path = root.join("plugins/host/studio-native");
--- a/crates/homeboy-release/src/deploy/safety_and_artifact.rs
+++ b/crates/homeboy-release/src/deploy/safety_and_artifact.rs
@@ -542,45 +542,52 @@ mod tests {
     #[test]
     #[cfg(unix)]
     fn test_directory_artifact_normalizes_group_write_and_setgid() {
-        let temp = tempfile::tempdir().expect("temp dir");
-        let source = temp.path().join("source");
-        let target = temp.path().join("plugins").join("sample-plugin");
-        fs::create_dir_all(&source).expect("source dir");
-        fs::create_dir_all(target.parent().expect("target parent")).expect("target parent dir");
-        fs::write(source.join("plugin.php"), "<?php").expect("source file");
-        fs::set_permissions(source.join("plugin.php"), fs::Permissions::from_mode(0o644))
-            .expect("source permissions");
+        // `deploy_artifact` resolves its permission modes through
+        // `defaults::load_defaults()`, which reads the host's Homeboy config. An
+        // operator-configured `permissions.remote` therefore decided whether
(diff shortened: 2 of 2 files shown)


chubes4:
## Pushed a fix for the timeout that failed this PR

The Test gate failed here at **1h39m52s** — `exit_code: 124`, `RESULTS: {"review test":"timeout"}`, and it timed out in **both candidate and baseline phases** (`Baseline homeboy review test exited 124`). Not caused by this branch. Every other gate passed.

The observation `stdout_tail` named the culprit:

```
test cook_rejects_local_detach_before_worktree_resolution has been running for over 60 seconds
```

A test whose entire job is to assert that a flag combination is rejected *before* worktree resolution. So I traced it.

### Root cause

It is not the cook path, and it is not a hang. `agent-task cook` pins an immutable content-addressed controller runtime, which means SHA-256 over the homeboy executable. That binary is **698 MB** unoptimized, and `sha2` compiled at `opt-level = 0` hashes it roughly 40x slower than optimized code — `sha256sum` does the same file in **0.54 s**. Two distinct files are hashed per run, so every single invocation paid ~44 s of pure hashing.

Confirmed by `strace -e trace=execve`: the process re-execs its pinned copy, and `/proc/<pid>/wchan` sits in `do_wait`. `--version`, `--help`, and even a flag-conflict rejection are all ≤0.05 s, so it is specific to commands that pin a runtime.

Every cook integration test spawns the real binary, so each one paid it. `tests/cook_lab_handoff_test.rs` alone has five.

### Fix

`70bc2645b` raises `opt-level` for the hash dependencies only (`sha2`, `sha1`, `digest`, `crypto-common`) in the dev profile. Our own crates stay fully unoptimized and debuggable. Four lines of `Cargo.toml`, no code change.

Measured on this host:

| | before | after |
|---|---|---|
| `agent-task cook` (cold pin) | 44.3 s | **1.7 s** |
| `agent-task cook` (warm pin) | 44.1 s | **1.3 s** |
| `cook_lab_handoff_test` (5 tests) | >100 s | **1.75 s total** |

That is a 26x reduction on the single most-repeated cost in the suite, and it should take a large bite out of the 2700 s budget rather than just raising it.

Cross-referencing #10687.

Issue #11151, opened 2026-08-01
Epic: consolidation burndown from the 84af2765e investigation (68 findings across 3 repos)


chubes4:
Tracking epic for the 2026-08-01 consolidation investigation, run against a fresh worktree at `origin/main@84af2765e` (plus homeboy-extensions `074c66b7` and chubes4/homeboy-rigs).

**Scope surveyed:** 930,390 lines of Rust · 1,856 files · 23 crates + 16 contract crates · 207k lines of test code · 173 docs. Ten parallel read-only investigations. Every finding below carries file:line evidence.

---

## The thesis

**The architecture is sound.** The debt is two things:

1. **Nine of homeboy's own best primitives were built, applied to 1–4 consumers, and then abandoned.** `CommandReport<T>` 3/41 · `TypedCommandAdapter` 3/41 · `DryRunArgs` 1 · `WriteModeArgs` 1 · `watch_loop` 2/3 · `ScopeArgs` 3/8 · `ops_command_descriptors!` 16/41 · `summary_json` partial · `CommandSpec.safety.dry_run_flag` descriptive-only. Almost nothing here is wrong design — it is correct design, unfinished.
2. **Self-enforcement is structurally incapable of firing**, and the gap was papered over with a 1,519-entry baseline of which 46% is one self-inflicted rule.

---

## Tier 0 — gates that report green while enforcing nothing

| # | Finding |
|---|---|
| #11093 | **`homeboy.json` has no `audit_rules` key** → 11 detectors always return `Vec::new()`. Two are in the **PR-blocking profile**. ~6,236 LOC of detector inert. **~30 lines of config is the highest-ROI change available.** |
| #11094 | `unbounded_output_capture` fails open on any `take(` in the file — 13 `wait_with_output()` sites unflagged |
| Extra-Chill/homeboy-extensions#2504 | **homeboy-extensions has no PR CI.** Any change to any extension gets zero validation. **Land this first — it is the precondition for #11121, #11123, and #2506.** |
| Extra-Chill/homeboy-extensions#2507 | 88 smoke tests have no runner at all (~15–20k lines that read as coverage and are not) |
| #11123 | `lint.findings` seeded three times and gated by an `is_file()` check that ignores the manifest declaration |

## Tier 1 — the boundary is decorative

| # | Finding |
|---|---|
| #11095 | 678 baseline fingerprints duplicated **byte-for-byte in the same file** (24% of a 2,821-line config), read by nothing |
[...]

Issue #10097, opened 2026-07-26
homeboy-agents baseline suite leaks global test state


chubes4:
## Problem

Current `origin/main` (`781789fbd`) fails `cargo test -p homeboy-agents --lib` in a clean primary checkout. The #10086 candidate reproduces the same failures, so they are baseline regressions rather than candidate behavior.

## Evidence

Clean command:

```sh
cargo test -p homeboy-agents --lib
```

Observed failure families include:

- `HOME environment variable not set` across promotion/dispatch tests.
- Fixture worktree handles rejected because no test worktree provider is configured.
- Provider scratch allocation reports missing synthetic run records.
- Shared artifact-root lock poisoning after the first evidence assertion fails.
- Global provider/runner reconciliation state crossing test boundaries.
- Controller scratch tests observing dirty state from another fixture.

The full workspace gate on #10086 reports 56-57 failures depending on scheduling. Serializing with `RUST_TEST_THREADS=1` does not fix the suite because tests also retain order-sensitive global state.

Relevant runs:

- `cook-homeboy-10086-gate-fix-2-20260726`
- `cook-homeboy-10086-current-main-20260726-attempt-1-782d33f1`

## Acceptance criteria

- Test-owned HOME, provider registries, worktree providers, run stores, artifact roots, and scratch roots are scoped/restored deterministically.
- `cargo test -p homeboy-agents --lib` passes repeatedly on clean current main with default parallelism.
- `cargo test --workspace` passes inside a Homeboy promotion gate.
- Coverage continues proving real provider, promotion, adoption, and scratch lifecycle behavior rather than disabling or serializing tests.

Issue #9774, opened 2026-07-23
Isolate process-global state across the homeboy-agents agent_task test gate


chubes4:
## Problem

The configured Cook gate:

```bash
cargo test -p homeboy-agents agent_task
```

produced a broad cluster of 18 failures when run with the default test-thread count after promoting the focused #9752 patch. Failures spanned scheduler retries, provider artifact materialization, promotion export, Cook recipe indexing, liveness reconciliation, and runtime pin recovery, none touched by that patch.

Running focused tests serially made the #9752 behavior pass and removed the broad cross-domain failure pattern. A separate deterministic artifact-projection baseline failure remains tracked independently. The pattern indicates tests are contending over process-global environment variables, HOME/data roots, temp runtime state, or shared daemon/config fixtures.

Gate evidence: `/Users/chubes/.local/share/opencode/tool-output/tool_f8c67c1cb001egbc48xs720U3P` from `cook-issue-9752`.

## Expected

The package gate should be deterministic at Cargo’s default concurrency and under concurrent Cook worktrees. Tests that mutate process-global state must use isolated subprocesses or one shared serialization/restoration primitive.

## Acceptance criteria

- Repeated `cargo test -p homeboy-agents agent_task` runs pass at default thread count.
- Inventory tests mutating environment, HOME, config roots, daemon state, and global synchronization.
- Replace ad hoc env mutation/restoration with a panic-safe shared isolation contract or subprocess boundary.
- Ensure independent worktree test processes use distinct data/runtime/temp roots.
- Add a stress/repetition job that exercises the affected scheduler, provider, promotion, Cook, and lifecycle families concurrently.
- Keep serial execution available for diagnosis, but do not make `--test-threads=1` the permanent gate workaround.

## Related

- The deterministic missing logical-artifact failure is separate and should not be hidden by this concurrency repair.

## AI assistance

- **AI assistance:** Yes
- **Tool:** OpenCode with OpenAI GPT-5.6 Sol
- **Used for:** Full-gate execution, failure-family analysis, serial comparison, duplicate search, and issue drafting.

chubes4:
## Investigation note (from a local reproduction attempt)

While trying to reproduce this on a shared VPS runner I found the `homeboy-agents` `agent_task` failures split into **two distinct populations**, which matters for how this is fixed and verified:

[...]

Issue #7505, opened 2026-07-04
test infra: replace global home_lock env mutation with injectable config/path roots


chubes4:
## Problem
`test_support::home_lock()` serializes tests behind a global mutex while mutating process env (`HOME`, `XDG_DATA_HOME`, artifact root, runtime tmpdir) — `src/test_support.rs:54-96`. Usage: 931 call sites in src inline tests + 169 in tests/. This is the primary structural cause of the slow suite: it defeats test parallelism across the entire lib test binary (6,314 inline tests), compounding with wall-clock sleeps (`tests/core/daemon_test.rs:85-395`) and repeated git fixture creation (3+ duplicate builders; 251 `Command::new("git")` refs in src tests).

## Desired outcome
Core APIs accept injected config/path roots (context struct or constructor params) so tests construct isolated roots without process-env mutation. `home_lock` and `HomeGuard` are deleted; one integration test covers real env-var wiring. Shared `test_support::{fs,git::GitFixture,http}` modules replace the duplicated per-module helpers (`tests/core/rig/support.rs`, `tests/core/stack/support.rs`, inline copies in `src/commands/infra/route.rs:1003-1090`, `src/core/extension/lifecycle/mod.rs:361-563`).

Root-cause fix under #7469 / #6804. Evidence off origin/main @ 340bd902e.

chubes4:
This is the identified **root cause** for the local test-suite reliability cluster. Cross-linking the symptom issues so they're tracked together:
- **#6804** — make local suite deterministic (env/temp/ordering isolation)
- **#7469** — hour-long local runs, machine-state-dependent failures
- **#7739** — `agent_task_provider` scheduler_tests flake under parallel load
- **#7902** — reverse-broker runner tests stall the full library suite

The global `home_lock` env mutation this issue replaces is the shared substrate behind most of these. Fixing #7505 (injectable config/path root) should let several of the symptom issues close.

chubes4:
**Unblocked releases with the narrow lever; the refactor this issue describes is still the right fix.**

Context: the release pipeline was blocked — `Test` → `Release Quality Policy` → `Create GitHub Release` skipped, no releases between v0.327.3 and today.

## What was actually happening

`HomeGuard` mutates process-global env (`HOME`, `XDG_DATA_HOME`, `HOMEBOY_ARTIFACT_ROOT`, `HOMEBOY_RUNTIME_TMPDIR`, invocation runtime dir) and restores on `Drop`.

[...]

PR #11271, opened 2026-08-03
fix(test): restore Rust test parallelism now the home-root race is fixed

chubes4:
Refs #7505, #10097, #9774. Removes the workaround from #11242 now that #11266 fixed the underlying race.

## This PR is deliberately self-proving

Touching `homeboy.json` widens the changed-file test scope, so **this PR's own gate is the wide-scope run that confirms #11266 holds.**

That matters because post-merge runs could not answer it. The only completed post-fix `Test` job on main was **narrow scope — 23 tests**, scoped to changed files in `homeboy-agents`. It reported `0 failed` and zero `HOME environment variable not set`, but 23 tests prove nothing about a concurrency race.

**Read the result this way:**
- **Green** → #11266 holds under full parallelism, and the serialization is correctly gone.
- **Red** → revert *this* PR, not #11266. That fix stands on its own; the serialization can go back on independently.

## Why the workaround was never the fix

`rust_cargo_test_threads = 1` serializes test **functions**. It does nothing for threads a test spawns **inside itself** — and the failing clusters (`cook`, `promotion`, `provider::scheduler`) are exactly the ones that spawn executor threads. It held for two PRs and then stopped:

| PR | setting present | Candidate Test |
|---|---|---|
| #11242 | yes | SUCCESS |
| #11243 | yes | SUCCESS |
| #11245 | yes (verified on merge-base) | FAILURE — 199 |
| #11246 | yes (verified on merge-base) | FAILURE — 199, identical list |

Counts across runs: **126 → 68 → 125 → 199**. Timing-dependent, i.e. a live data race — and every gate in that window was green or red by chance.

## What #11266 changed

`paths::homeboy()`, `paths::homeboy_data()`, and the invocation runtime root now resolve through a `Mutex`-guarded process-local home-root override rather than an unsynchronized `getenv`, so a concurrent repoint is ordered rather than torn. Environment remains the source of truth when no override is registered, so subprocesses are unaffected.

## The cost being recovered

Serialization removed all intra-binary parallelism from the largest crate in the workspace — estimated **400–900s** for `homeboy-agents` against roughly **302s** before.

From the 2026-08-01 consolidation investigation (#11151).

### DIFF
--- a/crates/homeboy-core/src/test_support.rs
+++ b/crates/homeboy-core/src/test_support.rs
@@ -372,6 +372,32 @@ fn process_tree_snapshot(_root: u32) -> String {
     "    (process-tree snapshot is only collected on Linux)".to_string()
 }
 
+/// Isolates one test's Homeboy state: config, data, artifact, runtime, and
+/// invocation roots.
+///
+/// # Concurrency contract
+///
+/// This guard repoints `HOME` for the whole process, which is not thread-safe:
+/// `getenv`/`setenv` race, so a reader landing mid-write can observe `HOME` as
+/// *absent* and fail with "HOME environment variable not set on Unix-like
+/// system" on a host where it is plainly set.
+///
+/// `home_lock()` is held for this guard's entire lifetime, but that only
+/// serializes **writers**. Readers never take it — including worker threads a
+/// test spawns inside itself, which is why running the suite with
+/// `--test-threads=1` did not stop the failures: serializing test *functions*
+/// does nothing for threads *within* a test.
+///
+/// The hot resolvers therefore do not read the environment at all. `new()`
+/// registers a process-local override via `paths::set_home_root_override`, and
+/// `paths::homeboy()`, `paths::homeboy_data()`, and the invocation runtime root
+/// read it under a `Mutex` so a concurrent repoint is ordered rather than torn
+/// (#7505, #11266). `set_var("HOME", ..)` is retained alongside it for readers
+/// that still consult `HOME` directly and for subprocesses.
+///
+/// `Drop` clears the override **before** restoring the environment, so no
+/// window exists where the override still names a tempdir it is about to
+/// delete.
 pub struct HomeGuard {
     prior: Option<String>,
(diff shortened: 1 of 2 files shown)


chubes4:
**Not merging this yet — I could not obtain the wide-scope run it needs, and I am not going to merge it on a green that proves nothing.**

## Two attempts, both narrow

| attempt | changed files | `selected_count` | tests run | result |
|---|---|---:|---:|---|
| 1 — `homeboy.json` only | 1 | **0** | **0** | `PASSED` |
| 2 — + `test_support.rs` doc comment | 1 | 1 | **5** | `PASSED` |

Attempt 1 is filed as #11272: the gate reports `PASSED` on **zero selected tests**, so any config-only PR gets a free green. That is the same meta-pattern as #11093 and #11094 — a gate reporting success while enforcing nothing.

Attempt 2 was my correction, on the theory that touching test-support widens scope. It does not: the rust extension's changed-file routing maps it to a `rust_filter` selecting **5 tests in one binary** (`homeboy_core`, 1.9s measured against a 1500s budget). Still not an answer.

## Where that leaves the decision

**What is established:**
- #11266 fixes the race at its source, verified locally by a test that spawns 4 reader threads while repointing the home root 500 times and asserts **zero** resolution failures — exactly the shape `--test-threads=1` cannot protect.
- **Zero `HOME environment variable not set` failures in every post-#11266 run**, across all scopes observed.

**What is not established:**
- That the full `homeboy-agents` suite passes under real parallelism with the serialization removed. No wide run has executed since #11266 landed.

## Recommendation

**Leave `rust_cargo_test_threads = 1` in place until a wide run happens naturally**, then merge this. The workaround costs runtime (est. 400–900s vs ~302s for `homeboy-agents`) but it is safe, and it composes with #11266 rather than conflicting — it only applies on the `cargo` path.

The failure mode of merging early is severe and already precedented: the last time the wide suite broke, it blocked every release for a day.

Blocked on #11272 in practice — until a zero/narrow scope is distinguishable from a real pass, no PR can prove this one safe.

Issue #11147, opened 2026-08-01
40 visible commands ship with no clap `about`, and the generated CLI reference already reports it


chubes4:
Found during the 2026-08-01 consolidation investigation at `origin/main@84af2765e`.

## Evidence

`docs/reference/cli/commands/index.md:59` — generated from the clap tree — reports:
> **"40 visible commands declare no clap `about`/`long_about`"**

That is the entire `agent-task` family, i.e. the most orchestration-relevant surface in the CLI.

`docs/reference/cli/commands/observe.md:30-35` shows **7 of 8 flags as `_no help text_`**.

The generated reference correctly detects and reports this, and nothing acts on it.

## Correction to a common misreading

`docs/commands/*.md` (44 files) and `docs/reference/cli/commands/*.md` (43 files) are **not duplicates** — `docs/reference/cli/commands/index.md:1-5` declares itself generated from the clap tree and points to `docs/commands/` for narrative, and the two are cross-linked in both directions. This is one of the healthier parts of the repo. The defect is content, not structure.

## Also stale (baselined but unfixed)

- **`docs/config-precedence.md:52`** cites **six paths that do not exist**: `crates/homeboy-core/src/component/model.rs:508/528/530/534`, `crates/homeboy-core/src/extension/build/mod.rs:52`, `.../extension/bench/run/workflow.rs:106`, `.../extension/trace/run/workflow.rs:36`, `.../extension/capability.rs:294`. Glob confirms `crates/homeboy-core/src/component/model.rs` and `crates/homeboy-core/src/extension/**` **do not exist** — component model is now `crates/contracts/homeboy-component-contract/src/model.rs`, extension code is `crates/homeboy-extension/`. The entire precedence table's evidence column is wrong. Baselined at `homeboy.json:1501`.
  This doc also lists **5 areas where the team could not determine effective precedence from the code** (`:110-116`) and proposes 6 removals (`:118-157`) — **none of which have shipped.**
- **`docs/architecture/core-runner-output-parse.md`** — `:10` cites `crates/homeboy-extension-contract/src/runner_contract.rs` (actual: `crates/contracts/...`); `:43` cites `crates/homeboy-core/src/extension/test/parsing.rs` (actual: `crates/homeboy-extension/src/test/parsing.rs`); `:23` claims `parse_output` is public API, but it is `fn parse_output` — private. 3 of its 4 code pointers are wrong. Baselined at `homeboy.json:1499`.
[...]

Issue #11145, opened 2026-08-01
Release metadata readback always pays a guaranteed-404 first tier and leads its error message with the irrelevant failure


chubes4:
Found during the 2026-08-01 consolidation investigation at `origin/main@84af2765e`.

## First, the good news — verified fixed at this SHA

The stranded-release bug class **is resolved**. `crates/homeboy-release/src/release/executor/github_release/gh_cli.rs:477-547` now resolves release metadata in three tiers:
1. `repos/{repo}/releases/tags/{tag}` (404s for drafts by design)
2. `gh_release_metadata_by_id` (`:554`) — `gh release view --json databaseId` (GraphQL, **sees drafts**) → `repos/{repo}/releases/{id}` (REST, has digests). Two O(1) calls.
3. `gh_draft_release_metadata` (`:656`) — the old paginating list scan, as last resort.

The `.ok_or_else()` clobbering documented in earlier postmortems is **gone**: `metadata_fallback_error` (`:627-646`) concatenates both failures and merges both diagnostic vectors.

**Recovery is sound.** `run.rs:194-275` re-reads metadata immediately before un-drafting so a concurrent asset edit cannot race validation (`:211-212`), and `run.rs:284-403` recognises a partially-uploaded draft from its own `dist-manifest.json` and **refuses to publish** an empty (`:334`) or incomplete (`:350`) draft. `results.rs:106-136` correctly returns `step_failed` for every unfinished state, with the reasoning written down:
> *"reporting success here publishes nothing while telling every downstream consumer — and the `verify-published` release-workflow gate — that the version shipped."*

All `Ok(...)`-wrapped results are `step_failed`, not false greens.

## The residual defect

Tier 1 is **guaranteed** to fail for the post-create readback — `run.rs:642` runs immediately after `gh release create --draft` at `:592`, and the tags endpoint 404s for drafts by design.

So every successful release pays a wasted 404 round-trip, and the composed message reads:
```
gh api release metadata exited with status 1; the draft fallback also failed: <real cause>
```
**The operator's eye lands on the expected, irrelevant 404 first.**

## Fix

Pass an `expect_draft: bool` and skip tier 1 on the post-create path. On failure, lead the composed message with the tier-2 cause.

~15 LOC.

## Risk
Low.

Related: #10519, #10854, #10403

Issue #10441, opened 2026-07-27
Release tags are created before publish succeeds, leaving orphaned tags and no delivery


chubes4:
## Problem

Two tags exist with **no published release**:

| Tag | Release |
|---|---|
| `v0.320.0` | **orphaned — no release, no assets** |
| `v0.319.3` | **orphaned — no release, no assets** |
| `v0.319.2` | published, 15 assets |

`gh release list` reports **`v0.319.2` (Jul 26 22:39) as latest**, while `main` is **188 commits ahead**. Every fix merged in the last ~16 hours is undeliverable, including a fanout repair that an operator is actively blocked on (#10419, #9373).

## Mechanism

`release.yml` tags before the publish path completes. When the run fails or is superseded after tagging, the tag survives and the release does not. The next run then computes versions against a tag that has no artifacts behind it.

The immediate cause of the failures is separate (`Release Quality Policy` → `Test` job red; see below), but the *durability* defect is this: **a partial release leaves persistent state that misrepresents what shipped.**

## Why the failures kept recurring

The Release workflow's `Test` job runs the **full** suite. PR CI (`ci.yml`) uses `differential-gating: true` / changed-scope. So a PR can be green on changed scope while `main`'s full suite is red — and it only surfaces at release time, after merge. Recent failed release runs (`30272803452`, `30271131168`) both failed identically:

```
failure    Test                      -> Run Extra-Chill/homeboy-action@v2
failure    Release Quality Policy    -> Enforce release-blocking commands
```

with 9 red tests in `crates/homeboy-lab-runner/src/workspace/tests/prune.rs`, all `left: 0, right: N`.

This is the concrete cost of the gap tracked in #6807 (no post-merge integration gate on main): today, 40+ PRs merged green and the release pipeline absorbed the breakage silently.

## Expected

1. **Tag only after the publishable artifacts exist**, or make tag creation transactional with publish — a failed run must not leave a tag behind.
2. **Recover the two orphaned tags.** `workflow_dispatch` already supports `release_tag` for exactly this ("Existing tag to publish/recover without preparing a new release", `release.yml:30-33`). That path should be run for `v0.320.0`, or the orphaned tags deleted so version resolution is honest.
3. **Detect the state.** Nothing currently reports "a tag exists with no release." That check is cheap and would have surfaced this ~16 hours earlier.
4. Consider surfacing release health where operators look — `homeboy self doctor` reports binary/runtime drift but not "your last N merges never shipped."

## Evidence

[...]

Issue #10325, opened 2026-07-27
Test suite: recover ~12-15s of wall clock, delete dead fixtures, fix two stale audit-config blocks


chubes4:
## Context — the aggressive kill-list is not there

I went looking for a large vacuous/legacy/redundant-test kill-list. **It mostly does not exist** — this codebase has already been through that exercise:

- **Vacuous tests: 0.** All 33 `assert!(true)`-class hits are string-literal fixture data fed to detectors, the detectors' own fixtures, or synthetic files written to a tempdir. `blank_string_literals` (`code-audit/src/test_quality.rs:635-719`) blanks string contents before scanning, which is why these are false leads.
- **Absence-of-deleted-functionality tests: 0.** All 100+ `legacy`/`no_longer`/`deprecated`-named hits traced to **live** product code (e.g. `commands/daemon.rs:469` guards `legacy_child_recovery_migration_error()`, defined `:299`, called `:217`). The 11 `#[serde(alias)]` shims all decode externally-produced JSON. Keep them all.
- A prior audit's two flagged "vacuous" tests were a **false positive against a detector that was already fixed** (`test_quality.rs:377-379` suppresses on `use super::*;`, with a regression test at `:947-977` whose fixture is verbatim one of the flagged tests). *Action: verify the audit tooling in CI is built from current `main` — a stale detector producing confident false positives on live cancellation tests is a bigger risk than any test LOC.*

## What is actually actionable

**1. Slow tests — the real win (~12-15s), by rewriting, not deleting.**
- `tests/core/rig/check_test.rs:370-407` `test_http_check_exhausts_budget_when_nothing_ever_listens` — **~10s**, asserts `elapsed >= Duration::from_secs(9)` (`:391`) by actually waiting out a hardcoded budget. *Fix:* make the budget injectable on `CheckSpec` and assert the behavior against 200ms. Preserves the contract (`:400-406`) at 2% of the cost.
- `check_test.rs:160-185` and `:188-214` — 1.1s each, `sleep(1100ms)` for mtime granularity. *Fix:* set mtimes explicitly (`File::set_times`, stable 1.75). Faster **and** removes real flakiness on coarse-granularity filesystems.
- `tests/agent_task_promotion_provider.rs:109` invokes `rustc` at test time; `tests/core/daemon_test.rs:2404` sleeps 1.6s.

[...]

PR #11273, opened 2026-08-03
chore: delete test archaeology and fix stale docs

### LINKED ISSUE #11146: Delete 6 source-policing tests that `include_str!` their own source and grep it, duplicating configured audit policies
Found during the 2026-08-01 consolidation investigation at `origin/main@84af2765e`.

## The pattern

Six tests read their own (or a sibling's) source with `include_str!` and assert on substrings. Every one duplicates a rule that **already has a configured detector and a baseline** — except these cannot be baselined, and they break on `cargo fmt` line-wrapping and on function reordering.

## Nuke list

**1. `crates/homeboy-release/src/release/pipeline.rs:8-28`** — a 29-line file whose entire content besides two `pub use` lines is:
```rust
let files = [("executor.rs", include_str!("executor.rs")), ...];
let forbidden_terms = ["Cargo", "cargo", "Rust", "rust"];
... assert!(!runtime_source.contains(term), "release runtime core must not branch on ecosystem-specific term ...");
```
A hand-rolled, unbaselinable reimplementation of `homeboy.json`'s `core-agnostic-source` policy (`:139-146`), which already has a 678-entry baseline. **Two enforcement mechanisms for one rule.** −21 LOC.

**2. `crates/homeboy-cli/src/commands/fuzz/tests/report_tests.rs:274-292`** — 10 `include_str!`s of sibling modules:
```rust
let forbidden = ["word", "press"].concat();
assert!(!source.contains(&forbidden));
```
The string-splitting to avoid matching itself is the tell. −19 LOC.

**3. `crates/homeboy-lab-runner/src/connection.rs:2790-2804`**:
```rust
let source = include_str!("connection.rs");
let start = source.find("pub fn statuses_indexed()").expect("indexed discovery exists");
let body = &source[start..source.find("pub fn disconnect").expect("next function")];
assert!(!body.contains("recover_dead_direct_tunnel"));
```
Renaming `disconnect` panics. **Reordering functions silently yields the wrong slice and passes.** −15 LOC.

**4. `crates/homeboy-cli/src/commands/runner/status.rs:1489-1497`** — asserts a literal call expression `"full_status_projections(sessions, runner::statuses_indexed())"` appears in source; **breaks on `cargo fmt` line-wrapping.** Delete `:1489-1497` only — **keep the genuine behavioural assertion at `:1499-1513`.**

**5. `crates/homeboy-cli/src/commands/runner/controller_ancestry.rs:205-225`** — `the_probe_never_inherits_the_parent_streams`, source-grep. −21 LOC.

**6. `crates/homeboy-release/src/release/executor/github_release/gh_cli.rs:1641`** — same pattern.

## Also in this cleanup

**7. `crates/homeboy-cli/src/cli_surface/mod.rs:1242-1256`** `legacy_placement_flags_are_rejected`:
```rust
for flag in [format!("--{}-{}", "force", "hot"), format!("--{}-{}", "allow-local", "hot"), ...] {
    let Err(error) = Cli::try_parse_from(["homeboy", flag.as_str(), "bench", "example"]) else { panic!(...) };
    assert_eq!(error.kind(), clap::error::ErrorKind::UnknownArgument);
}
```
This asserts clap rejects four flags that no longer exist. **`--banana` would pass identically.** It carries zero information and can only fail by someone re-adding those exact names. The `format!("--{}-{}", ...)` obfuscation is the archaeology tell. The positive test immediately below (`:1258-1272`) is the real one. −15 LOC.

**8. `crates/homeboy-cli/src/commands/mod.rs:317-322` + `:333-345`** — a `LegacyCommandReport` twin struct defined **inside the test module** solely to prove it serializes identically to `CommandReport`. The literal-JSON assertion three lines later (`:346-349`) already fully subsumes it. Delete the twin and assertion #2; keep the literal. −20 LOC.

**9. `crates/homeboy-cli/src/command_contract/public_variants.rs:526-533`** — a "golden" test that only asserts `fixtures.join(fixture).exists()`. **A golden test that does not diff is not a golden test.** The real comparison lives in `homeboy.json:19-76` `command_status_contracts.scenarios`. (Note: `bench_contract.json` is existence-checked and has **no scenario entry** — it is never validated.) −30 LOC.

**10. `tests/cook_lab_handoff_test.rs:229-238`** `cook_help_does_not_advertise_queue_only` — spawns the binary to assert a removed flag is absent from `--help`. −10 LOC.

**11. Two fully orphaned fixtures** (zero references anywhere in source, docs, JSON, extensions, rigs):
`tests/fixtures/agent_task_output_dag_plan.json`, `tests/fixtures/sample_runtime_materialization_manifest.json`.

## The acceptance criterion

**Explicitly KEEP `tests/audit_debt_workflow_test.rs`.** It is the reference example of a *legitimate* absence test: it pins a deliberate decision (Main Guard removed on purpose in #10696), re-adding it is a plausible mistake, and it **pairs every negative assertion with a positive one** (`contains("schedule:")`, `contains("workflow_dispatch:")`, `contains("commands: review audit")`).

Use those three properties as the test for the whole pass: *(a) was the removal a decision or a cleanup, (b) is re-adding plausible, (c) does it also assert what should be there.*

## Honest accounting

~200 LOC and roughly 10-15s of wall clock. **Against a 2,700s budget this is <1%.** The value here is refactor velocity — these are precisely the tests that break "for free" during structural work (cf. #5300) — not CI time.

Related: #10325, #1969 (extensions)

chubes4:
Closes #11146. Refs #11147, #11151, #5300, #10325, #10324.

`#11147` is only partly done — see "Not done" at the bottom.

## The acceptance criterion

`tests/audit_debt_workflow_test.rs` is the reference **legitimate** absence test and is **kept untouched**. It pins a deliberate decision (Main Guard removed on purpose in #10696), re-adding it is a plausible mistake, and it pairs every negative assertion with a positive one.

Applied as: delete only if **(a)** the removal was a cleanup rather than a decision, **(b)** re-adding the thing is not a plausible mistake, and **(c)** it does not also assert what *should* be there.

**Six candidates failed that test and were kept.** They are listed below with reasons.

## Deleted

| Test | Why it goes |
|---|---|
| `release::pipeline::tests::release_runtime_core_stays_ecosystem_agnostic` | Hand-rolls the `core-agnostic-source` policy already configured in `homeboy.json` — same `crates/` scope, same `#[cfg(test)]` cutoff, and `Cargo`/`cargo`/`Rust`/`rust` are all configured terms. The policy has a baseline; this cannot be baselined. (a) duplicate, not a decision. (c) purely negative. |
| `fuzz::tests::report_tests::fuzz_command_tests_keep_core_fixtures_product_neutral` | Same policy, same verdict — 10 `include_str!`s to re-derive a configured detector. `wordpress` is a configured term. |
| `connection::indexed_inspection_tests::indexed_discovery_has_no_reconnect_or_ssh_path` | Purely negative (c), and **unsound**: it slices its own source between `pub fn statuses_indexed()` and `pub fn disconnect`, so reordering yields the wrong slice. `!body.contains("connect(")` is a substring that also matches `disconnect(`/`reconnect(`, while the slice terminates immediately before `pub fn disconnect`. The design intent survives as the inline comment inside `statuses_indexed`. **Closest call in the set** — see "Judgement calls". |
| source-grep prologue of `no_id_full_status_uses_indexed_projections_without_remote_fallback` | Asserted the literal call expression `full_status_projections(sessions, runner::statuses_indexed())` appears in source, so it breaks on `cargo fmt` line-wrapping. **Only the prologue is deleted**; the behavioural assertion after it is kept intact. |
| `legacy_placement_flags_are_rejected` | Asserted clap rejects four flags that no longer exist. `--banana` passes it identically. The `format!("--{}-{}", "force", "hot")` obfuscation existed only to dodge the repo's own CLI-surface scanner. (b) re-adding `--lab-only` as a global flag is not a plausible mistake, and the positive test immediately below covers the `--placement` replacement. |
| `LegacyCommandReport` twin struct + its comparison | A struct defined in the test module solely to prove `CommandReport` serializes identically to it. The literal-JSON assertion three lines later already subsumes it. **The literal assertion is kept.** |
| `tests/fixtures/agent_task_output_dag_plan.json`, `tests/fixtures/sample_runtime_materialization_manifest.json` | Re-verified zero references across source, docs, JSON, extensions, and rigs before deleting. |

## Kept — and why

These were on the list. They fail the criterion, so they stay.

1. **`controller_ancestry::the_probe_never_inherits_the_parent_streams`** — fails **(b)** and **(c)**. It guards a real production leak (`git`'s `fatal:` line landing in front of the `homeboy/command-result/v3` envelope), swapping `.output()` for `.status()` is exactly the plausible mistake, and it **pairs a positive assertion (`.output()` present) with the negative**. There is also no behavioural alternative: a child process's stdio configuration is not observable from the parent.

2–4. **The three `gh_cli.rs` structural guards** (`draft_resolution_prefers_the_id_lookup_over_scanning_every_release`, `an_expected_draft_skips_the_tag_endpoint_that_404s_by_design`, `the_post_create_readback_declares_that_it_expects_a_draft`) — fail **all three** properties. They pin decisions from the repeatedly-recurring release-stranding outage (#10519, #11145, #10441), re-introduction is demonstrated-plausible rather than hypothetical, and each asserts what *must* be present (`gh_release_metadata_by_id` is attempted, the list scan survives as a last resort, `expect_draft: bool` exists, `--draft` is still passed) in addition to ordering. One carries an explicit in-code justification: *"Asserted structurally because the alternative is a live `gh` call."*

5. **`public_variants::public_variant_contracts_have_discriminators_or_fixtures`** — **this one contradicts the brief.** The brief says the real comparison lives in `homeboy.json`'s `command_status_contracts.scenarios`, so the `fixtures.join(f).exists()` check is dead weight. It is not: those scenarios cover `deploy`, `release`, `runs`, and `stack` — **but not `bench_contract.json`**, which only `public_variants.rs` references. Deleting the check would leave that fixture with nothing verifying it exists at all. It is also not an absence test — it is a referential-integrity check on a hand-maintained table, and `.exists()` is a positive assertion. The "golden test that does not diff" complaint is about the name, not the value.

6. **`cook_help_does_not_advertise_queue_only`** — fails **(c)** outright. It asserts `--queue-only` is absent *and* that `--detach-after-handoff` and `--wait` are present. Those two positives are the replacement surface and are the valuable half. It spawns the binary, but the whole file already does; the marginal cost is one `--help`.

## Honest accounting

**~200 LOC and roughly 10–15s against a ~2700s budget.** That is noise. **The value is refactor velocity, not CI time** — these are the tests that break "for free" during structural work, in the same failure mode as #5300, where the breakage surfaces as a *test-binary compile error* that CI reports as "no tests ran" rather than as a named failure.

Net diff: 200 insertions, 320 deletions. Most of the insertions are the doc comments below.

## #11147 — docs

**40 clap doc comments added.** Every one of the 40 visible commands shipping without `about`/`long_about` was in the `agent-task` family, across three enums: `AgentTaskCommand` (30), `AgentTaskFanoutCommand` (6), `AgentTaskLoopCommand` (4). Wording is taken from the hand-written narrative in `docs/commands/agent-task.md` so the two surfaces agree.

Because the debt is now zero, `commands_without_help_text_are_confined_to_agent_task` becomes **`every_visible_command_ships_help_text`** and asserts the set is empty rather than pinning the shape of the remaining gap. Verified by walking all 8 `Subcommand` enums reachable under `agent_task/`: zero variants without a doc comment or `hide = true`.

**6 more** on `homeboy observe`, which rendered 6 of its 8 options as `_no help text_`.

Docs corrected:

- **`docs/config-precedence.md`** — 12 distinct cited files no longer existed. The component model and config moved to `crates/contracts/homeboy-component-contract/`, extension code to `crates/homeboy-extension/`, deploy path roots to `crates/homeboy-release/`, rig spec to `crates/homeboy-rig/`, runner loading to `crates/homeboy-lab-runner/`, tunnel entity to `crates/homeboy-tunnel/`. Every `path.rs:LINE` citation had also drifted, so citations are now file-level — the granularity this document can actually keep true. Added a status note and marked the six-item "Simplification Proposal" as unshipped, since the document read as if it described current behavior.
- **`docs/architecture/core-runner-output-parse.md`** — 3 of 4 pointers wrong, and it documented `parse_output(text, spec)` as the API. That function is **private**; the public entry point is `ParseSpec::parse(&self, text)`. The re-export is `crate::engine::output_parse`, not `crate::core::engine::output_parse`.
- **`docs/internals/developer-guide/architecture-cleanup-map.md`** — dropped the `--self-checks-json` row. That string appears **zero** times in the tree; the row was its only occurrence anywhere.
- **`docs/internals/docs-maintenance/structure.md`** — its rule banning `index.md` in subdirectories was violated by **9** files in the same tree, including the `index.md` sitting beside it and `docs/reference/cli/commands/index.md`, which the reference generator emits as a required page. Rule withdrawn and replaced with the convention actually in use. WordPress/PHP examples and the invented `api/components/configuration` tree replaced with this repo's real layout. Also recorded `docs/changelog.md` as the intentional exception to "no CHANGELOG in /docs" and noted that `docs/commands/` and `docs/reference/cli/commands/` are a deliberate narrative/generated split — **not** duplication, per #11147's note.

## Not done

**The generated tree under `docs/reference/cli/` is not regenerated here.** It needs `HOMEBOY_WRITE_CLI_REFERENCE=1 cargo test -p homeboy-cli --lib cli_surface::reference_docs`, which is a build this VPS cannot run. That tree is deliberately **not CI-gated** (`reference_docs_tests.rs` documents that both enforcing workflows were removed on purpose), so nothing breaks — but `index.md` will keep claiming 40 commands lack help text until someone runs the regenerator. The source fix is complete; only the artifact is stale.

`homeboy.json` is untouched. Two baseline entries (`docs::docs/config-precedence.md::StaleDocReference`, `docs::docs/architecture/core-runner-output-parse.md::StaleDocReference`) now suppress findings that should no longer fire. The baseline is a known-fingerprint allowlist with no unobserved-entry enforcement, so leaving them is harmless, and skipping `homeboy.json` avoids conflicting with the two sibling PRs in flight.

## Verification

`cargo fmt --check` clean. `cargo check --lib --tests` exit 0 on `homeboy-cli`, `homeboy-release`, and `homeboy-lab-runner` (every crate where tests were deleted). No `cargo build`/`cargo test`/`--workspace` per the VPS constraint.

Avoided `crates/homeboy-core/src/lib.rs` and `crates/homeboy-release/src/deploy/` for the sibling PRs — confirmed untouched in the diff. Expect a rebase.


chubes4:
Rebased onto `origin/main` after #11274 (deploy extraction) and #11276 (core relocation) landed.

**One conflict, in `crates/homeboy-release/src/release/pipeline.rs`** — and it was a genuine disagreement worth recording rather than silently resolving.

#11274 *updated* the `release_runtime_core_stays_ecosystem_agnostic` test (adjusting it after `version.rs` moved to the new `homeboy-version` crate, and deliberately relocating that half of the invariant rather than reaching across crates via `include_str!`). This PR *deletes* it.

**I took the deletion**, because the claim behind it was verified rather than assumed: `homeboy.json`'s `core-agnostic-source` policy has `include_path_contains: ["crates/"]`, applies `ignore_after_line_equals: ["#[cfg(test)]"]` — exactly what the test was doing by hand with `.split("#[cfg(test)]")` — and `cargo`/`rust`/`wordpress` are configured terms with real baseline coverage. The detector already enforces this, and unlike the test it can be baselined.

**Follow-up worth someone's judgement, not mine to decide here:** #11274 created a sibling `version_core_stays_ecosystem_agnostic` in `homeboy-version` with the same shape. It did not exist when this PR was written, so it is untouched. If the reasoning above holds it is redundant too — but I would rather flag it than delete a test another PR just deliberately placed.

Rebase verified: `cargo fmt --all --check` clean, `cargo check --lib` exit 0 on `homeboy-release` and `homeboy-cli`.

chubes4:
**Blocked on one test, and my earlier diagnosis of it was wrong. Recording what is and is not established.**

Failing: `promotion_gate_binds_a_socket_in_the_short_invocation_tmpdir_for_a_long_run_id` — `1 passed; 1 failed`, no compile errors, and **zero** `HOME environment variable not set` failures.

## What I got wrong

I attributed this to #11266 (routing the invocation runtime root through `paths::homeboy()`) and reverted that hunk in #11279. **That revert is in this branch — and the test still fails.**

Worse, #11279's own green was not evidence: it changed only `crates/homeboy-core/src/engine/invocation/runtime.rs`, and changed-scope selection does not pull in a root-package integration test. **That test never ran in the gate that "verified" the revert.** I should have checked before merging it.

The revert is still defensible on its own terms — the invocation runtime root was not where the race lived, it is already pinned by `HOMEBOY_INVOCATION_RUNTIME_DIR_ENV` in hermetic tests, and its path shape is `sockaddr_un`-constrained in a way the config root's is not. But it did **not** fix this, so the cause is elsewhere.

## What is established

- The test passed in three pre-#11266 candidate gates (#11234, #11235, #11246) and has failed in every post-#11266 one (#11273, #11276, and here).
- It is **in-process** — calls `agent_task_promotion::promote(..)` directly with no `HomeGuard`, so it resolves against CI's real `HOME`.
- It asserts `socket_path.len() < 104` (the `sockaddr_un` budget) with a deliberately ~200-char `run_id`.
- The observed failure is `status: GateFailed`, not a length assertion — so the gate itself fails before the length check.

## What is not established

Which change actually broke it. #11266's remaining surface is the `homeboy-paths` home-root override plus its use in `homeboy()` / `homeboy_data()`, and the `HomeGuard` registration. A plausible next step is checking whether the sibling test in the same binary (`promotion_provider_applies_a_typed_patch_request`, which passes) leaves an override registered — `Drop` clears it, but that is worth confirming empirically rather than by reading.

I could not reproduce locally: the test compiles a helper with `rustc` and drives the promotion gate, which needs a full build, and this host cannot run one.

## Recommendation

Do not merge this PR on a force or a retry. The docs and deletions in it are unrelated to the failure and are ready; the failing test is a **pre-existing main-level regression** that two other PRs (#11274, #11276) also hit and that merged around rather than through. It deserves its own investigation with a machine that can build and run the test.

Happy to rebase this again once that is fixed.

chubes4:
Rebased onto `main` (`98cc7876d`). The red Test gate was not this PR.

`promotion_gate_binds_a_socket_in_the_short_invocation_tmpdir_for_a_long_run_id`
failed with `path must be shorter than SUN_LEN`. That is the exact regression
fixed by #11281 (`fix(gate): keep the gate temp root short and non-symlink`),
which names this test in its commit message — #11267 canonicalized the
invocation temp directory and resolved away the short symlink alias #10829
introduced to keep gate sockets inside the `sockaddr_un` budget.

This branch was based on `98eeb844c`, one commit before #11281 landed. Nothing
in the diff goes near the gate temp root.

## One conflict, resolved deliberately

`crates/homeboy-lab-runner/src/connection.rs`. Main added
`status_read_purity_tests` in #11325 (`keep status observationally read-only`)
right where this PR deletes `indexed_inspection_tests`.

**Kept `status_read_purity_tests`, deleted only `indexed_inspection_tests`.**
Confirmed against the original commit — it removed 16 lines, only the indexed
module. `status_read_purity_tests` did not exist when this branch was cut, so it
was never in scope.

It would also survive this PR's own criteria: it pins a deliberate decision
(#11325), re-introducing lifecycle mutation into `status` is a plausible
mistake, and it is the kind of absence test the acceptance criterion protects.

Verified: `cargo fmt --all --check` clean, `cargo check -p homeboy-lab-runner`
clean.

Separately, #11367's red gate is also a pre-existing main failure —
`successful_zero_finding_routes_initialize_and_clean_evidence`, fixed in #11438.

Issue #11383, opened 2026-08-03
24 cook tests fail deterministically on main across three distinct causes, one a regression since v0.327.1


chubes4:
Follow-up to #11382, which fixed the first family (unpinned fixture initial branch) and recovered 11 of the 35 tests failing on main. These are the remaining 24.

All verified **single-threaded**, so none of this is a parallelism artifact.

## Family A — missing `aggregate.json` (**contains a real regression**)

```
Error {
  code: InternalIoError,
  message: "IO error",
  details: {
    "context": ".../.local/share/homeboy/agent-task-runs/<run_id>/aggregate.json",
    "error": "No such file or directory (os error 2)"
  }
}
```

**`cook_rejects_test_claim_without_matching_durable_gate` passed at v0.327.1 and fails on main.** That is a genuine regression, not a chronic failure.

It is also mis-reported: the test asserts on a *validation* message ("matching successful visible durable gate"), but the run now dies on an IO error long before reaching that check. The assertion failure hides the real cause — worth surfacing the error rather than only asserting a substring.

Also in this family: `historical_orphan_recipe_adoption_uses_recorded_policy_without_provider_replay`, where the record shows `lifecycle_state: "Succeeded"` alongside `reason_code: "internal.io_error"` for the same missing aggregate — a run that succeeded but whose aggregate is gone.

Regression window: `18f5f7ad2` (v0.327.1) → `origin/main`, 80 commits. Candidates touching this path: `ab7ef3287` (#11326, bind verification claims to gate evidence), `46f3965ae` (#11344), `445130421` (#11340), `e49e3cf32` (#11318), `fd5bae5cd` (#11256).

Note that `store::read_aggregate` is propagated with `?` at `lifecycle_ops.rs:1410` and `:2447`, but tolerated with `.ok()` / `if let Ok` at `:999`, `:1550`, `:1590`, `:1794`. Whether the strict sites should be strict is the question worth answering.

## Family B — `Ok(durable_failure)` where the test expects `Err` (chronic)

`cook_transport_preparation_failure_does_not_exhaust_cook_retries` asserts:

```rust
let error = run_cook(options, UnusedExecutor)
    .expect_err("transport preparation remains outside cook retries");
```

`run_cook` returns `Ok` with `status: "durable_failure"`. `durable_cook_error_report` converts post-materialization errors into the normal Cook result contract when `recipe_exists`, so the error no longer propagates.

[...]

PR #11403, opened 2026-08-04
Settle the pre-provider recipe correction contract

chubes4:
Refs #11383. You asked me to make the call, so this PR is the decision plus the code.

## The situation

`6d0495a7e` (`fix(cook): allow pre-provider recipe corrections`, 2026-07-24) changed two Cook contracts and left the tests encoding the old ones. Main has been red on them for ten days.

Established by automated bisect over the 259-commit window between a passing tree 10 days ago (`2ba66bd74`) and a failing tree 9 days ago (`d57893d13`):

```
6d0495a7ef09e2bec91cdaa8beaade42e0596914 is the first bad commit
    fix(cook): allow pre-provider recipe corrections
```

**Both behavior changes are correct. The tests are what needed to move.** Here is the reasoning, since that was the open question.

## Decision 1 — a durable failure is reported, not raised

Once a Cook owns a durable recipe, `durable_cook_error_report` converts a post-materialization error into the normal result contract instead of propagating `Err`. `6d0495a7e` makes a recipe exist in more situations, so more failures take that path.

**Keeping the conversion.** A bare `Err` carries no cook id and no legal recovery actions — exactly what an orchestrator needs in order to do anything but give up. The report carries both. Reverting to an opaque error to satisfy a test would make the failure *less* actionable.

Nothing is lost. The diagnostic survives verbatim:

```
failure_context.diagnostic =
  "Invalid argument 'runner': fixture runner is unavailable"
```

And the invariant the test is actually named for is untouched:

```rust
assert_eq!(record.metadata["provider_executions_consumed"], 0);
```

The `expect_err` was incidental to that. It now asserts through the report.

## Decision 2 — `title` is frozen at finalization, not at dispatch

Under `RecipeFreezeBoundary`'s derived ordering (`Provider < Candidate < Promotion < Gate < Finalization`), the guard `mismatch_freeze_boundary(field) <= reached` means a `title` correction on a Cook that has only reached provider dispatch is **deliberately permitted**.

**Keeping that too.** No PR exists yet at provider dispatch, so nothing immutable has been published — correcting the title before finalization is precisely the "pre-provider correction" the commit set out to allow. Freezing it earlier would forbid a safe, useful fix.

The test asserted the removed rejection message, so it now asserts acceptance plus the property worth protecting: **the provider is not re-dispatched**. Renamed, because `..._rejects_changed_inputs` no longer describes what it verifies.

## Also: stop comparing whole run records across a replay

That test asserted full `AgentTaskRunRecord` equality after an idempotent replay. Diffing the two records field by field, the *only* difference was a heartbeat timestamp inside `metadata`, moved by re-observing runner handoff state:

```
"updated_at": "...:27.701576787"   vs   "...:27.487901946"
```

Everything else identical, and `dispatches == 1` still held — so no duplicate provider work. It now compares the fields a replay must not disturb (`run_id`, `state`, `runner_job_id`, `submitted_at`); not re-dispatching is already asserted directly.

## Verification

- `cook_transport_preparation_failure_does_not_exhaust_cook_retries` — **ok**
- `orphaned_recipe_materializes_once_and_accepts_pre_provider_corrections` — **ok**

Both single-threaded, so not a parallelism artifact. Against the full 35-name failing set on main, this branch alone takes it from 0 to 2 passing; with #11382 the combined recovery is 13 of 35.

## What is left

22 tests. Several are the same durable-failure conversion surfacing through other paths (the Family A `aggregate.json` group in #11383), and now that the contract question is settled they can follow this pattern — but each needs its own verification rather than a blanket edit, since some may be masking real defects behind a stale assertion the way this one masked an IO error.

## AI Assistance

- Tool: OpenCode
- Model: Anthropic Claude
- Used for: bisected to the causing commit, probed the runtime values behind each failure, decided both contract questions, and updated the tests. Chris Huber remains responsible for every line.

### DIFF
--- a/crates/homeboy-agents/src/agent_task_service/cook_tests.rs
+++ b/crates/homeboy-agents/src/agent_task_service/cook_tests.rs
@@ -3262,10 +3262,28 @@ fn cook_transport_preparation_failure_does_not_exhaust_cook_retries() {
         options.initial_run_id = "cook-runner-exhaustion-attempt-1".to_string();
         options.max_attempts = 2;
 
-        let error = run_cook(options, UnusedExecutor)
-            .expect_err("transport preparation remains outside cook retries");
+        // A cook that already owns a durable recipe reports its failure through
+        // the normal result contract rather than an opaque `Err`. That is what
+        // `durable_cook_error_report` exists for: the caller needs the cook id
+        // and its legal recovery actions, and a bare error carries neither.
+        // The invariant this test is named for — transport preparation not
+        // consuming provider budget — is asserted below and is unaffected.
+        let reported = run_cook(options, UnusedExecutor)
+            .expect("durable transport failure is reported, not raised");
 
-        assert!(error.message.contains("fixture runner is unavailable"));
+        assert_eq!(reported.value.status, "durable_failure");
+        let diagnostic = reported
+            .value
+            .failure_context
+            .as_ref()
+            .and_then(|context| context.diagnostic.as_ref())
+            .expect("durable failure carries its diagnostic");
+        assert!(
+            diagnostic
+                .to_string()
+                .contains("fixture runner is unavailable"),
+            "transport failure must stay legible: {diagnostic}"
+        );
         let record = agent_task_lifecycle::status("cook-runner-exhaustion")
             .expect("transport failure remains inspectable");
         assert_eq!(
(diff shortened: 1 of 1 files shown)


chubes4:
Closing as superseded by #11401, which landed the same two fixes while this was open.

Verified on `origin/main`:

```
cook_transport_preparation_failure_does_not_exhaust_cook_retries ... ok
orphaned_recipe_materializes_once_and_replays_from_durable_inputs ... ok
```

#11401 reached the same two conclusions independently — the durable-failure conversion is the correct contract, and the whole-record replay comparison had to become field-wise. Good convergence; no reason to merge a second version.

One small thing this branch had that main does not: it asserted the failure stays *legible*, not just that it is durable.

```rust
let diagnostic = reported.value.failure_context.as_ref()
    .and_then(|context| context.diagnostic.as_ref())
    .expect("durable failure carries its diagnostic");
assert!(diagnostic.to_string().contains("fixture runner is unavailable"));
```

Main asserts `status == "durable_failure"` and `exit_code == 1`, which would still pass if the underlying cause were dropped from `failure_context` — and losing the cause is exactly how this test previously hid a missing-`aggregate.json` IO error behind a stale assertion. Worth adding on its own if anyone agrees; not worth a merge conflict on its own.

State of #11383 on current main: **20 of 35 passing, 14 still failing.**

PR #11417, opened 2026-08-04
Give the rotation baseline test a durable run record

chubes4:
Refs #11383. Same fixture gap as #11416, one more test.

## Fix

`rotation_preserves_uncommitted_candidate_and_dispatches_next_provider_from_clean_baseline` scheduled against `with_run_id("run-8081")` without creating that run. `reserve_provider_execution` reads the record before letting a provider execute, so every task failed admission with:

```
agent-task run record not found: run-8081
```

Adds the `submit_plan` + `mark_running` prelude the passing tests in this file already use. Verified: **14 passed** in `provider_rotation_tests`, up from 13.

## Two sibling tests I applied this to and backed out — worth reading

`rotation_adopts_only_the_explicit_portable_candidate` and `timed_out_attempt_rotates_after_recovering_a_malformed_scratch_index` have the *identical* `run record not found` failure (`run-adopt`, and the `run_id` variable). Adding the same prelude makes that error go away — and both then fail differently:

```
rotation_adopts_only_the_explicit_portable_candidate
  left: PartialRecoverable   right: Succeeded
  "provider execution stopped because its configured execution budget was exhausted"

timed_out_attempt_rotates_after_recovering_a_malformed_scratch_index
  left: Failed               right: Succeeded
```

That budget exhaustion is suspicious and may be a real defect rather than a fixture problem. `enable_rotation` sets a deliberately generous budget:

```rust
max_provider_executions: 10,
max_provider_rotations: 10,
```

Two executions should not exhaust ten. The reading worth checking is whether `reserve_provider_execution` and the rotation path each record an execution against the same budget — which would deplete it at twice the expected rate, and which would have been invisible for as long as the reservation was failing outright on the missing record.

I left those two alone rather than ship a change that trades one red for another, or paper over it by inflating a budget. Recorded here and on #11383 so the next person starts from the real question.

## AI Assistance

- Tool: OpenCode
- Model: Anthropic Claude
- Used for: traced the shared missing-run-record signature across the rotation family, fixed the one that goes green, and isolated the budget interaction the other two expose. Chris Huber remains responsible for every line.

### DIFF
--- a/crates/homeboy-agents/src/agent_task_scheduler/tests/scheduling_tests/provider_rotation.rs
+++ b/crates/homeboy-agents/src/agent_task_scheduler/tests/scheduling_tests/provider_rotation.rs
@@ -269,6 +269,13 @@ mod provider_rotation_tests {
         plan.options.rotation = Some(rotation_policy(vec![entry("fallback-backend-a")]));
         enable_rotation(&mut plan);
 
+        // Scheduling against a run id needs that run to exist durably:
+        // `reserve_provider_execution` reads the record before it will let a
+        // provider execute, so without this every task fails admission with
+        // "agent-task run record not found" instead of rotating.
+        crate::agent_tasks::lifecycle::submit_plan(&plan, Some("run-8081")).expect("submit plan");
+        crate::agent_tasks::lifecycle::mark_running("run-8081").expect("mark running");
+
         let aggregate = scheduler.run(plan);
 
         assert_eq!(


chubes4:
**Correcting the budget hypothesis in the description — I checked it and it is wrong.**

I suggested the two backed-out tests might be double-counting executions against the budget. They are not. `attach_execution_budget_evidence` does not gate anything; it only decorates an already-terminal outcome:

```rust
let exhausted = terminal_is_failure.then(|| {
    if executions_used >= budget.max_provider_executions {
        "total_executions"
    } else if rotations_used >= budget.max_provider_rotations as usize {
        "provider_rotations"
    } else {
        "same_provider_retries"      // <- fallback
    }
});
```

Since `enable_rotation` sets `max_same_provider_retries: 0`, **any** terminal failure lands on that fallback and gets an `agent_task.execution_budget_exhausted` diagnostic — regardless of whether a budget was actually the cause. Two executions against a limit of ten never exhausted anything. The diagnostic was a red herring and I should not have read it as the cause.

The real failure in `rotation_adopts_only_the_explicit_portable_candidate` is the outcome status, and the diagnostics that matter are:

```
"a non-empty base-bound patch was retained without treating it as verified or promotable"
"attempt workspace retained for lifecycle cleanup: attempt workspace has uncommitted changes"
```

`PartialRecoverable` instead of `Succeeded`, because a candidate patch is retained rather than adopted. That is candidate-adoption semantics, not budgeting, and it needs its own investigation.

Worth noting separately, since it cost me a wrong hypothesis and would cost the next person one too: **a failure with nothing to do with budgets is labelled "execution budget exhausted."** That fallback makes the diagnostic actively misleading on any non-budget failure. Probably worth its own issue.

The fix in this PR is unaffected — it stands on its own.

Issue #11898, opened 2026-08-07
disconnected runner status recommends reconcile that rejects disconnected runners


chubes4:
## Problem

The runner recovery UX emits a circular, impossible next action:

1. `homeboy runner status <runner> --full` reports the runner disconnected and explicitly recommends `homeboy runner reconcile <runner>`.
2. Running that exact command fails validation: `runner <runner> is not connected`.
3. `runner connect` directs the operator back toward recovery/reconciliation and remains blocked.

The typed status therefore advertises a command whose precondition the same status proves is false. This forces manual session-file/process spelunking and makes deterministic automation impossible.

## Expected contract

- Every emitted `next_action` is executable in the state that emitted it.
- Reconciliation needed to restore connectivity supports disconnected SSH runners through bounded remote inspection, or status recommends the actual disconnected recovery verb.
- Status, doctor, reconcile, and connect share one state-transition model and never form a circular command loop.
- Typed blockers explain whether the operator must connect, adopt a lease, refresh a binary, or supply exact evidence.

## Acceptance criteria

- Add a disconnected-runner fixture with persisted generations and stale daemon evidence.
- Assert every rendered recovery command passes command-level preflight for that fixture.
- Either allow safe disconnected reconciliation or stop recommending it while disconnected.
- Preserve fail-closed behavior when remote ownership cannot be proven.
- Cover status -> recommended action -> converged or terminal actionable result end to end.

## Existing tracker review

Closed #11746 covered retained active counts looping between connect and reconcile, but the current failure is more fundamental: status directly recommends `reconcile` while the reconcile command rejects all disconnected runners before inspecting generation evidence. The closed contract has regressed at the command-precondition boundary.

## AI assistance

OpenAI GPT-5.6 Sol (`gpt-5.6-sol`) via OpenCode exercised the emitted commands, inspected typed status evidence, searched existing issues, and drafted this report under Chris Huber's direction. Chris remains responsible for the report.

chubes4:
Stack verification: #11903 does not resolve this issue's status/reconcile contract. Its complete PR diff is limited to `crates/homeboy-lab-runner/src/homeboy_refresh.rs` and `homeboy_refresh/tests/{part_a,part_b}.rs`; it contains no change to `connection.rs`, `runner status`, or `runner reconcile`.

Reproduce the ownership check:
```sh
git diff --name-only origin/main...b08cf0ae455b2e9a68d362ddfcdd6acfee124b94
```

[...]

PR #11905, opened 2026-08-07
fix(runner): reconcile disconnected generation state

chubes4:
## Summary
- skip live terminal-job settlement when a runner is disconnected, allowing the existing bounded generation reconciliation path to run
- add a persisted disconnected-generation fixture that verifies the emitted reconcile action passes command preflight

## Tests
- cargo fmt --check
- cargo check -p homeboy-lab-runner
- cargo test -p homeboy-lab-runner disconnected_generation_reconcile_action_passes_reconcile_preflight --lib

## AI assistance
OpenAI GPT-5.6 Sol via OpenCode inspected the runner recovery flow, implemented the guarded terminal settlement, and added deterministic command-contract coverage. Chris Huber remains responsible for every line.

### DIFF
--- a/crates/homeboy-lab-runner/src/connection.rs
+++ b/crates/homeboy-lab-runner/src/connection.rs
@@ -1809,9 +1809,12 @@ pub fn reconcile_status(runner_id: &str) -> Result<RunnerStatusReport> {
     if report.connected {
         reconcile_session_metadata_with_observed_daemon(&runner, &mut report.session, true)?;
     }
-    // Terminal-job settlement was formerly coupled to status. Keep it under
-    // this explicit lifecycle owner so status remains safe to parallelize.
-    reconcile_terminal_jobs(runner_id)?;
+    // Terminal-job settlement needs a live session transport. Generation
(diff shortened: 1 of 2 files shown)


chubes4:
Follow-up verification: #11905 was merged as aacae45b before this review. I merged current main into the retained branch and aligned its test-only disconnected-status assertion with the current fail-closed `runner connect` recovery contract. The focused runner contract passed, and the historical controller test passed from a clean tree. The broader `homeboy-lab-runner` suite remains blocked by unrelated baseline failure `lab::provider_preflight::tests::parsed_provider_availability_accepts_unique_extension_alias` at `crates/homeboy-lab-runner/src/lab/provider_preflight.rs:1113`; the assertion expects a missing-selector message but current main returns a different result. The earlier all-shard CI failure also recorded `agent_task.committed_harvest_dirty_workspace`, outside #11898.

AI assistance: OpenAI GPT-5.6 Sol via OpenCode inspected CI artifacts, reconciled the branch with current main, reproduced the relevant tests, and updated the compatible test contract. Chris Huber remains responsible for every line.

Issue #12090, opened 2026-08-11
Preserve tracked safe in-tree symlinks in sealed Lab source packages


chubes4:
#12071 should omit ignored injected symlinks such as an externally targeted agent-context projection without traversing controller paths. An omit-all-symlinks implementation would also silently remove tracked safe repository symlinks, changing the runner source tree. Historical #8374/#9070 cover snapshot transport, not sealed source packages.

## Desired contract

Add a versioned, capability-negotiated source-package representation for regular files and tracked safe symlinks. Use repository tracking metadata to distinguish source content from ignored/untracked injection. Preserve only lexically repository-contained relative target text; never transfer target bytes. Non-Git source directories retain safe omission.

## Acceptance

- Tracked internal relative file and directory symlinks are represented and materialized without traversal.
- Tracked unresolved in-tree links preserve exact target text deterministically.
- Ignored/untracked external links are omitted and target bytes are never read.
- Tracked absolute or escaping links fail before runner admission with path/target policy diagnostics.
- Special files remain rejected.
- Format revision is negotiated so older runners never receive unreadable artifacts.
- Extraction validates paths and targets before creating links and cannot write outside root.
- Tests cover deterministic identity, receiver materialization, safe/unresolved/unsafe links, and injected-link omission.

## AI assistance disclosure

OpenAI GPT-5.6 Sol and GPT-5.6 Terra via OpenCode were used to audit sealed-package and snapshot symlink contracts, search existing trackers, and draft this issue. Chris Huber reviewed and remains responsible for the report.

PR #12097, opened 2026-08-11
fix(runner): omit symlinks from sealed source packages

### LINKED ISSUE #12071: Lab source packages reject injected AGENTS.md symlinks before runner staging
## Problem

Sealed Lab source-package staging rejects a source tree containing an injected `AGENTS.md` symlink before the runner receives the workload:

```text
Invalid argument source_path: source package accepts only regular files and directories
```

The rejection comes from `SourceArtifactTransfer::from_directory()`, which inspects entries without following links but currently fails the entire package on every symlink. Agent worktrees can legitimately contain ignored context-projection symlinks, so otherwise valid Lab workloads fail before execution.

## Reproduction

1. Create a clean source directory containing a regular file and `AGENTS.md` as a symlink to a readable file outside the source directory.
2. Submit a Lab workload that uses sealed source-package staging.
3. Observe package construction fail with `source package accepts only regular files and directories`.
4. Replace `AGENTS.md` with an inline regular file and observe package construction succeed.

This blocked the post-fanout Static Site Importer fixture matrix before any fixture executed. Evidence run: `ssi-post-fanout-20260811`.

## Expected

Keep link traversal prohibited while allowing safe regular-file snapshots:

- omit symlink entries without following them;
- continue rejecting non-file/non-directory special entries;
- retain deterministic ordering and package bounds;
- test that a symlink to an external file is absent from the manifest and that external target bytes are not transferred.

A future need to preserve repository symlinks should use an explicit versioned package format rather than following links during snapshot creation.

## Acceptance

- A source tree containing regular files and an injected symlink packages successfully.
- The package excludes the symlink and external target content.
- Special-file rejection remains fail-closed.
- Existing deterministic package tests pass.

## AI assistance disclosure

OpenAI GPT-5.6 Sol via OpenCode was used to investigate the failed Homeboy run, trace the source-package call path, and draft this issue. Chris Huber reviewed and remains responsible for the report.

chubes4:
Fixes #12071.

## Summary
- Omit source-tree symlink entries without resolving their targets.
- Reject a symlink source root and retain fail-closed handling for special files.
- Cover external symlink omission, target-byte exclusion, and special-file rejection.

This intentionally does not preserve tracked symlink fidelity; #12090 owns a versioned, capability-negotiated representation for that future behavior.

## Tests
- `cargo fmt --check`
- `cargo test -p homeboy-lab-runner runner_staging_operation::tests_support::source_tree_package` (3 passed)
- `cargo test -p homeboy-lab-runner` was attempted but did not complete within 10 minutes; it reported unrelated failures in existing runner/refresh/provider tests before timeout.

## AI assistance
OpenAI GPT-5.6 Sol via OpenCode used for implementation and testing; Chris Huber is responsible.

### DIFF
--- a/crates/homeboy-lab-runner/src/runner_staging_operation.rs
+++ b/crates/homeboy-lab-runner/src/runner_staging_operation.rs
@@ -59,8 +59,9 @@ pub struct SourceArtifactTransfer {
 
 impl SourceArtifactTransfer {
     /// Packages a controller-owned source tree into the versioned runner
-    /// manifest. Traversal is lexical and deterministic; links and non-files
-    /// are refused rather than crossing a controller filesystem boundary.
+    /// manifest. Traversal is lexical and deterministic; symlinks are omitted
+    /// and special files are refused rather than crossing a controller
+    /// filesystem boundary.
     pub fn from_directory(artifact_id: impl Into<String>, root: &Path) -> Result<Self> {
         fn collect(
             root: &Path,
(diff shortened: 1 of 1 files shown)


chubes4:
Superseded by #12099, which includes the #12071 fix while preserving tracked safe in-tree symlinks through a versioned, capability-negotiated package format and adds descriptor-relative no-follow traversal plus retry-safe extraction.

AI assistance disclosure: OpenAI GPT-5.6 Sol via OpenCode was used to review the overlapping implementations and verify the superseding security/fidelity contract. Chris Huber reviewed and remains responsible for this closure.

PR #12112, opened 2026-08-11
fix(deferred-workload): make worker ownership provable and reconcilable

### LINKED ISSUE #12081: fix: deferred workload workers survive completed runs and deleted worktrees
## Problem

The live Extra Chill host currently has 336 orphaned Homeboy deferred-workload worker processes, including 113 whose current working directory is a deleted DMC worktree. Several have survived for more than two days.

This was discovered during a disk-pressure audit with the root filesystem at 90% usage.

## Evidence

`pgrep -fc '/usr/local/bin/homeboy deferred-workload worker'` returns `336`.

`lsof +L1 | grep -c 'homeboy.*cwd.*deleted'` returns `113`.

Representative process:

`/usr/local/bin/homeboy deferred-workload worker --startup-token <token>`

Representative deleted cwd:

`/var/lib/datamachine/workspace/data-machine@fix-3088-claim-order-index (deleted)`

Affected deleted worktrees include completed release-lint and claim-index branches across Data Machine and Extra Chill repositories.

## Expected behavior

- Deferred workload workers terminate when their owning run reaches a terminal state.
- Worktree finalization/cleanup must not leave workers anchored to deleted directories.
- Homeboy should provide a bounded status/reconciliation command that identifies and safely terminates workers whose durable run ownership is terminal or absent.
- Worker cleanup must verify ownership through startup token/run state rather than broadly killing processes by command name.

## Operational impact

- Hundreds of long-lived processes accumulate.
- Deleted worktree directories remain referenced by process CWDs.
- Host process and inode pressure grows silently.
- Disk diagnostics become harder during an active capacity incident.

No processes were killed during this investigation.

chubes4:
Closes #12081.

## The bug

A deferred-workload worker could never prove it was the singleton owner.

`worker_is_live` proves ownership with `pid_has_ownership_token`, which on Linux reads `/proc/<pid>/environ` — the environment block the kernel copied at **execve**. The worker published its marker by calling `std::env::set_var("HOMEBOY_DEFERRED_WORKLOAD_OWNER", ...)` on itself, which that file never reflects. The probe therefore always answered false.

`cli_runtime` calls `restart_worker_if_pending()` before **every mutating command**. With liveness permanently false and pending records that never drained, each of those invocations spawned another worker (and burned the 5s liveness wait before giving up). Each spawned worker inherited the caller's working directory — a DMC worktree — and kept it open for its whole life.

That is the 336 workers and 113 deleted-worktree CWDs in the issue.

## The fix

**Ownership is set where the kernel can see it.** `ensure_worker` sets `HOMEBOY_DEFERRED_WORKLOAD_OWNER` on the child `Command` — the same shape the daemon already uses for `HOMEBOY_DAEMON_STARTUP_TOKEN`. A worker started by hand re-execs itself once with the marker, so the documented `homeboy deferred-workload worker` invocation still publishes provable ownership (`exec` keeps the pid the spawner is waiting on).

**The singleton stops pinning worktrees.** The worker runs from the Homeboy config directory. A controller-owned process that outlives its caller must not hold an ephemeral directory open.

**The worktree travels with the record instead.** Deferral records `source_directory`, the replay is dispatched from it, and it is part of the record identity — so two worktrees deferring the same `review test` are now two workloads instead of one silently-collapsed record that replayed only the first tree. A record whose worktree is gone fails, rather than replaying against whatever directory the worker happens to be standing in; that was a latent wrong-source-sync as well as a leak.

**`homeboy deferred-workload reconcile`** terminates workers that no live durable ownership backs, with `--dry-run` to inspect the decision first.

Nothing is killed by command name. The command line only nominates candidates; the disposition comes from durable state:

- no `Deferred`/`Claimed` record remains → orphaned
- no live singleton owner → orphaned
- pid is not the durable owner pid → orphaned
- the process environment does not prove its startup token → orphaned
- the token does not match the durable owner token → orphaned

Claims are released back to the queue before signaling, so the next worker picks them up immediately instead of waiting out the 60s lease. A pass is bounded at 60s with a 500ms SIGTERM grace per target; whatever it does not reach is reported under `remaining` for a re-run.

`deferred-workload` is now exempt from the startup worker restart as a family, not just `worker` — otherwise `reconcile` would spawn the worker it is about to judge.

## Operating the existing orphans

The 336 workers on the live host were started without a probe-visible marker, so reconciliation classifies all of them as orphaned and terminates them. That is the intended one-time transition: their claims are released and a correct singleton picks the work back up.

```sh
homeboy deferred-workload reconcile --dry-run
homeboy deferred-workload reconcile
```

## Verification

- `cargo clippy -p homeboy-cli --lib --tests -j2` — no new findings
- `cargo test -p homeboy-cli --lib -- deferred_workload` — 34 passed

New tests cover: the disposition matrix (only a proven owner survives; a stale token, an unprovable token, a foreign pid, an empty queue, and an absent owner are all orphaned), candidate selection and token parsing, claim release on termination, per-worktree record identity, backward-compatible loading of records written before `source_directory`, dispatch from the recorded worktree, and failing a record whose worktree was deleted.

The CLI reference is regenerated; `cleanup.md` and `runner.md` carry unrelated help-text drift the generator picked up.


chubes4:
## Second commit: repairing `route::tests::dispatch`

The first push went red on all four test shards. None of it came from the deferred-workload change — recording `source_directory` puts `route/tests/dispatch.rs` back into the `--changed-since` test scope, and that file has not been executed by CI since July. Five tests in it were broken by three unrelated merged changes.

| Test | Broken by | Since |
| --- | --- | --- |
| `managed_promotion_handoff_does_not_require_runner_side_artifact_hydration` | `--runner`/`--placement` exclusivity (#11829) | Aug 7 |
| `unmanaged_explicit_lab_handoff_keeps_runner_connection_requirements` | same | Aug 7 |
| `lab_cook_materializes_goal_and_explicit_task_as_one_durable_cell` | `--goal`/`--task` conflict (#10070) + repo inference (#11987) | Jul 25 / Aug 8 |
| `lab_cook_materializes_derived_provider_destination_and_preserves_it_for_retry` | destination repository validation (#11987) | Aug 8 |
| `detached_retry_materializes_failed_plan_and_persists_bounded_preacceptance_failure` | Lab submission boundary restored (#10855) | Jul 29 |

Note the failure mode on the first two: `Cli::parse_from` with conflicting global flags calls `std::process::exit(2)`, so the whole test binary dies mid-run. Under nextest that reads as one failed test; run as a single binary it silently truncates the suite.

Each fixture was updated to the contract that replaced the one it was written against, not to whatever made it pass:

- The promotion fixture is about its explicit runner, so it keeps `--runner`. The unmanaged-handoff fixture is about Lab intent, so it keeps `--placement lab`; its assertion moved from the `runner.not_found` code to the guarantee itself, because the requirement survived under a different code. Naming a runner there would carry the request past readiness into the offload provider, which is exactly what the fixture exists to prove cannot be skipped.
- The two Cook fixtures now configure the checkout remote and the component it maps to. A bare `git init` is not a checkout Cook can accept under #11987, and the shared helpers are in `route/tests/mod.rs` so the next fixture gets them for free.
- The one-cell fixture uses the supported `--goal` + `--prompt` pairing and is renamed accordingly.
- The retry fixture now asserts the controller-proxy binding that staging *does* write and pins the absence of the typed handoff, since #10855 deliberately moved that claim to the submission request.

143 route tests pass, 34 deferred-workload tests pass, `cargo check --workspace --tests` and clippy are clean.

Worth noting separately: shard-4's first-run failure was neither a test failure nor mine — it was a containment failure from leaked `sleep` descendants in this same module. That is a pre-existing flake shape in these detached fixtures and is not addressed here.

PR #12193, opened 2026-08-12
fix(upgrade): never suggest an unrunnable extension recovery command

### LINKED ISSUE #12181: fix(upgrade): refresh extension source before setup mutates checkout
## Problem

Running `sudo homeboy upgrade --method binary` successfully upgraded the controller from `0.327.5` to `0.338.0`, but the WordPress extension refresh was skipped.

The upgrade ran extension setup first. That setup modified tracked files in the extension source checkout:

- `wordpress/scripts/lib/validation-dependencies.sh`
- `wordpress/scripts/lint/phpstan-runner.sh`

It then attempted to update the WP Codebox source dependency and failed during checkout:

```text
error: Your local changes to the following files would be overwritten by checkout:
    package-lock.json
Please commit your changes or stash them before you switch branches.
Aborting
```

The final structured result reported `success: true`, `extensions.status: partial`, `0 updated, 1 skipped`, and both `wordpress` and `dependency-adapters` 145 commits behind with a `git pull --ff-only` recovery command. That recovery cannot succeed while setup-authored tracked modifications remain.

## Expected

- Refresh extension source before running setup that can mutate tracked files, or run setup without modifying the source checkout.
- Do not suggest a recovery command that is guaranteed to fail against the resulting dirty checkout.
- Preserve user-authored local changes; never reset or stash them implicitly.
- Report the extension failure prominently while retaining truthful controller-upgrade success.

## Environment

- Previous controller: `homeboy 0.327.5+018b5a7a52b9`
- New controller: `homeboy 0.338.0+93bc3e103459`
- Extension source: `/home/opencode/.config/homeboy/rig-packages/extra-chill-homeboy-extensions`
- Invoking upgrade required `sudo`; extension checkout belongs to `opencode`.

chubes4:
Closes #12181 (reporting half). The root cause — the `package-lock.json` checkout error itself — is fixed in the companion PR Extra-Chill/homeboy-extensions#2601.

## Correction to the issue title

The issue asks to "refresh extension source before setup mutates checkout". Investigation shows **that ordering is already correct**: `update_linked_extension()` runs clean-gate → `git fetch`/`switch`/`pull --ff-only` → shared assets → `run_setup_if_configured` (setup last). No ordering change was made here. The real mutation bug is that setup writes into its own tracked source checkout, which lives in `homeboy-extensions` and is fixed there.

What *is* wrong in this repo is the reporting.

## 1. A recovery command guaranteed to fail

`detect_unrefreshed_symlinked_extensions()` emitted `sudo -u <user> git -C <root> pull --ff-only` for every stale clone, decided **solely** by `git_commits_behind_upstream() > 0`. It never inspected whether the worktree was dirty — and `git pull --ff-only` aborts unconditionally against uncommitted changes. In the reported run both `wordpress` and `dependency-adapters` were handed a command that could not possibly work.

Now, before choosing a command, detection checks the target with the **same generated-metadata tolerance the update gate uses** (`.source-url` / `.source-revision` only), via the new `extension_update_dirty_paths()`:

- **verified clean** → `git pull --ff-only`, unchanged
- **dirty** → reported as blocked, with the offending paths in the new `dirty` / `dirty_paths` fields, and a read-only `git status` inspection as the command
- **unverifiable** → treated as blocked rather than risking a command we cannot verify

Detection stays strictly **read-only**: it may fetch and inspect, and never resets, stashes, checks out, or otherwise mutates a worktree owned by another user. Preserving user-authored local changes was an explicit requirement of the issue.

## 2. A failure buried in a count

`extensions.status: "partial"` carried only `"0 updated, 1 skipped"`. The actual error text reached a `log_status!` line and nothing else, so no machine consumer of the structured result could tell *why*.

- New `extension_skips: Vec<ExtensionUpgradeSkip>` on `UpgradeResult` carries `{ extension_id, reason }` per skip.
- The `partial` summary and the upgrade banner now name the extension and its reason (`EXTENSIONS PARTIAL: …`), so a real extension failure is prominent rather than a passing count after a success banner.
- `success` still reports the **controller** upgrade truthfully — the controller genuinely did upgrade, so it is not flipped.
- `extensions_skipped` keeps its id-only shape for existing readers.

## 3. An unactionable gate error

`is_extension_update_workdir_clean()` said only "Linked extension source repo has uncommitted changes … Use --force to proceed" without naming a single file — and `--force` is dangerous advice when the user cannot see what it would apply the update over. It now returns `Result<(), Vec<String>>` and both dirty-gate messages name the offending paths. Gate semantics are unchanged: it still refuses by default, and unknown/unreadable state is still treated as dirty so updates never proceed over unverified state.

## Compatibility

All new fields are `#[serde(default, skip_serializing_if = ...)]`, so old payloads deserialize and old consumers ignore the new keys. The `is_extension_update_workdir_clean` signature change is `pub(crate)`.

## Verification

- `cargo check --workspace --tests -j2` — clean (3m12s)

`cargo test` / `cargo build` are not run on this host (a test build consumes ~28G and has previously exhausted the root volume); the full suite runs in CI on this branch.

New tests cover the behaviour rather than just compilation: `detect_unrefreshed_in_extensions_dir()` was factored out so detection can be driven against **real local git fixtures** (a bare remote plus a clone deliberately one commit behind), asserting that a clean clone still yields `pull --ff-only` and a dirty one does not. Serialization tests assert a skip reason reaches the structured result and that the new fields are omitted when empty.

### DIFF
--- a/crates/homeboy-cli/src/commands/upgrade.rs
+++ b/crates/homeboy-cli/src/commands/upgrade.rs
@@ -375,6 +375,7 @@ mod tests {
             restart_required: false,
             extensions_updated: Vec::new(),
             extensions_skipped: Vec::new(),
+            extension_skips: Vec::new(),
             runners_updated: Vec::new(),
             runners_skipped: Vec::new(),
             extensions_unrefreshed: Vec::new(),
--- a/crates/homeboy-extension/src/lib.rs
+++ b/crates/homeboy-extension/src/lib.rs
@@ -104,8 +104,9 @@ pub use homeboy_extension_contract::{DeployArchiveInstallPolicy, DeployRequiredH
 pub use lifecycle::source_metadata::SourceMetadataRepair;
 pub use lifecycle::source_metadata::{resolve_source_url, resolve_source_url_read_only};
 pub use lifecycle::{
-    derive_id_from_url, install, install_for_component, install_with_revision, refresh, slugify_id,
-    uninstall, update, InstallForComponentResult, InstallResult, RefreshResult, UpdateResult,
+    derive_id_from_url, extension_update_dirty_paths, install, install_for_component,
+    install_with_revision, refresh, slugify_id, uninstall, update, InstallForComponentResult,
+    InstallResult, RefreshResult, UpdateResult,
 };
 pub use maintenance::{exec_tool, update_all};
 pub use manifest::{
--- a/crates/homeboy-extension/src/lifecycle/mod.rs
+++ b/crates/homeboy-extension/src/lifecycle/mod.rs
@@ -333,9 +333,9 @@ mod update;
 pub use homeboy_core::extension_update_check::{read_source_revision, read_source_url};
 #[cfg(test)]
 use update::is_extension_update_workdir_clean;
-pub use update::update;
 pub(crate) use update::write_requested_source_ref;
 pub(crate) use update::write_source_metadata;
+pub use update::{extension_update_dirty_paths, update};
 
 /// Uninstall a extension. Automatically detects symlinks vs cloned directories.
 /// - Symlinked extensions: removes symlink only (source preserved)
@@ -1065,7 +1065,7 @@ exec '{}' "$@"
             .expect("source revision metadata");
 
         assert!(
-            is_extension_update_workdir_clean(repo, &extension_dir),
+            is_extension_update_workdir_clean(repo, &extension_dir).is_ok(),
             "generated extension metadata should not block update"
         );
     }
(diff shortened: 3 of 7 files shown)


chubes4:
Rebased onto current `main` (was 26 commits behind). Two notes for reviewers:

1. The earlier `Required Gates Declaration` failure was **not** this PR: main added a `Validate reusable workflow pins resolve to commits` step (`bash .github/validate-action-pin.sh`) while the job checks out the PR head, so a branch predating that script exits 127. The rebase resolves it.
2. The `Rustfmt` gate will stay red here until #12195 merges — `cargo fmt --all -- --check` currently fails on `main` itself in two files this PR never touches (`artifact_finalization.rs`, `cook_promotion.rs`).

Re-verified post-rebase: `cargo fmt --all -- --check` clean for every file this PR owns, and `cargo check --workspace --tests -j2` passes.

chubes4:
Rebased onto `main` now that both blockers are merged (#12195 rustfmt, #12198 the git-identity test fixture that was taking down shards 2/3/4 here).

Re-verified locally on the rebased tree:
- `cargo test -p homeboy-upgrade --lib -- symlinked_extension_tests` — 6 passed, including the two new git-fixture tests, and they also pass with an empty `HOME` + `GIT_CONFIG_NOSYSTEM=1` (no global git identity), so they will not repeat the #12198 failure mode on a runner.

PR #12220, opened 2026-08-12
fix(deploy): reuse the canonical release asset for exact tag refs

### LINKED ISSUE #12215: fix(deploy): exact tag refs should reuse the canonical release asset
## Problem

`homeboy deploy <project> <component> --ref <tag>` always selects `local_build` even when `<tag>` has an authenticated reusable release asset. That can deploy bytes different from the canonical release package and then make `homeboy deploy --check` report `remote_modified` against that package.

## Production evidence

For `Extra-Chill/data-machine-events` v0.55.1:

- `homeboy deploy extrachill-site data-machine-events --ref v0.55.1 --dry-run --confirm-dangerous` reports `artifact_source: local_build` and reason `--ref deploys an exact materialized commit`.
- The reusable release asset `data-machine-events.zip` has SHA-256 `63d353a15ee86631336a70373cff61863c8286f7f07a39a9d2c9fa998225c950`.
- That canonical asset contains `build/v0.52.1-release-notes.md` through `build/v0.55.0-release-notes.md` and `inc/Blocks/EventsMap/build/images/layers.416d9136.png`, `layers-2x.8f2c4d11.png`, and `marker-icon.2b3e1faf.png`.
- The installed tree after the exact-ref deploy is missing the release notes and contains different bytes for those image paths. For example, canonical `layers.416d9136.png` hashes to `09124afeef11190e6a6f7fd0d695471eaed9aec2f8b2a9a26f3b1a5cb5458e0f`; installed hashes to `1dbbe9d028e292f36fcba8f8b3a28d5e8932754fc2215b9ac69e4cdecf5107c6`.
- `homeboy deploy extrachill-site data-machine-events --check` correctly reports those ten paths as `remote_modified` against release-package provenance.

The WordPress deploy extension already declares `deploy.archive_install`; Homeboy 0.338.0 already stages adjacent to the target, atomically swaps directory trees, removes stale target files, restores the backup on install failure, and has regression coverage for stale-file removal. The replacement primitive is not the gap.

## Root cause

`crates/homeboy-deploy/src/execution/release_plan.rs::release_artifact_plan()` returns `LocalBuild` for every requested ref before resolving whether the ref is a release tag with a reusable asset. Generated release bytes are not necessarily reproducible from the tagged source alone, so exact Git identity does not imply exact package identity.

## Expected

- When `--ref` resolves to a release tag and the matching release asset exists, reuse and verify that canonical asset using the existing run-scoped release-asset path.
- Continue materializing and locally building branches, raw SHAs, non-release refs, missing assets, and explicit `--tagged` requests according to the existing fail-closed contract.
- Bind the asset tag/commit to the resolved ref before remote mutation.
- Cover plugins/themes and any generic artifact component without ecosystem special cases.
- Preserve existing exact-ref source safety, remote transport behavior, and rollback/replacement semantics.

## Regression coverage

Add tests proving a tag-valued `--ref` with a matching release asset selects `release_asset`, while branch/SHA refs and `--tagged` continue selecting `local_build`. Verify the post-deploy canonical package/tree manifest matches the selected asset.

## Related

- Extra-Chill/homeboy-extensions#2604
- #8111

chubes4:
Closes #12215.

## Problem

`homeboy deploy <project> <component> --ref <tag>` always selected `local_build`, even when `<tag>` had a reusable release asset. **Generated release bytes are not necessarily reproducible from the tagged source alone, so exact Git identity does not imply exact package identity** — the locally rebuilt tree then reads back as `remote_modified` against the canonical package.

`crates/homeboy-deploy/src/execution/release_plan.rs::release_artifact_plan()` returned `LocalBuild` for *any* requested ref before anything resolved whether that ref named a release with a reusable asset:

```rust
if config.requested_ref_for(&component.id).is_some() {
    return local_release_artifact_plan("--ref deploys an exact materialized commit");
}
```

## Change

A tag-valued `--ref` now plans `Reuse`, **bound to the requested ref** so the deployed bytes are the ones published for the ref the user asked for, fetched through the existing run-scoped release-asset path. No second download path, no change to replacement/rollback/remote transport.

| ref | plan |
|---|---|
| `v0.55.1` (component's release namespace) | `release_asset` |
| `main`, `release/v0.55.1`, `vnext`, `v2` | `local_build` |
| raw SHA | `local_build` |
| anything, with `--tagged` | `local_build` |

## How "is this ref a release tag" is decided

Lexically, against **the component's own tag namespace** — new `component_release_tag_version` in `homeboy-version`, the exact inverse of `component_tag_name`, placed beside the tag naming contract it inverts so the two cannot drift. That crate is already the declared home for shared tag-naming primitives and is explicitly ecosystem-agnostic, so this adds **no ecosystem special cases** and covers plugins, themes, and any generic artifact component alike.

Parsing the version is load-bearing rather than decorative: it is what separates the tag `v1.2.3` from same-shaped branch names like `vnext` and `v2`. A scoped component also cannot claim a bare `vX.Y.Z` or another component's namespace.

Deciding this lexically keeps `release_artifact_plan` **pure and network-free**, which matters for two reasons: its existing in-memory test seam (`should_try_download_release_artifact`) keeps working, and branch/SHA refs never pay for a release lookup.

## How I resolved the missing-asset ambiguity

The issue lists *"missing assets"* among things that keep local-building, but also says to preserve the *"existing fail-closed contract"*. Those conflict, so I split the cases:

- A ref that is **not a release tag** (branch, raw SHA, foreign tag) selects `local_build`. That is "not a release ref", not "missing asset", and it must not start failing deploys that work today.
- A ref that **is** a release tag whose asset cannot be fetched follows the existing fail-closed contract — `release_asset_download_error` is untouched and still refuses to degrade to `local_build`.

This is deliberately **consistent with the pre-existing non-`--ref` path**, which already resolves a local tag via `latest_component_tag` and then hard-fails if that tag has no asset. A tag-valued `--ref` is now treated exactly like the tag the tag-less path would have chosen, rather than getting its own weaker rule. The reviewer-relevant consequence: deploying a release-shaped tag that has no published asset now fails closed instead of silently local-building. `--tagged` is the escape hatch, and the fail-closed error already names it.

`--tagged` moved above the ref check so an explicit local tag build always wins. Both arms were `local_build` before, so only the reported reason changes.

## Verification

- `cargo test -p homeboy-version --lib -- release_tag_version` — 5 new tests pass (root vs scoped namespaces, foreign namespaces, branches/SHAs, prerelease tags)
- `cargo test -p homeboy-deploy --lib` — the 4 new plan tests pass alongside every pre-existing release-artifact test

New deploy tests: a tag-valued `--ref` selects `release_asset` **and binds the asset URL/tag to that ref**; branch, `release/`-prefixed, `vnext` and `v2` refs select `local_build`; a raw SHA selects `local_build`; `--tagged` outranks a tag-valued ref. All hermetic — no network, no GitHub, reusing the existing in-memory seam.

## Pre-existing failures NOT introduced here

`cargo test -p homeboy-deploy --lib` reports 5 failures that fail **identically on unmodified `main`** (verified by stashing this change and re-running):

`check_tolerates_missing_canonical_package_and_dry_run_skips_resolution`, `deploy_preflight_preserves_preexisting_artifact_after_failed_build`, `equal_inserted_request_builds_once_and_cleans_only_its_payload_copy`, `failed_post_build_validation_removes_only_new_source_output`, `test_deploy_artifact_fails_when_pre_extract_cleanup_fails`

They are permission/cleanup-failure scenarios, which do not fail when the suite runs as root — host-environment artifacts rather than logic breaks.

Separately worth flagging: this crate has **parallelism-flaky tests**. A full-crate run on `main` failed 9, including four `exact_ref_*` tests that pass cleanly in isolation on that same `main`. I verified against isolation rather than trusting either full-run number.

### DIFF
--- a/crates/homeboy-deploy/src/execution/mod.rs
+++ b/crates/homeboy-deploy/src/execution/mod.rs
@@ -847,6 +847,109 @@ mod tests {
         }
     }
 
+    /// A component that is eligible for release-asset reuse, so these tests
+    /// isolate the ref/tag decision rather than the eligibility preconditions.
+    fn release_asset_component() -> Component {
+        Component {
+            id: "example".to_string(),
+            remote_url: Some("https://github.com/example/example".to_string()),
+            build_artifact: Some("build/example.zip".to_string()),
+            ..Component::default()
+        }
+    }
+
+    fn deploy_config_with_ref(requested_ref: Option<&str>, tagged: bool) -> DeployConfig {
+        DeployConfig {
+            component_ids: Vec::new(),
+            all: false,
+            outdated: false,
+            behind_upstream: false,
+            dry_run: false,
+            check: false,
+            force: false,
+            skip_build: false,
+            keep_deps: false,
+            skip_deps_hydration: false,
+            expected_version: None,
+            no_pull: false,
+            allow_stale_source: false,
+            allow_downgrade: false,
+            head: false,
+            requested_ref: requested_ref.map(str::to_string),
+            requested_refs: Default::default(),
+            resolved_refs: Default::default(),
+            preflighted_source_paths: Default::default(),
+            preflighted_component_identities: Default::default(),
+            prepared_projection: None,
+            tagged,
+            prepared_artifact: None,
+            resume_run_id: None,
+        }
+    }
+
+    /// The regression: an exact tag ref used to force `local_build`, which ships
+    /// locally rebuilt bytes instead of the canonical release package and then
+    /// reads back as `remote_modified` against it (#12215).
+    #[test]
+    fn tag_valued_ref_reuses_the_canonical_release_asset() {
+        let component = release_asset_component();
+        let config = deploy_config_with_ref(Some("v0.55.1"), false);
+
+        match release_artifact_plan(&component, &config, false, false) {
+            ReleaseArtifactPlan::Reuse { tag, url } => {
+                // The asset is bound to the requested ref, not to whatever the
+                // latest local tag happens to be.
+                assert_eq!(tag, "v0.55.1");
+                assert!(url.contains("v0.55.1"), "{url}");
+                assert!(url.ends_with("example.zip"), "{url}");
+            }
+            ReleaseArtifactPlan::LocalBuild { reason } => {
+                panic!("tag-valued --ref must reuse the release asset, got local_build: {reason}")
+            }
+        }
+    }
+
+    #[test]
+    fn branch_valued_ref_still_builds_locally() {
+        let component = release_asset_component();
+
+        for branch in ["main", "release/v0.55.1", "vnext"] {
+            let config = deploy_config_with_ref(Some(branch), false);
+            assert!(
+                !should_try_download_release_artifact(&component, &config, false, false),
+                "branch ref '{branch}' must build locally"
+            );
+        }
+    }
+
+    #[test]
+    fn raw_sha_valued_ref_still_builds_locally() {
+        let component = release_asset_component();
+        let config =
+            deploy_config_with_ref(Some("9b780f558f0e4a1b2c3d4e5f60718293a4b5c6d7"), false);
+
+        assert!(!should_try_download_release_artifact(
+            &component, &config, false, false
+        ));
+    }
+
+    /// `--tagged` is the escape hatch for "I want a local build of this tag",
+    /// including when the tag has a reusable asset, so it must outrank `--ref`.
+    #[test]
+    fn tagged_outranks_a_tag_valued_ref() {
+        let component = release_asset_component();
+        let config = deploy_config_with_ref(Some("v0.55.1"), true);
+
+        match release_artifact_plan(&component, &config, false, false) {
+            ReleaseArtifactPlan::LocalBuild { reason } => {
+                assert!(reason.contains("--tagged"), "{reason}")
+            }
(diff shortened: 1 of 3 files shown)


chubes4:
All four Test shards are red, and **none of it is this PR**. Filed as #12226.

The failures are the agent-task lifecycle/cook/fanout and audit suites in `homeboy-agents` and `homeboy-cli`. This PR touches `homeboy-version` and `homeboy-deploy` only.

Evidence:
- A **full workflow re-run reproduced shard-1's 8 failures identically**, so they are deterministic, not flaky.
- Four of them reproduce on **unmodified `main`** (`5e016085f`) locally, across both crates: `active_pinned_run_does_not_block_controller_promotion` + `adoption_green_candidate_missing_review_form_...` (agents), and `audit_detects_outliers_in_convention_group` + `diagnose_falls_back_to_the_generic_set_...` (cli). The rest are in the same suites.
- These shards only run here because touching `homeboy-version` widens the `--changed-since` scope to every dependent crate — note the ~26min shard runtime vs ~8min for a narrow-scope PR like #12223.

This PR's own scope is green: `cargo fmt --all -- --check`, `cargo check --workspace --tests -j2`, the 5 new `homeboy-version` tests, and the 4 new `homeboy-deploy` plan tests alongside every pre-existing release-artifact test.

Issue #11751, opened 2026-08-06
[EPIC] Wave 1 — CI cost and iteration speed


chubes4:
Parent epic from the 2026-08-06 tech debt investigation. Full report is on the VPS at `/var/lib/datamachine/tmp/opencode/homeboy-tech-debt-sweep-2026-08-06.md`.

**Baseline measured against `origin/main@99e15769c`:** CI median 47.7 min, p90 79.7, max 120.4. **433 billable runner-minutes (7.2 runner-hours) per PR run**, of which the 16 test shards are 352 min (81%). ~74 full-workspace Rust compilations per PR.

Wave 0 (merged/in-flight separately) removes the always-dirty build script, the missing shard failure re-raise, and the duplicated pre-test lint. That takes 433 -> ~110 runner-min. **This epic is everything after that.**

### Burn-down

- [ ] **W1-1 — 16 shards do not actually fan out.** All 16 created at the same instant, started over a 17-minute window 43-60 min later, ~one per 60-90s: they start only as other jobs free a slot. ~76 of the 120 min run is queue wait. `homeboy / Plan` (pure bash) has median 2.3 min and **max 79.7 min** for the same reason. After Wave 0 a shard costs ~13 min of which ~1.7 is tests and ~11 is fixed overhead, so 16 shards = 208 runner-min for 27 min of test work; **4 shards ≈ 71 runner-min at nearly identical wall time.** Set `test-shards: '4'` and re-measure. *(~-140 runner-min, ~-30 min queue)*
- [ ] **W1-2 — `Audit` is 34.5 min producing `no_measurement`.** Build 599s + candidate audit 711s + baseline 684s, then: *"neither the candidate nor the baseline produced comparable counts (baseline command exited 2). Nothing is known about this command."* Reports success because `no_measurement` is a warning by design. It is the longest job in CI and yields zero information on every PR. Root-cause the exit 2 (it fails identically on candidate and base, so it is environmental, not a code finding); until then move Audit to the existing weekly `audit-debt.yml` sweep. *(-34.5 min)*
[...]

Issue #11037, opened 2026-07-31
Test gate runs 'cargo test --workspace' twice: empty test scope defaults to the full workspace, and the baseline reruns it after the candidate already timed out


chubes4:
The `homeboy / Test` gate has a median duration of **43.8 minutes** (max 103). Median PR open→merge is **11.1 minutes**, so the gate structurally cannot return a verdict before the PR is gone (#10997).

This is not a slow test suite. It is the gate running `cargo test --workspace` twice.

## Causal chain

**1. A PR that changes no test files gets no scope at all.**

`changed_test_scope_env()` in `crates/homeboy-extension/src/test/mod.rs` ends with a bare `Vec::new()` (line 564). A source-only change — the common shape of a bug fix — produces no `HOMEBOY_TEST_SCOPE_KIND`, no runner args, nothing.

**2. Empty scope defaults to the maximum.**

`rust/scripts/test-runner.sh:301`:

```bash
SCOPE_KIND="$(printf '%s' "$SCOPE_JSON" | jq -r '.kind // "workspace"')"
```

Absent scope → `workspace`.

**3. `--workspace` is then appended unconditionally.**

Since #10477, `test-runner.sh:348-357` adds `--workspace` unless the scope args already name packages. An empty scope names nothing, so it always qualifies:

```
Rust test scope kind: workspace
Rust test package selection: --workspace (all members)
Rust test invocation: cargo test --manifest-path …/Cargo.toml --workspace
```

That is a full compile plus the full suite across every crate in the workspace.

**4. The differential gate then does it again for the baseline.**

## Measured, from one run

[Run 30573308954](https://github.com/Extra-Chill/homeboy/actions/runs/30573308954), job `90975250905` — a run whose job conclusion was `success`:

| phase | window | duration | outcome |
|---|---|---|---|
| `dependency_build_setup` | — | 429s | ok |
| candidate `review test` | 19:56:37 → 20:41:50 | **45 min** | `TIMED OUT (exit code 124)` |
| baseline `review test` | 20:41:57 → 21:28:47 | **47 min** | `exited 124` |

**~100 minutes of runner time. Both halves timed out. Neither produced a test count.** The gate then reported `baseline_red`, which is non-blocking, so the job went green.

Note what the second 47 minutes bought: the candidate had *already* timed out at 20:41:50, and the baseline was run anyway, to completion of its own 47-minute timeout, in order to convert an unmeasurable candidate into a non-blocking green. That time cannot change the outcome.

[...]

Issue #10997, opened 2026-07-31
Test gate is terminated mid-run in 16 of 26 runs and reports it as a red check; 1 of 26 was a real test failure


chubes4:
A killed or preempted runner is classified as a test failure. `run-homeboy-commands.sh` maps exit `0` to `pass`, exit `124` to `timeout`, and **everything else to `fail`** — so exit `143` (SIGTERM: runner shutdown, preemption, or cancellation) produces the same verdict as a genuinely red test suite.

#10639 closed the identical hole for exit `124`. `143` was left behind.

```bash
# homeboy-action scripts/core/run-homeboy-commands.sh:113-128
  if [ "${CMD_EXIT}" -eq 0 ] && [ "${STRUCTURED_OUTPUT}" = true ]; then
    RESULTS=$(... '. + {($cmd): "pass"}')
  elif [ "${CMD_EXIT}" -eq 0 ]; then
    RESULTS=$(... '. + {($cmd): "fail"}')      # missing structured output
  elif [ "${CMD_EXIT}" -eq 124 ]; then
    RESULTS=$(... '. + {($cmd): "timeout"}')   # #10639
  else
    echo "::error::homeboy ${CMD}: FAILED (exit code ${CMD_EXIT})"
    RESULTS=$(... '. + {($cmd): "fail"}')      # <-- 143 lands here
  fi
```

There is no verdict in the vocabulary that means *the command never produced a measurement*. `143` is not a measurement of anything; it is the absence of one. This is the same fail-closed-on-no-data defect class as #10685, pointing the other way.

## Reproduction: PR #10984, two attempts, two mechanisms, zero failing tests

[Run 30595400071](https://github.com/Extra-Chill/homeboy/actions/runs/30595400071). Every other check on that PR was green (Audit, Lint, Required Gates Policy, Rustfmt, Windows Compile, Workspace Tests Compile).

**Attempt 1** — job `91046579211`, `01:05:30Z → 01:42:04Z`, conclusion `cancelled`:

```
01:20:49  ##[warning]command_execution exceeded its 300s budget
01:42:02  ##[error]Process completed with exit code 143
01:42:02  ##[error]The runner has received a shutdown signal. This can happen when
          the runner service is stopped, or a manually started runner is canceled.
01:42:02  ##[end-action id=__self.run-commands;outcome=failure;conclusion=success;duration_ms=1573407]
```

[...]

PR #12233, opened 2026-08-12
Repin Test to homeboy-action v2.11.24 (skip baseline when candidate passed)

chubes4:
Delivery of Extra-Chill/homeboy-action#379 — **#11751 W1-10**.

## What it does

A command whose candidate **passed** no longer reruns its baseline at the base ref.

A baseline exists to excuse the candidate for breakage that *predates* it. If the candidate passed, there is nothing to excuse — the rerun could never change the verdict. Measured here:

| | duration |
|---|---|
| Audit baseline | **684s** |
| Lint baseline | 110s, against 137s of real work (**+80%**) |

Audit's log shows it directly: `command_execution` runs `06:43:11 → 06:55:18`, then starts **again** at `06:55:21`.

The sharded Test path already made exactly this call in `select-test-baseline.sh`. This aligns the unsharded audit/lint path with it rather than inventing new policy.

Fail-safe: absent or unparseable candidate status keeps **every** baseline. Failing, timed-out and unrecorded commands still run theirs.

## Pinning

`v2.11.24^{commit}` = `72aa2778`. The tag ref's own object is `37465fec`, an annotated tag — the shape that broke `main` in #12152. `.github/validate-action-pin.sh` (added in #12187) confirms:

```
action-pin: Extra-Chill/homeboy-action@72aa2778... is a commit
```

Workflow YAML only. 10 jobs intact.

## Note on the expected red `Test` check

This PR changes no Rust files, so `scope: auto` resolves to zero Rust tests, the extension reports `review test: pass` and writes no inventory, and `Plan candidate Test shards` fails on *"Test inventory is missing"*.

That is **pre-existing and unrelated to this change** — it is #11037 (*"empty test scope defaults to the full workspace"*) surfacing at the inventory stage. Any YAML-only PR hits it; it is normally invisible because runs get cancelled long before reaching the plan job. #12201 merged through the same red check for the same reason.

Refs #11751, #10997

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -229,18 +229,18 @@ jobs:
     name: homeboy
     needs: [pr-state, ci-capacity-admission]
     if: ${{ needs.pr-state.outputs.active == 'true' }}
-    # v2.11.23, DEREFERENCED to its commit. These tags are annotated, so the tag
+    # v2.11.24, DEREFERENCED to its commit. These tags are annotated, so the tag
     # ref's own object SHA is a tag object, which `uses:` cannot resolve — pinning
     # one fails the whole workflow at startup with zero jobs scheduled (#12153).
     # Always pin `git rev-parse v<tag>^{commit}`.
     #
-    # Carries homeboy-action#372 (jq-portable exact shard partitioning),
-    # homeboy-action#376 (a cancelled run no longer publishes a red `Test`
-    # verdict), homeboy-action#377 (the unreachable sharded baseline phase is
-    # removed, closing a latent no-enforcement green), and homeboy-action#378,
-    # which builds the test binaries once into a nextest archive instead of
-    # recompiling the workspace in every shard. All refs #10997.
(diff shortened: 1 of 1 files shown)


chubes4:
Verified the saving is strictly free, which is stronger than the PR description claims.

`apply-differential-gate.py:203`:

```python
if status not in {"fail", "timeout"}:
    continue
```

The differential gate **never consults a baseline for a passing candidate** — it skips the command before reading any baseline metadata. So the 684s Audit baseline and 110s Lint baseline were computed and then unconditionally discarded on every green run.

This is not a trade-off between speed and signal. There is no signal being given up: the value was already being thrown away downstream.

It also bounds what this does **not** fix. #11751 W1-2 reports Audit landing on `no_measurement`, which per `apply-differential-gate.py:234` requires a candidate that **failed or timed out** plus a baseline that produced no comparable counts. Those runs still execute their baseline after this change, exactly as they should. W1-2 remains open.

Issue #12226, opened 2026-08-12
Agent-task and audit test suites are red on main (8 tests, verified locally)


chubes4:
## Summary

`main` is currently red on the agent-task lifecycle / cook / fanout and audit test suites. This is **not** surfaced by most PRs because the Test shards are `--changed-since` scoped — a PR only runs these suites if its changed files pull `homeboy-agents` / `homeboy-cli` into scope.

Found while landing #12220, whose change to `homeboy-version` (a low-level crate many others depend on) widened the shard scope enough to run them. All four of that PR's shards failed with 8 tests, and the failures reproduced **identically on a full workflow re-run** — deterministic, not flaky.

## Failing tests

From #12220 shard-1 (identical on re-run):

```
homeboy-agents  agent_task_lifecycle::tests::handoff_and_proxy::active_pinned_run_does_not_block_controller_promotion
homeboy-agents  agent_task_service::cook::tests::adoption_green_candidate_missing_review_form_runs_form_only_follow_up_and_finalizes
homeboy-cli     commands::agent_task::fanout::tests::cook_batch_unknown_provider_profile_warns_without_core_defaults
homeboy-cli     commands::agent_task::tests::controller_run::controller_from_spec_resume_reconcile_stale_recovers_without_manual_cleanup
homeboy-cli     commands::agent_task::tests::lifecycle::controller_proxy_run_resumes_on_its_recorded_runner_workspace
homeboy-cli     commands::agent_task::tests::lifecycle::cook_runner_preflight_failure_is_visible_and_resumable_through_public_commands
homeboy-cli     commands::agent_task::tests::lifecycle::diagnose_falls_back_to_the_generic_set_for_an_unclassified_failure
homeboy-cli     commands::audit::tests::audit_detects_outliers_in_convention_group
```

Shard-2 additionally reported `cook_batch_does_not_warn_for_specific_backend_names_in_core` and `cook_without_gate_and_finalizing_reports_actionable_gate_error`.

## Verified pre-existing on main

Four were reproduced locally against **unmodified `main`**, in both affected crates:

```
# origin/main @ 5e016085f
cargo test -p homeboy-agents --lib -- active_pinned_run_does_not_block_controller_promotion \
    adoption_green_candidate_missing_review_form_runs_form_only_follow_up_and_finalizes
  -> test result: FAILED. 0 passed; 2 failed

cargo test -p homeboy-cli --lib -- audit_detects_outliers_in_convention_group \
[...]

PR #12237, opened 2026-08-12
fix(controller-runtime): prove pin identity by digest instead of executing it

chubes4:
Item 1 of #12226 (identity-check hardening). Stands on its own merits; the test-isolation and audit-corpus items remain open there.

## The bug

`executable_identity()` in `crates/homeboy-core/src/controller_runtime.rs` asked a candidate who it was by running it:

```rust
Command::new(path).args(["self", "identity"])
```

When that candidate is a **libtest binary**, `self` and `identity` are not a subcommand — libtest parses them as two test **name filters**, runs every test matching them, and returns their output. Hence the CI diagnostic:

```
pinned controller executable identity check returned invalid output:
running 63 tests\ntest agent_task::tests::attempt_workspace_identity...
```

Every one of those 63 tests has `identity` in its name. A controller-runtime pin is a byte-identical **copy** of the executable it was taken from, so under `cargo test` the pinned candidate is a copy of the test binary, and probing it re-enters the suite.

## The change

A candidate whose contents are identical to the running executable **is** the running executable, so its identity is the one already compiled into this process. Resolve it by digest rather than by asking:

- **Strictly stronger evidence.** Content equality beats a self-report — a binary that lies is caught by its bytes.
- **Removes the spawn**, and with it the possibility of executing something that is not a controller at all.
- **Free where it counts.** Both callers of `verify_self_identity` already hash the candidate and compare it against the pinned record immediately beforehand, and that verified digest is already threaded through as `verified_digest`. The running executable's own digest is hashed at most once, memoized.

Path equality is kept as a fast path. Anything that cannot be established answers "not us", so foreign binaries are still executed and still fail closed — the fallback is exactly the pre-existing behaviour.

Also caps the probe output carried into the error. That output is whatever an arbitrary executable wrote, and the error is serialized into durable run records; one failed probe produced a **~40 KB diagnostic that buried its own cause**. Truncation is char-boundary safe and reports the true byte count.

## What this does and does not fix

It **does not** fix the test-isolation half of #12226. Tests reaching `activate_current_generation()` still inherit a controller-runtime pin from their neighbours under `cargo test`, and run alone under nextest.

It **does** remove the reason that inheritance mattered:

```
cargo test -p homeboy-agents --lib -- active_pinned_run_does_not_block_controller_promotion
  before: FAILED. 0 passed; 1 failed
  after:  ok.     1 passed; 0 failed
```

Run in isolation — which is the nextest condition, and the reason this test is red in CI but green in a local full-suite run.

## Verification

- `cargo test -p homeboy-core --lib -- identity_probe_tests` — 7 new tests pass
- `cargo test -p homeboy-core --lib -- controller_runtime` — **failure set is byte-identical to unmodified `main`** (verified by checking out `origin/main` in the same worktree and diffing the sorted failure lists). Passing count rises 42 → 49, i.e. the 7 new tests and no regressions.

New tests cover: the running executable resolving without spawning; a **byte-identical copy** resolving without spawning (the pin case, which path equality alone would miss); a foreign `verified_digest` not being mistaken for us while our own digest still identifies us through that seam; unresolvable paths answering "not us"; and truncation behaviour including multi-byte characters.

Note the 9 pre-existing `controller_runtime` failures in that local run are `noexec /tmp` artifacts on my host — they write fake controller binaries to a temp dir and exec them — not logic breaks, and unchanged by this PR.

### DIFF
--- a/crates/homeboy-core/src/controller_runtime.rs
+++ b/crates/homeboy-core/src/controller_runtime.rs
@@ -2184,6 +2184,20 @@ fn executable_identity(path: &Path, verified_digest: Option<&str>) -> Result<Str
     if let Some(identity) = test_controller_identity(path, verified_digest) {
         return identity;
     }
+    // Never spawn our own bytes to ask who we are. A candidate whose contents
+    // are identical to the running executable *is* the running executable, so
+    // its identity is the one compiled into this process — proven by digest
+    // rather than self-reported, which is strictly stronger than asking it.
+    //
+    // This also removes a hazard: the candidate is not always a controller.
+    // Under `cargo test` the running executable is a libtest binary, and a pin
+    // taken from it is a byte-identical copy. Executing that copy makes libtest
+    // parse `self identity` as two test *name filters* rather than a
+    // subcommand, run every test matching them, and return their output as the
+    // "identity" (#12226).
+    if is_current_executable_content(path, verified_digest) {
+        return Ok(crate::build_identity::current().display);
+    }
     let output = Command::new(path)
         .args(["self", "identity"])
         .output()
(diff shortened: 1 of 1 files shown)


chubes4:
Shard-2's single failure, `controller_runtime::tests::publishing_a_pin_seeds_the_digest_its_validation_needs`, is **not this PR** — it is another instance of #12226 item 2.

```
# origin/main, unmodified, run alone:
cargo test -p homeboy-core --lib -- publishing_a_pin_seeds_the_digest_its_validation_needs
  assertion `left == right` failed: publication reads the source and the staged copy, and nothing else
    left: 3   right: 2
```

It asserts an exact digest-computation count, which only holds when the memo is already warm from **neighbouring tests**. Run alone it sees a third computation and fails. That is the nextest (process-per-test) condition, identical in shape to the failure this PR fixes. It surfaces here only because this PR touches `homeboy-core`, which pulls it into the changed-file test scope.

Shards 1, 3 and 4 pass — previously all four failed.

Pushed one amendment while confirming that: both digests are now taken with `content_hash::sha256_file` rather than `executable_digest`, deliberately staying out of that memo. The memo is keyed by observed file identity precisely so it expires when a pin is replaced underneath it — the reason pins are validated at all — and an identity question about a candidate has no business seeding a cache whose job is to notice change. Re-verified afterwards: `cargo fmt --all -- --check` and `cargo check --workspace --tests -j2` clean, 7 identity-probe tests pass, and the count above is 3 both with and without this change.

Issue #11932, opened 2026-08-08
Restore a hermetic green PR test baseline


chubes4:
## Problem

Every currently open Homeboy PR fails all 16 candidate test shards with varying failures across unrelated crates. The workflow has no full push-to-main baseline run and tests PR heads rather than the prospective merge result, so the gate cannot distinguish inherited failures from candidate regressions.

## Evidence

- Representative run: https://github.com/Extra-Chill/homeboy/actions/runs/31220882391
- Failures vary by shard across extension environment, trace canonicality, SSH runner diagnostics, workspace pruning, deploy checkout, durable promotion, and controller reconciliation tests.
- Merged PR #11919 repaired an earlier deterministic baseline set, but no post-merge full-suite main run exists because candidate jobs are skipped on close and the workflow has no main push trigger.

## Acceptance Criteria

- The full required test suite is hermetic under the CI shard environment.
- Main receives a full baseline test run whose result can be traced from candidate checks.
- PR checks validate the prospective merge result or otherwise prove compatibility with current main.
- Existing open PRs can be rerun against a demonstrated green baseline and expose only candidate-specific failures.

## AI Assistance

OpenAI GPT-5.6 Sol via OpenCode analyzed the open PR checks, Actions logs, and shard artifacts and drafted this issue. Chris Huber remains responsible for the issue and resulting changes.

chubes4:
**Triage note — this is the durable tracker for "main can be red and nobody sees it".** Closed #12914 (22 failing tests, invisible) as overtaken, and #12008 (process-tree reaping) as fixed — but the structural gap this issue names is still open and is why both existed.\n\nConfirmed at `origin/main` `5f3ae101f`: `.github/workflows/ci.yml` is still `pull_request`-only with `differential-gating: true`, and there is no `push` or `merge_group` trigger establishing a main baseline. Without one, a red main is only ever discovered by hand.\n\nThe CI-cancellation half that made #12914 invisible was fixed separately in `0934b4824` (see #13765).
---
*AI assistance disclosure: this triage was performed by Claude (Anthropic) running in OpenCode, driven by Chris Huber. The model cross-referenced the open-issue set against `origin/main` git history and current source, and every fix claim above was verified against a named commit or a `file:line` in the tree before this issue was closed.*

PR #12375, opened 2026-08-13
fix: derive child test names so a module move cannot silently disarm them

chubes4:
Refs #11932. Removes another failure from the currently-red `main` — and two silently vacuous tests that nobody was going to notice.

## What was wrong

Three tests re-invoke their own binary with `--exact <name>` to get a second process (a lock holder, a lease claimant). All three named the child with a **string literal** carrying a `core::` module prefix that the crate extraction removed.

A stale `--exact` name does not error. libtest matches zero tests, prints `running 0 tests`, and **exits 0**.

### Two passed while running nothing

```rust
child.args(["--ignored", "--exact",
    "core::runtime_promotion::tests::authorized_child_process_acquires_lease"]);
lease.authorize_subprocess(&mut child);
assert!(child.status().expect("run authorized child").success());
```

That assertion is true for *any* name, including one that addresses nothing. `authorized_child_reenters_only_its_parent_transaction` and `unrelated_process_without_capability_is_denied` were both green while the lease authorization and denial they exist to prove never executed.

### One failed loudly, and misleadingly

`daemon_operation_lock_recovers_after_owner_exits_without_drop` waits on a ready file the child never writes, then panics:

```
operation lock holder did not become ready
```

— a message describing a *lock* problem when the actual fault was a name that addressed nothing. That is the failure currently on `main`. Caught by running the parent with `--nocapture` and seeing the child report `running 0 tests`.

## The fix, in two parts

**Remove the instance.** Derive the name from `module_path!()`. libtest strips the crate segment, so `homeboy_core::daemon::daemon_test` is addressed as `daemon::daemon_test`, and any future module move carries the child name with it.

```rust
.arg(crate::test_support::harness_test_name(
    module_path!(),
    "daemon_operation_lock_recovers_after_owner_exits_without_drop",
))
```

**Remove the class.** `run_child_test` asserts the child did not print `running 0 tests` before the caller reads its exit status:

```
child test `runtime_promotion::tests::…` matched nothing, so the assertion on its
exit status would have passed without executing anything
```

`status.success()` alone is not proof of anything — the child's own summary line is the measurement.

## Verification

```
test daemon::daemon_test::daemon_operation_lock_recovers_after_owner_exits_without_drop ...
running 1 test
ok
test runtime_promotion::tests::authorized_child_reenters_only_its_parent_transaction ... ok
test runtime_promotion::tests::unrelated_process_without_capability_is_denied ... ok

test result: ok. 3 passed; 0 failed
```

Note `running 1 test` in the child — previously `running 0 tests`.

With the names fixed, both previously-vacuous tests execute their child logic and **pass**. The behaviour was correct all along; it simply was not being tested.

`cargo check --workspace --tests -j2` clean, `cargo fmt --check` clean. `run_child_test` is not unix-gated, since the two `runtime_promotion` callers are not either and the Windows compile job builds them.

## AI assistance

Claude (chubes-bot) diagnosed and wrote the fix. Chris Huber remains responsible for the change.

### DIFF
--- a/crates/homeboy-core/src/runtime_promotion.rs
+++ b/crates/homeboy-core/src/runtime_promotion.rs
@@ -1597,14 +1597,15 @@ mod tests {
         crate::test_support::with_isolated_home(|_| {
             let lease = acquire("parent", "lab").expect("parent acquires lease");
             let executable = std::env::current_exe().expect("resolve test executable");
+            let child_test = crate::test_support::harness_test_name(
+                module_path!(),
+                "authorized_child_process_acquires_lease",
+            );
             let mut child = Command::new(executable);
-            child.args([
-                "--ignored",
-                "--exact",
-                "core::runtime_promotion::tests::authorized_child_process_acquires_lease",
-            ]);
+            child.args(["--ignored", "--exact", &child_test]);
             lease.authorize_subprocess(&mut child);
-            assert!(child.status().expect("run authorized child").success());
+            let output = crate::test_support::run_child_test(&mut child, &child_test);
+            assert!(output.status.success());
         });
     }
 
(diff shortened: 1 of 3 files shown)


chubes4:
Rebased onto current `main`.

**Conflict note.** The `daemon_test.rs` stale name was fixed independently on `main` by correcting the string literal to `"daemon::daemon_test::daemon_operation_lock_recovers_after_owner_exits_without_drop"`. I kept the derived form here, because a literal is what went stale in the first place — `module_path!()` carries the name through the next module move on its own.

**The rest of this PR is untouched on `main` and is the part that matters.** Both `runtime_promotion` tests still carry the stale `core::` prefix there:

```
crates/homeboy-core/src/runtime_promotion.rs:1604  "core::runtime_promotion::tests::authorized_child_process_acquires_lease"
crates/homeboy-core/src/runtime_promotion.rs:1662  "core::runtime_promotion::tests::unrelated_child_process_is_denied"
```

Both assert `status.success()` on a child that matches zero tests and therefore exits 0, so both are still green while proving nothing. `main` has neither `harness_test_name` nor `run_child_test`.

Re-verified after the rebase:

```
test daemon::daemon_test::daemon_operation_lock_recovers_after_owner_exits_without_drop ... ok
test runtime_promotion::tests::authorized_child_reenters_only_its_parent_transaction ... ok
test runtime_promotion::tests::unrelated_process_without_capability_is_denied ... ok
test result: ok. 3 passed; 0 failed
```

`cargo check --workspace --tests -j2` clean, `cargo fmt --check` clean.

The failing Test shards are the `homeboy-cli` `agent_task` cluster tracked in #12377, which fails on `main` independently of this branch. I posted the full diagnosis there today — it reduces to two tests and a process-global fixture directory being recursively deleted mid-process.

Issue #11897, opened 2026-08-07
main has failing and flaky tests: 6 in cook::tests, 1 in http_api


chubes4:
Found while verifying unrelated branches during the tech-debt sweep — **every branch I checked showed these, and none of them caused them.**

## `homeboy-core` — 1 deterministic failure

```
test http_api::http_api_test::artifact_content_serves_encoded_artifact_store_locator ... FAILED
```
Confirmed on `origin/main` at `0cad5a542` with a clean, isolated target dir. Fails identically on branches that do not touch `http_api.rs`.

## `homeboy-agents` — 6 failures in `cook::tests`, and they are flaky

On `origin/main`:
```
cook_successful_concrete_attempt_publishes_reviewer_body
manual_preflight_intent_does_not_block_normal_cook_finalization
manual_preflight_recovers_without_a_persisted_promotion_and_rejects_tampering
recovered_cook_finalization_uses_latest_resumed_gate_contract
re_materialize_follow_up_baseline_recovers_after_worktree_deletion
replacement_gate_proof_recovers_failed_candidate_without_hiding_evidence_or_republishing
```

**They are not stable.** Consecutive runs of the same tree produced **9**, then **5**, then **6** failures, with overlapping but non-identical sets. The names point at worktree materialization, baseline re-materialization, and gate-proof recovery — all filesystem- and ordering-sensitive.

## Why this matters more than the count

I only detected these by diffing failure lists between a branch and its merge base. **A branch that genuinely broke one of these tests would be indistinguishable from the noise**, and a reviewer seeing "6 failed" on their branch would reasonably assume it was theirs.

It also interacts badly with two things the sweep just landed:

- CI shards are `--changed-since`-scoped, so a given shard may never run these.
- `#11835` bounded several poll loops with deadline-based tests. Flaky wall-clock tests in the same suite make it harder to tell a real timing regression from this background noise.

## Suggested handling

1. Fix or `#[ignore]` the deterministic `http_api` failure — a permanently-red test teaches people to ignore red.
2. Triage the `cook::tests` six as a group. If they share a fixture-isolation root cause (likely, given the names), one fix probably covers them.
3. Until then, consider a documented known-failing list so a branch author can diff against it, which is what I had to reconstruct by hand.

chubes4:
## Update: the deterministic half is fixed; the scale was understated

**#11919 fixes 8 of these.** What is left is larger and more important than this issue originally described.

### What the 6 "flaky" cook tests actually were

[...]

PR #12385, opened 2026-08-13
fix(test): cap oversized changed-scope test selections before execution

### LINKED ISSUE #12365: fix: shard oversized changed-test selections before candidate execution
## Problem

Data Machine PR #3173 changes three files but Homeboy selected 161 tests and sent them through one candidate test process. The run timed out twice at the 1,500-second budget before reporting any test counts.

Evidence:

- Workflow: https://github.com/Extra-Chill/data-machine/actions/runs/31735416840
- First candidate test: timed out after 1500s, zero test counts
- Rerun candidate test: timed out after 1500s, zero test counts
- `selected_count`: 161
- `Plan candidate Test shards`: skipped
- `Candidate Test shard`: skipped
- Failure category: `infrastructure`

The changed test set may be conservatively broad, but Homeboy must not execute an oversized set as one unsharded candidate process when its own planner/shard jobs exist.

## Acceptance criteria

- Candidate planning recognizes an oversized changed-test selection before execution.
- The candidate test set is split across bounded shards, or fails early with an actionable planning error instead of consuming the full 1,500-second budget.
- Aggregate evidence reports per-shard test counts and preserves provenance.
- Small changed-test selections continue using the direct candidate test path.
- Add regression coverage for a changed selection large enough to exceed one candidate budget.

## AI assistance

- **AI assistance:** Yes
- **Tool:** OpenCode (GPT-5.6)
- **Used for:** Diagnosing repeated Data Machine #3173 CI timeouts from Homeboy evidence.

chubes4:
Closes #12365.

## The failure

Data Machine PR #3173 changed **three files**. Changed-scope resolution selected **161 test files**, handed all of them to the WordPress extension as one `HOMEBOY_CHANGED_TEST_FILES` list, and the extension ran 159 host PHP smokes serially in a single process. The phase blew through the 1,500-second budget and reported `test phase timed out after 1500s before reporting test counts, suite incomplete` — twice, on the original run and on the rerun.

Evidence: https://github.com/Extra-Chill/data-machine/actions/runs/31735416840 — the runner logged `CHANGED_SCOPE_SUMMARY:selected=161 routed=159 excluded=2` at t+27min, well past the budget.

## Why this lands in core and not in homeboy-action

The obvious place looks like `homeboy-action`'s reusable `ci.yml`, which already has `Plan candidate Test shards`, `Candidate Test shard`, and `Aggregate candidate Test shards` jobs. It is not implementable there:

- Those jobs are gated on the static `test-shards` input (default `1`) **and** on a `homeboy/test-inventory/v1` producer that only the Rust extension implements. WordPress has none, so data-machine cannot shard today at all.
- The action's own `scripts/scope/resolve.sh` computes the raw git diff. For this PR it would have seen 3 files, not 161. The 3 → 161 expansion is a *mapping* performed by `compute_changed_test_files_with_drift_options` in this repo.
- Inventory-only mode cannot serve as a cheap probe: `run.rs` deliberately discards changed scope there (`let changed_test_files = if inventory_mode { None } else { ... }`).

The only place that knows the selection size before the extension is invoked is this crate.

## What changed

`HOMEBOY_MAX_CHANGED_TEST_FILES` — a bounded, opt-in cap on the changed-scope selection, enforced in `run_main_test_workflow_inner` after the existing zero-selection fail-closed check and before `resolve_test_command`/`build_test_runner`.

**Env channel:** the process environment, via `std::env::var`, mirroring `test_timeout()`/`HOMEBOY_TEST_TIMEOUT_SECONDS`. Chosen over `args.ci_env` because `ci_env` is populated from the resolved CI profile job spec, whereas the Homeboy Action needs to set this from a workflow step `env:` block — which the CLI process inherits directly.

**Default is off.** Unset, unparseable, or non-positive values disable the guard. Every existing consumer is unaffected; a dedicated test pins that.

## What a consumer sees

Finding rule `changed_scope_selection_exceeds_cap` (category `test-scope`, severity `error`), exit code 1, no extension invocation:

> Changed-scope test gate selected 161 test file(s), exceeding the HOMEBOY_MAX_CHANGED_TEST_FILES cap of 40: `tests/a.php`, `tests/b.php`, … and 151 more. A single extension invocation of this size would exceed the test-phase budget before reporting any test counts.

with hints to narrow the drift/changed-test mapping, run the full suite explicitly, raise the cap to at least the selected count, or split the selection across shards.

`test_scope` still carries the **full** selection, so `runs.json` and `metadata_json.test_scope.selected_files` show exactly which 161 files triggered it. Inventory mode never fires the cap.

## Coverage

Five tests in `crates/homeboy-extension/src/test/run.rs`, serialized against each other on a mutex because they mutate a process-global env var:

- oversized selection fails fast, names count and cap, preserves `test_scope`, and provably does not spawn the runner (fixture extension touches a marker file);
- at-cap selection takes the normal path and invokes the runner;
- **unset cap leaves a large selection running exactly as before**;
- malformed / `0` / negative / empty values are ignored rather than fatal;
- inventory mode ignores the cap.

## Verification

- `cargo fmt --check` — clean
- `cargo check --workspace --tests -j2` — clean (3m23s)

The suite itself runs in CI; this VPS is disk-bound and a full `cargo test` build there consumes 28 GiB.

## Deliberate follow-up

The `homeboy-action` side — a `max-changed-test-files` input on the reusable `ci.yml`, forwarded into the candidate and baseline test phases so consumers can actually set the cap — is a separate PR in `Extra-Chill/homeboy-action`. It is intentionally not part of this change.

---
🤖 Cooked with `homeboy agent-task cook` (opencode). The cook produced and applied this patch, then died in its own harness on `validation.invalid_argument: agent-emitted review_form is malformed: invalid type: sequence, expected a string`; gates were run and the PR finalized by hand.

### DIFF
--- a/crates/homeboy-extension/src/test/run.rs
+++ b/crates/homeboy-extension/src/test/run.rs
@@ -77,6 +77,7 @@ const TEST_INVENTORY_PUBLIC_FILE: &str = "homeboy-test-inventory.json";
 #[cfg(unix)]
 const MAX_TEST_INVENTORY_BYTES: u64 = 64 * 1024 * 1024;
 const DEFAULT_TEST_TIMEOUT_SECONDS: u64 = 25 * 60;
+const MAX_CHANGED_TEST_FILES_ENV: &str = "HOMEBOY_MAX_CHANGED_TEST_FILES";
 
 pub(crate) fn test_timeout() -> Duration {
     std::env::var("HOMEBOY_TEST_TIMEOUT_SECONDS")
@@ -85,6 +86,20 @@ pub(crate) fn test_timeout() -> Duration {
         .filter(|seconds| *seconds > 0)
         .map(Duration::from_secs)
         .unwrap_or_else(|| Duration::from_secs(DEFAULT_TEST_TIMEOUT_SECONDS))
+}
+
+/// Resolve the optional changed-scope selection cap.
+///
+/// Mirrors `test_timeout()`: the value is read from the process environment
+/// (what a workflow step `env:` block sets and the CLI process inherits),
+/// must parse as a positive integer, and unset, unparseable, or zero values
+/// disable the guard entirely. Every existing consumer is therefore
+/// byte-for-byte unaffected when the cap is not configured (#12365).
+fn max_changed_test_files() -> Option<usize> {
+    std::env::var(MAX_CHANGED_TEST_FILES_ENV)
+        .ok()
+        .and_then(|value| value.parse::<usize>().ok())
+        .filter(|cap| *cap > 0)
 }
 
 #[derive(Deserialize)]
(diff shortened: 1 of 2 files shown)


chubes4:
CI caught a real defect in the first push, and it is the exact class `cargo check` cannot see.

Four of the five new tests were **vacuous** — they asserted on a selection that was always empty, so the guard was never reached.

`component_relative_changed_files` strips the component's path prefix within its enclosing git repository. The test tempdir lives under `target/.test-tmp`, *inside this repo*, so the non-git fixture resolved a prefix of `target/.test-tmp/run.X/.tmpY`, failed to strip it from `tests/component_N.php`, and dropped every changed path. The workflow then took the no-tests-in-scope early return: `runner_exit_code` was `None` where the tests expected `Some(0)`, and the oversized case returned `"passed"` instead of `"failed"`.

Reproduced locally and instrumented to confirm:

```
PROBE2 prefix=Some("target/.test-tmp/run.PFdAhE/.tmp1E0XpI") rel=[]
PROBE  status=passed scope=TestScopeOutput { selected_count: 0, selected_files: [], ... }
```

Fixed in a9fc815:

- `git init` the fixture source so the component is its own repo root, where the prefix is correctly `None`;
- the unset and malformed cases now assert `selected_count` too, so a regression back to an empty selection fails loudly instead of silently passing;
- the cap lock absorbs mutex poisoning instead of turning one genuine assertion failure into a `PoisonError` cascade across every sibling;
- a `ChangedScopeCapEnv` drop guard clears the variable on panic, so a failing test cannot leak the cap into another.

All five now pass locally and genuinely exercise the guard:

```
test test::run::tests::changed_scope_selection_exceeding_the_cap_fails_before_the_runner_spawns ... ok
test test::run::tests::changed_scope_selection_at_or_below_the_cap_runs_normally ... ok
test test::run::tests::unset_changed_scope_cap_leaves_selection_behaviour_unchanged ... ok
test test::run::tests::malformed_or_non_positive_changed_scope_cap_values_are_ignored ... ok
test test::run::tests::inventory_mode_ignores_the_changed_scope_cap ... ok
```

`cargo fmt --check` and `cargo check --workspace --tests -j2` both clean.

Note on the first run's other reds: the shard failures in `homeboy-agents` and `homeboy-cli` (`agent_task`, `cli_surface`, `extension`, `infra::route`) are pre-existing main breakage tracked in #12226 / #11897 / #11932, not from this change. `Audit` reported `fail` but the job actually **cancelled** at 24m37s — its `Run ./.homeboy-action` step conclusion is `cancelled` with every other step green.

Issue #12377, opened 2026-08-13
Remaining red main: the homeboy-cli agent_task cluster, triaged (7 real, 3 CI-only, 2 root causes)


chubes4:
## Summary

After the five fixes below landed, the remaining `main` CI failures are a `homeboy-cli` cluster. This records exactly which ones are real, which are environment-dependent, and the two distinct root causes identified so far — so the next person does not repeat the triage.

## Already fixed elsewhere (not part of this cluster)

| failure | fix |
|---|---|
| 3 measurement-registry tests | #12371 |
| `build_exec_env_includes_toolchain_path` | #12372 |
| `waiver_is_bound_to_the_named_non_required_check_and_exact_head` | #12374 |
| `daemon_operation_lock_recovers_after_owner_exits_without_drop` | #12375 |
| silent executable-fixture fallthrough (whole class) | #12369 |

`normal_completion_reaps_descendants_before_collecting_output` and `bounded_output_timeout_reaps_term_resistant_descendants` pass locally with and without every change above, so they are CI-environment only and belong with #12008.

## Reproduces locally — real failures

Verified against `origin/main` on a Linux host:

```
commands::agent_task::tests::controller_run::controller_run_from_spec_reconcile_stale_recovers_without_manual_cleanup
commands::agent_task::tests::dispatch::cook_explicit_repo_skips_unrelated_portable_git_enrichment
commands::agent_task::tests::dispatch::cook_rejects_immediately_resolved_provider_destination_from_another_repository
commands::agent_task::tests::lifecycle::diagnose_next_actions_name_the_artifacts_that_were_not_produced
commands::agent_task::tests::lifecycle::execution_states_prefer_adopted_normalized_gate_outcome_over_stale_attempt_failure
commands::agent_task::tests::promotion_review::cook_preserves_successful_candidate_when_provider_response_has_wrong_schema
commands::runner::jobs::tests::dropped_transport_reconnects_to_authoritative_generation_without_replaying_events
```

Plus one CI did not report, failing in the same module and almost certainly the same cause:

```
commands::agent_task::tests::dispatch::from_spec_dispatch_defaults_replace_stale_cwd_in_snapshot_workspace
```

These are **not** ordering-dependent: they fail identically when the whole `commands::agent_task::tests::dispatch` module runs (37 passed, 3 failed) and when run individually.

## CI-only — do not chase as logic bugs

```
[...]

Issue #12345, opened 2026-08-13
Hermetic test runner unsets TMPDIR, silently routing executable fixtures to real host binaries on noexec /tmp


chubes4:
## Problem

`scripts/nextest-hermetic-test-environment.sh` — the cargo `runner` for all unix targets — does:

```sh
unset TMPDIR TMP TEMP
```

So every test binary runs with `TMPDIR` unset, and `tempfile::tempdir()` / `std::env::temp_dir()` fall back to `/tmp`.

On a host where `/tmp` is mounted `noexec`, any test that writes an **executable fixture** into its temp dir silently breaks. Bash's `PATH` search calls `access(X_OK)`, which Linux fails on a `noexec` mount, so the mock is skipped rather than refused — and the **real** binary further down `PATH` is invoked instead. The test does not error on the missing mock; it quietly talks to the real tool.

## Evidence

`tests/release_asset_completeness_test.rs` writes a mock `gh` into its temp dir. On such a host all 7 tests fail with:

```
::error::Could not read the asset inventory for v1.2.3: release not found
```

which is the *real* `gh` resolving the repo from the checkout's git remote and correctly reporting that `v1.2.3` does not exist.

The same compiled test binary, unchanged, run directly:

| invocation | result |
|---|---|
| `env -u TMPDIR -u TMP -u TEMP <bin>` (what the cargo runner does) | `7 failed` |
| `TMPDIR=/var/lib/datamachine/tmp <bin>` | `7 passed` |

Nothing about the test or the code under test is wrong. The wrapper's environment scrub is the entire delta.

## Why it matters

CI is unaffected — GitHub runners mount `/tmp` exec — so this is invisible there and will stay invisible. It lands entirely on local development, where the symptom is maximally misleading: a green CI badge and a whole test file that fails locally with an error message pointing at a release tag rather than at the environment.

This VPS mounts `/tmp` `noexec` deliberately, and `TMPDIR=/var/lib/datamachine/tmp` is set globally (systemd drop-in plus login shells) precisely to route around it. The hermetic wrapper unsets that workaround for every test binary in the workspace.

## Root cause of the scrub

[...]

Issue #12008, opened 2026-08-09
Repair shared process-tree reaping contracts in sharded tests


chubes4:
## Parent

Subtask of #11932.

## Problem

Current `main` and the combined #11962/#11964 candidate still fail a cross-crate cluster around process-group cancellation, background output holders, and descendant reaping. These contracts share the same owning process-execution primitives and should be repaired together rather than by weakening individual assertions.

## Confirmed Failures

- `homeboy-engine-primitives::command`: cancellable/completed background output-holder deadlocks and cancellation/controller-loss process-group reaping
- `homeboy-core::server::client`: normal and timed piped execution descendant reaping
- `homeboy-core::daemon`: cooperative, TERM-resistant, and escalation cancellation reaping
- `homeboy-agents::agent_task_gate`: baseline/toolchain preflight descendant reaping
- `homeboy-lab-runner::resource_metrics`: failed child-identity persistence reaping
- `homeboy-upgrade::execution`: source-upgrade timeout process-group reaping

## Evidence

https://github.com/Extra-Chill/homeboy/actions/runs/31286617424

## Acceptance Criteria

- Shared process execution closes reader joins without waiting on inherited background descriptors.
- Cancellation and timeout paths terminate and reap the complete isolated process group.
- Lifecycle/cleanup evidence is published in the intended order.
- Focused tests pass repeatedly under nextest without extending timeouts or weakening assertions.
- Platform-specific behavior remains explicit and Windows compilation stays green.

## AI Assistance

OpenAI gpt-5.6-sol via OpenCode clustered the current shard failures and drafted this issue under human direction. Chris Huber remains responsible for the report.

chubes4:
Fixed on main. The shared process-tree reaping contracts were repaired by `de7f7f655` (#14606) and `7423fd467` (#14608).\n\nNote the broader "hermetic green baseline" goal remains tracked by #11932, which is still open — `ci.yml` is `pull_request`-only with `differential-gating: true` and no `push`/`merge_group` main baseline.
---
*AI assistance disclosure: this triage was performed by Claude (Anthropic) running in OpenCode, driven by Chris Huber. The model cross-referenced the open-issue set against `origin/main` git history and current source, and every fix claim above was verified against a named commit or a `file:line` in the tree before this issue was closed.*

PR #12388, opened 2026-08-13
fix: keep the hermetic temp root outside the repository, and unbusy two provider fixtures

chubes4:
Refs #12377. `commands::agent_task::tests::dispatch`: **37 passed / 3 failed → 40 passed**.

Three fixes. The first is a regression I introduced in #12369 and should land quickly.

## 1. The hermetic temp root moves back out of the repository

#12369 put it under `target/`, reasoning that cargo runs test binaries from there so it is exec-capable by construction. That reasoning was sound and the conclusion was still wrong: **a temp workspace nested inside a git checkout is not equivalent to one in `/tmp`.** Code that walks up looking for a repository root escapes the temp directory and finds the real one.

```
assertion `left == right` failed
  left:  "/var/lib/datamachine/workspace/homeboy@…"          <- the real checkout
 right:  "/var/lib/datamachine/workspace/homeboy@…/target/.test-tmp/run.DxBZIY/.tmpTcOhqP"
```

`from_spec_dispatch_defaults_replace_stale_cwd_in_snapshot_workspace` resolved its isolated workspace to the homeboy checkout itself.

The root is now a system temp directory **proven executable by running a script from it** — preferring the inherited `TMPDIR`, then `/tmp`, `/var/tmp`, `/dev/shm`. `target/` remains only as a last resort for a host where nothing else can execute at all.

The runner exports `HOMEBOY_TEST_TMP_SOURCE=system|in_tree` so the test can assert *the preference held* rather than hard-coding a location that is only correct on some hosts:

```rust
if field("source=") == "system" {
    assert!(!Path::new(&tmpdir).starts_with(workspace), …);
}
```

Cost: ~7ms per invocation on a host whose `/tmp` is noexec (two probes); less where the first candidate works. The existing clean-exit timing guard still passes.

## 2. `NamedTempFile` cannot host an executable fixture

Two provider fixtures were built with `NamedTempFile`, which keeps a **write descriptor open for the whole lifetime of the value**. Linux refuses to `execve` a file any process still holds open for writing:

```
worktree provider `fixture` resolve command could not start: Text file busy (os error 26)
```

So `cook_rejects_immediately_resolved_provider_destination_from_another_repository` got ETXTBSY instead of the repository mismatch it asserts on. `into_temp_path()` closes the descriptor while keeping the file and its self-deletion.

### The same bug was hiding a second test in plain sight

`cook_cwd_is_authoritative_when_provider_lookup_times_out` was **green while its provider never executed**. Its assertion holds whether the lookup times out *or* fails to start, so a passing test proved nothing about timeouts. It now actually runs the sleeping provider.

That one was not on any failure list and would not have been found by chasing red tests.

## 3. A test that never passed

`cook_explicit_repo_skips_unrelated_portable_git_enrichment` was written 2026-08-12 — **ten days after** #11241 (2026-08-02) made `--task-url` mandatory whenever `--to-worktree` is omitted. It calls `resolve_cook_destination` without one, so it was refused before reaching the identity enrichment its timing assertion exists to measure. Every sibling test in the file already passes `--task-url`; this one now matches.

## Verification

- `commands::agent_task::tests::dispatch` — **40 passed, 0 failed**
- `cargo test -p homeboy-core --lib hermetic_runner` — 4 passed
- `cargo test --test release_asset_completeness_test` — 10 passed (the original #12345 case still holds)
- `cargo check --workspace --tests -j2` clean, `cargo fmt --check` clean

## Still open in the cluster

`controller_run` and `lifecycle` failures remain and have a different cause — the fixture executor never receives a request (`expect("provider saw request")`). Not a temp-file or argument issue. Tracked in #12377.

## AI assistance

Claude (chubes-bot) diagnosed and wrote the fixes. Chris Huber remains responsible for the change.

### DIFF
--- a/crates/homeboy-cli/src/commands/agent_task/tests/dispatch.rs
+++ b/crates/homeboy-cli/src/commands/agent_task/tests/dispatch.rs
@@ -1105,13 +1105,20 @@ fn cook_cwd_is_authoritative_when_provider_lookup_times_out() {
         })
         .expect("register linked worktree");
 
-        let provider = tempfile::NamedTempFile::new().expect("provider file");
-        std::fs::write(provider.path(), "#!/bin/sh\nsleep 2\n").expect("write provider");
-        let mut permissions = std::fs::metadata(provider.path())
+        // `into_temp_path` closes the write handle. A `NamedTempFile` keeps one
+        // open for its whole lifetime, and Linux refuses to `execve` a file that
+        // any process still holds open for writing -- ETXTBSY, surfaced as
+        // "Text file busy (os error 26)". The file stays alive and still
+        // self-deletes; only the descriptor goes away.
+        let provider = tempfile::NamedTempFile::new()
+            .expect("provider file")
+            .into_temp_path();
+        std::fs::write(&provider, "#!/bin/sh\nsleep 2\n").expect("write provider");
+        let mut permissions = std::fs::metadata(&provider)
             .expect("provider metadata")
             .permissions();
         permissions.set_mode(0o755);
-        std::fs::set_permissions(provider.path(), permissions).expect("make provider executable");
+        std::fs::set_permissions(&provider, permissions).expect("make provider executable");
         let mut config = homeboy::core::defaults::HomeboyConfig::default();
         config.worktree_providers.insert(
             "timeout".to_string(),
@@ -1122,7 +1129,7 @@ fn cook_cwd_is_authoritative_when_provider_lookup_times_out() {
                 lookup_timeout_ms: 1,
                 lookup_output_limit_bytes: 64 * 1024,
                 commands: homeboy::core::defaults::WorktreeProviderCommands {
-                    resolve: Some(vec![provider.path().display().to_string()]),
+                    resolve: Some(vec![provider.display().to_string()]),
                     ..Default::default()
                 },
                 list_result_mapping: None,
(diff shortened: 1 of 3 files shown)


chubes4:
Rebased onto current `main` and re-verified.

**Conflict resolution note.** `cook_explicit_repo_skips_unrelated_portable_git_enrichment` was fixed independently on `main` while this was open, using `--to-worktree` where I had used `--task-url`. I took `main`'s version and dropped mine. Theirs is the better fix: `--to-worktree` makes `resolve_cook_destination` return immediately after `normalize_cook_repository_identity`, so the timing assertion measures *only* the enrichment step this test is about, instead of also covering worktree-provider lookup and branch derivation. Same defect, tighter isolation.

The other two fixes in this PR are unaffected and still needed — the temp-root regression and the two `NamedTempFile` ETXTBSY fixtures.

Re-verified after the rebase:
- `commands::agent_task::tests::dispatch` — **40 passed, 0 failed**
- `cargo check --workspace --tests -j2` clean, `cargo fmt --check` clean

**On the two failing `hermetic_runner_reaps_descendants_*` tests:** they are not from this branch. I A/B tested it directly — the tests load the runner from disk at runtime, so swapping the script needs no rebuild:

| script | result |
|---|---|
| `origin/main`'s | `0 passed; 2 failed` |
| this branch's | `0 passed; 2 failed` |

Identical. They are the same process-reaping flakes tracked in #12008, which also fail on `main` and on CI independently of any change here. A standalone probe of both scripts against a TERM-resistant descendant reaps correctly in both, so the script logic is not the cause.

chubes4:
Measured this locally against `main` because it bears directly on why homeboy has not published a release since `v0.345.0` — release runs keep failing at `Release Quality Policy` because the `Test` gate is red.

Same command, same machine, `cargo test -p homeboy-extension --lib -j2`, full suite, repeated:

| branch | failures per run |
|---|---|
| `main` @ `05bde7217` | **20, 20, 1, 2** |
| this PR @ `5d0f60239` | **0, 1, 2, 1, 0** |

So this takes `homeboy-extension` from "1–20 failures, wildly variable" to "usually green".

It also fixes a class I kept tripping over while working on #12392 and #12394. On `main`, three tests fail deterministically here:

```
test::run::tests::inventory_mode_publishes_valid_producer_output_for_an_empty_changed_scope
test::run::tests::inventory_mode_runs_the_producer_for_an_empty_changed_scope_and_fails_without_evidence
test::run::tests::workspace_fingerprint_matches_python_producer_for_universal_newlines
```

because their fixtures write a `Cargo.toml` under `target/.test-tmp`, which the enclosing cargo workspace then claims as a member. On this branch all 51 `test::run::` tests pass. I had to explain those three away as "pre-existing local-env" in two separate PRs — this removes the need.

The same root cause bit me in application code, not just fixtures: in #12392 the drift tests silently selected *zero* files because `component_relative_changed_files` resolved a component prefix of `target/.test-tmp/run.X/.tmpY` — the temp root being inside the repo made four tests vacuous rather than failing. Worth knowing that this bug class produces false greens, not just red.

## One residual flake, pre-existing

A single `bench::run::tests::*` test fails intermittently on this branch — a different one each time:

```
panicked at crates/homeboy-extension/src/bench/run/mod.rs:241:40:
run dir: acquire runtime cleanup lock
  .homeboy-test-tmp/run.IGGuM9/hb-test-836632-dppbz6/.cleanup.lock
  No such file or directory (os error 2)
```

The path is keyed by process id while the failing thread differs each run, so sibling threads share that per-process runtime temp root and one removes it while another is still acquiring the lock.

**This is not introduced by this PR** — `component_script_bench_env_serializes_empty_passthrough_args` failed the same way on `main` in one of the runs above. It is a separate shared-parent cleanup race and shouldn't block merging this.

Recommend landing. It is a clear net improvement on a gate that is currently blocking every release.

## AI assistance

- **AI assistance:** Yes
- **Tool:** OpenCode
- **Used for:** Repeated local A/B of the full `homeboy-extension` suite against `main`.

PR #12603, opened 2026-08-16
Give the lifecycle read surface explicit stores

chubes4:
## Summary

`lifecycle_ops.rs` exposes ~104 public free functions, **90 without an explicit-store sibling**, over ~140 ambient `store::` calls that each resolve `default_store()` — literally `AgentTaskLifecycleStore::from_current_environment()`.

That is what forces the ~2,712 `with_isolated_home` call sites: **a test cannot read state from its own root unless the production read path accepts a store.**

90 is too big for one change. This slice takes the **pure-read subset**: 23 public siblings plus `resolve_run_id`.

Each existing function keeps its signature, doc comment, and input transformation, and delegates through the ambient store. `sanitize_run_id` is preserved verbatim wherever the original applied it.

## `status()` is not a read

The most useful thing this slice established. `status_with_options_inner`:

- calls `store::write_record` at **~12 sites**
- takes an advisory file lock via `reconcile_deferred_candidate`
- runs `expire_unaccepted_lab_handoff`, `reconcile_controller_job_cancellation`, `record_terminal_artifact_projection`, `enqueue_terminal_continuation`

It is the single largest mutation path in the file. So `status`, `status_with_options`, `exact_status`, and everything built on them — `aggregate_source`, `run_status`, `list_records`, `list_records_with_health` — are **out**.

Worth knowing: `AgentTaskLifecycleStore::aggregate_source_exact` already exists as the rooted, non-reconciling counterpart.

## `resolve_run_id_in_store` was mandatory, not optional

Six siblings resolve a Cook alias through the durable cook index. One that took a store and then resolved the alias against the environment would be exactly the **same-kind shadowing** blind spot #12595 found — a sibling that ignores the store it was handed, which the #12591 scanner structurally **cannot** see.

## Everything else evaluated and excluded

| function(s) | why |
|---|---|
| `status`, `status_with_options`, `exact_status`, `_inner` | writes at ~12 sites, takes a lock — see above |
| `aggregate_source`, `run_status`, `list_records*` | call `status()` |
| `load_plan_for_execution` | **writes** — migrates a legacy execution-budget envelope and re-persists. Its own doc comment says it is the only read path allowed to upgrade |
| `validate_controller_runtime`, `pinned_runtime_for_mutation` | persist a migrated pin via `write_record` |
| `runner_pinned_runtime_for_mutation` | genuinely read-only, but its sibling writes — rooting one alone leaves a misleading asymmetric pair |
| `require_detached_cook_handoff_fence_open` | read-only, but a cancellation fence |
| `durable_notification_route`, `cook_terminal_notification_outcome` | read-only, but notification family |
| `claim_*`, `expire_*`, `heartbeat_*`, `*_cancel*` | coordinate and mutate |
| `select_cook_candidate` | **already had** a sibling |
| 8 others | pure functions over an already-loaded record/plan — no store to inject |

## Store methods

**Reused as-is:** `read_record`, `read_record_bounded`, `read_cook_index`, `cook_index_exists`, `record_exists`, `read_aggregate`, `read_aggregate_bounded`, `read_controller_plan`, `aggregate_path`.

**Added 6**, because the ambient shims reached straight for `ObservationStore::open_readonly()` / `open_initialized_for_lifecycle()` rather than the store's own roots: `record_exists_readonly`, `read_records`, `read_records_with_health`, `read_records_with_health_bounded`, `read_all_records_with_health`, `read_retry_successors`.

Those resolve `paths::observation_db()` = `homeboy_data()/homeboy.sqlite`, which is exactly `store.observation_db_path()` for an env-rooted store — so **ambient behavior is unchanged**.

## The proof's negative half is the actual assertion

`lifecycle_read_siblings_answer_from_the_injected_store_and_not_a_second_root`, beside the existing `cook_progress_recorder_writes_to_the_injected_store_not_an_ambient_root`.

Two `HermeticTestContext` roots, `path_roots()` only, **no `with_isolated_home`, no `HomeGuard`, no `set_var`** — it asserts up front the two data dirs differ. State is seeded through store methods only, then all 23 siblings run against the seeded store.

Then the same ids are re-read through a store on the **second** root:

- `run_record_exists*` / `cook_index_exists` → `false`
- 12 read siblings → `is_err()`
- `resolve_run_id` echoes the **unresolved** cook id back instead of the attempt id
- `reconcile_scope_run_ids` returns `vec![cook_id]`, not the attempt
- `run_id_for_aggregate_path` returns `None` for a path that **does exist on disk** in the other root
- all three registry snapshots empty

**A sibling that ignored its store parameter would pass every positive assertion and fail these.**

## Dead code

`lib.rs:36-41` carries `#[allow(dead_code, unused_imports, …)]` on `pub mod agent_task_lifecycle`, and `lifecycle_store.rs` is pulled in underneath it via `#[path]`. Verified with a standalone `rustc` repro that an outer allow on a `pub mod` **does** suppress the lint inside a nested `#[path]` submodule even under `#![deny(warnings)]` — so the `-Dwarnings` gate is safe. The new `pub` siblings are re-exported through `pub use lifecycle_ops::*` anyway, so they are externally reachable.

## Compatibility

Purely additive plus delegation. No existing signature or behavior changes. **No caller is migrated** — adding the surface is the scope; moving callers onto it is a later slice.

## AI assistance

- **AI assistance:** Yes
- **Tool(s):** OpenCode general coding subagent, outside the Homeboy control plane
- **Used for:** Enumerating the read set, implementation, test authoring, and standalone `rustc` verification. Review by hand; compilation deferred to CI.

Refs #7505

### DIFF
--- a/crates/homeboy-agents/src/agent_task_batch.rs
+++ b/crates/homeboy-agents/src/agent_task_batch.rs
@@ -453,8 +453,16 @@ pub fn expire_stalled_fanout_admission(batch_id: &str) -> Result<bool> {
 /// actionable state. The admission deadline is independent of the heartbeat:
 /// a live process that is stuck before it can create a child must not renew its
 /// way into an indefinite `admitting` state.
+///
+/// The "did a child ever start" probe follows the injected lifecycle root for
+/// the same reason the batch roster follows the injected batch root: this is one
+/// decision about one wave. Read ambiently, another home's copy of the same run
+/// id could report a started child and hold a genuinely stranded coordinator in
+/// `admitting` forever, or report no child and terminalize a wave that is
+/// running somewhere else (#7505, #12619).
 fn expire_stalled_fanout_admission_in_store(
     store: &AgentTaskBatchStore,
+    lifecycle_store: &agent_task_lifecycle::AgentTaskLifecycleStore,
     batch_id: &str,
 ) -> Result<bool> {
     store.with_batch_lock(batch_id, || {
@@ -472,7 +480,11 @@ fn expire_stalled_fanout_admission_in_store(
                 || batch.metadata["child_finalizations"]
                     .get(&child.run_id)
                     .is_some()
-                || agent_task_lifecycle::run_record_exists(&child.run_id).unwrap_or(true)
+                // This used to call agent_task_lifecycle::run_record_exists(&child.run_id),
+                // which probed whatever home the environment pointed at while
+                // the batch roster came from the injected one (#7505, #12619).
+                || agent_task_lifecycle::run_record_exists_in_store(lifecycle_store, &child.run_id)
+                    .unwrap_or(true)
         });
         let expired = batch.state == AgentTaskBatchState::Admitting
             && coordinator["stage"].as_str() == Some("admitting")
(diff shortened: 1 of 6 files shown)


chubes4:
## Rebase note: #12619's scanner caught two pre-existing hazards

Rebasing onto #12621 turned up two genuine findings from `rooted_siblings_never_reach_ambient_store_state`:

```
agent_task_batch.rs:456  expire_stalled_fanout_admission_in_store
  reaches ambient `run_record_exists(`
agent_task_batch.rs:957  artifacts_in_store
  reaches ambient `artifacts(`
```

Both take an `AgentTaskBatchStore` and reach the **lifecycle** ambient half, splitting a batch operation across two homes: batch state follows the injected root while the run-record existence check and the per-child artifact read follow whichever home the process sits in.

**Neither is introduced here.** They have had this shape since the batch store was introduced in #12560. This PR only makes them *visible*, by rooting the lifecycle counterparts `run_record_exists` and `artifacts` — which is what turns them into recognised ambient wrappers for the scanner.

Closing them means threading an `AgentTaskLifecycleStore` alongside the batch store through both call chains. That is a batch-side slice with its own equivalence argument, not a rider on the read surface that merely exposed them — so they are added to `KNOWN_AMBIENT_REACHING_ROOTED_FUNCTIONS` with that reasoning. The list is self-pruning, so the rows cannot outlive the debt.

Also fixed in this rebase: a **silently duplicated** `AgentTaskLifecycleStore::read_retry_successors` method. #12621 added it as `pub(crate)` and this branch added it as `pub`; git merged both with no conflict markers, which would not have compiled. Main's form kept. That is the third time this file has produced a marker-less duplicate across a rebase.

Scanner now passes 7/7.

PR #12707, opened 2026-08-19
Root the deferred-workload tests and delete their ambient wrappers

chubes4:
## Summary

Deleting *unused* wrappers (#12705) only reached the ones no caller wanted. The wrappers that remain are held in place by **tests** — so the way to delete them is to root the tests that need them. That's the same work #7505 wants anyway, approached from the end that terminates.

| | before | after |
|---|---|---|
| `with_isolated_home` in this module | **12** | **0** |
| `HermeticTestContext` | 0 | 12 |
| ambient wrappers | 8 | **deleted** |

## The substitution is exact, not approximate

This is the part worth checking, so here it is explicitly:

```
with_isolated_home  →  sets XDG_CONFIG_HOME = <root>/.config
                    →  ambient_config_root() = paths::homeboy() = <root>/.config/homeboy

HermeticTestContext →  config_dir() = home().join(".config/homeboy")
                    →  home() returns root()
                    →  <root>/.config/homeboy
```

**Identical.** The tests read the same directory they always did — they just no longer mutate the process to say so.

Adds `claim_next_in_roots` and `claim_next_at_in_roots`, which two tests needed for entry points that had no rooted form. Neither calls `ambient_config_root()`.

## Two deletions I got wrong on the way in

I handed out a keep-list built from a caller census. It said `update` had **37** production callers and `store_path` had **2**. Both are in fact `#[cfg(test)]` **private** fns in this file, with doc comments reading *"Retained for the hermetic tests."*

```rust
/// Ambient mutation. Test-only for the same reason as [`store_path`].
#[cfg(test)]
fn update<T>(mutate: impl FnOnce(&mut Vec<DeferredWorkload>) -> Result<T>) -> Result<T> {
```

A bare-name frequency count across a workspace this size **measures the name, not the function.** Every unrelated `update(` in every crate scored as a caller.

<callout accent="#f59e0b">
It fails toward *keeping dead code* rather than *deleting live code* — the right direction, but still wrong. Left in place, both would have become unreachable `#[cfg(test)]` fns and failed CI under `-D warnings` as `dead_code`.
</callout>

## One test still reads the environment, deliberately

`worker_lock_contenders_reach_readiness_once_across_processes` re-execs itself. **An in-process root cannot cross an exec boundary**, so the child derives its root from the `HOME` the parent injects:

```rust
let config_root = PathBuf::from(std::env::var("HOME").expect("test home"))
    .join(".config").join("homeboy");
```

This is the one place in the file that touches the environment, and it's the *real env-var wiring* #7505's desired outcome asks be covered by an integration test rather than eliminated. Flag it in review if you'd rather it used a dedicated variable than `HOME`.

## Verification

`cargo check --workspace --tests` is clean, zero warnings.

**It does not run these tests.** The assertions here are about filesystem layout — precisely the class that compiles green and fails at runtime. This wants a real Test gate before it merges; please don't admin-merge it.

Refs #7505

### DIFF
--- a/crates/homeboy-cli/src/deferred_workload.rs
+++ b/crates/homeboy-cli/src/deferred_workload.rs
@@ -252,10 +252,27 @@ pub fn claim_next(runner_id: &str, owner: &str) -> Result<Option<DeferredWorkloa
     claim_next_at(runner_id, owner, now_ms())
 }
 
+pub fn claim_next_in_roots(
+    config_root: &Path,
+    runner_id: &str,
+    owner: &str,
+) -> Result<Option<DeferredWorkload>> {
+    claim_next_at_in_roots(config_root, runner_id, owner, now_ms())
+}
+
 /// Claim the next record using the supplied clock. The worker uses this seam to
 /// make lease recovery deterministic without changing the durable protocol.
 pub fn claim_next_at(runner_id: &str, owner: &str, now: u64) -> Result<Option<DeferredWorkload>> {
     claim_next_matching_at(runner_id, owner, now, |_| true)
+}
+
+pub fn claim_next_at_in_roots(
+    config_root: &Path,
+    runner_id: &str,
+    owner: &str,
+    now: u64,
+) -> Result<Option<DeferredWorkload>> {
+    claim_next_matching_at_in_roots(config_root, runner_id, owner, now, |_| true)
 }
 
 /// Claim the next deferred workload accepted by the selected runner. Records
@@ -369,10 +386,6 @@ pub fn defer_claim_in_roots(config_root: &Path, id: &str, owner: &str) -> Result
 /// Reconciliation terminates a worker that can no longer prove ownership, so
 /// its claims must not sit out the full lease before another worker may take
 /// them. Returns the ids that were released.
-pub fn release_claims_for_owner(owner: &str) -> Result<Vec<String>> {
-    release_claims_for_owner_in_roots(&ambient_config_root()?, owner)
-}
-
 pub fn release_claims_for_owner_in_roots(config_root: &Path, owner: &str) -> Result<Vec<String>> {
     update_in_roots(config_root, |records| {
         let now = now_ms();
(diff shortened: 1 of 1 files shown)


chubes4:
## CI result: 20 failures, none of them from this PR

All four Test shards are red. **Zero failures are in `deferred_workload`** — the only file this PR touches.

| group | count | status |
|---|---|---|
| `homeboy-lab-runner workspace::tests::prune::*` | 11 | known-flaky, failing on unrelated PRs before this branch existed |
| scattered (`status_and_recovery`, `promotion`, `bench`, `notification_route_resolver`, `deploy`, `engine-primitives`, `connection`) | 9 | pre-existing |
| **in `deferred_workload`** | **0** | — |

Every one of the 12 rooted tests passed, including the ones whose whole job is filesystem layout:

```
deferred_workload::tests::next_claim_heartbeats_and_publishes_durable_worker_status   ok
deferred_workload::tests::corrupt_store_fails_closed_without_resetting_records        ok
deferred_workload::tests::deferred_workload_is_idempotent_and_survives_restart_before_claim  ok
deferred_workload::tests::only_the_proven_singleton_owner_survives_reconciliation     ok
deferred_workload::tests::matching_claim_skips_a_live_claimed_head_for_a_later_deferred_record  ok
```

That is the evidence I actually wanted, because `cargo check` cannot see any of it.

## One gap, stated plainly

`worker_lock_contenders_reach_readiness_once_across_processes` — the re-exec test, and the only place I deviated — **was enumerated in the shard plan and never executed.** It appears in the run exactly twice, both in the test listing, never as a start/ok/fail event.

So the subprocess change is **unverified by this run.** What I can offer instead is the resolution chain, read rather than executed:

```
paths::homeboy()      = resolved_home_root()/.config/homeboy
resolved_home_root()  = $HOME, when no in-process override is registered
                        (a freshly exec'd child has none by construction)
with_isolated_home    sets HOME = ctx.root()
config_dir()          = root/.config/homeboy
```

Three paths, one directory. The child deriving its root from `HOME` lands where `ambient_config_root()` landed. `XDG_CONFIG_HOME` is set by `with_isolated_home` but `homeboy()` never reads it on Unix — it reads `HOME` — so the substitution holds for the parent tests too.

<callout accent="#f59e0b">
I would rather this test had run. If there is a filter excluding self-respawning tests from sharded runs, that is worth knowing independently of this PR — it means a whole class of cross-process tests is silently not gating anything.
</callout>

## Correction

My first read of these shards said "zero tests failed." That was wrong — I matched a `"failed":0` line from one suite and generalized it. The shard totals plainly said `failed=5`, `failed=6`, `failed=9`. The conclusion happens to survive the correction, but the reasoning that produced it was bad and I'd rather flag it than let it stand.

Issue #11380, opened 2026-08-03
Lab offload reaps run workspace while remote job state is explicitly uncertain


chubes4:
## Problem

A pinned Lab benchmark failed with `runner.controller_disconnected`. The failure envelope explicitly said:

```text
the remote job may still be in flight
runner `homeboy-lab` has daemon job `6ed32d1d-91e5-406f-b238-9000e4c98b0f` still in flight
```

Before returning that uncertainty, the same command printed:

```text
Lab offload: reaped run-scoped workspace `/home/chubes/Developer/_lab_workspaces/static-site-importer-...` on runner `homeboy-lab` (policy=delete-on-terminal, outcome=failure, lifecycle_owner=runner.workspace).
```

The controller classified transport loss as a terminal workspace-cleanup outcome while simultaneously preserving the remote daemon job as potentially live. If the job survived, its source checkout disappeared underneath it. If it did not survive, Homeboy still lacked authoritative terminal evidence when cleanup ran.

## Impact

- A recoverable tunnel interruption can destroy the inputs of an expensive live Lab workload.
- Durable job recovery becomes meaningless when the associated workspace has already been removed.
- The operator is pushed toward `--preserve-workspace-on-failure` for every run, even though transport uncertainty is not a workload failure.
- Product evidence cannot distinguish genuine test failure from controller-induced workspace removal.

## Acceptance criteria

- `controller_disconnected` and other uncertain remote states never trigger delete-on-terminal workspace cleanup.
- Cleanup requires authoritative terminal daemon-job state and completed artifact/result handoff.
- Uncertain workspaces transition to a typed retained/reconciliation-needed lifecycle with bounded TTL.
- Reconnect/reconcile either resumes the known job or terminalizes it before cleanup.
- Evidence records workspace identity, owning generation/job, cleanup disposition, and the observation that authorized deletion.
- End-to-end coverage drops polling during a live workload and proves its workspace remains available through recovery.

## AI assistance

OpenAI gpt-5.6-sol via OpenCode ran the matrix, correlated the cleanup line with the contradictory in-flight recovery envelope, searched the tracker, and drafted this issue. Chris Huber reviewed and owns the report.

PR #12711, opened 2026-08-19
Serialize nextest shard replay until the shared stores take roots

chubes4:
## Summary

Main has ~21 tests failing on every run. This fixes the largest cause, which was **mine** — #12626 set `rust_nextest_shard_threads: 0` two days ago, and the reasoning I used to justify it was wrong.

```
rust_nextest_shard_threads:  0  →  1
```

## The argument #12626 made, and where it breaks

> nextest runs every test in its own process, so process-global state is already isolated and serial replay is only the conservative default.

The first half is true. **The conclusion does not follow.**

Process-per-test isolates process-global *memory and environment* — the `HOME`/`XDG_*` mutation `HomeGuard` performs, which is the race #7505 exists to remove.

It does nothing for machine-global *filesystem* state. The controller-runtime admission store and the rig registry are **directories on disk beneath the real `$HOME`**. Concurrent test processes share those exactly as thoroughly as concurrent threads would.

<callout accent="#3b82f6">
A fresh process is not a fresh disk. That is the whole bug.
</callout>

## The evidence

Run [32294476539](https://github.com/Extra-Chill/homeboy/actions/runs/32294476539). `status_preserves_existing_terminal_runtime_evidence` asserts a record is unchanged across a `status` call. Its two sides differed in **exactly one field**:

```
before  metadata.controller_admission.owner = { pid: 9913, advisory_lock: true,
                                                request_id: "run_store-contract", .. }
after   metadata.controller_admission.owner = Null
```

**`run_store-contract` is the run id of a different test in the same file.** One test observed another's admission lease through the shared on-disk store, mid-flight — with process isolation fully in effect and completely irrelevant.

The ten `workspace::tests::prune::*` failures are the same shape through the rig registry: a peer's `create` leaves registry state the assertions were never written against. `prune.rs:62` expects 1 orphan candidate and finds 0.

## What has to be true before this returns to `0`

**Not "the flakes stopped."** The stores have to stop being machine-global and take explicit roots like everything else under #7505.

Until then a parallel value isn't faster, it's red — and a gate that goes red for reasons unrelated to the change under review is a gate nobody reads. That's the state main has been in, and it's why ~40 PRs merged this week past a signal that had stopped meaning anything.

## Also fixed: the drift that hid this twice

The value has moved three times and only one move announced itself:

| commit | value | announced? |
|---|---|---|
| `716370152` restore parallel hermetic shards | `0` | yes — updated the docs |
| `fcc6a5029` ci: bound archived test replay | `1` | **no** — one-line message, unrelated subject, docs left saying `0` |
| `522c7cda2` (#12626) | `0` | yes — read the silence as accident, flipped back |

`docs/internals/test-tiers.md` documented `0` throughout the window it was `1`. That doc is now corrected and the guard test pins `1` with the evidence attached, so the next person to change it has to argue with something.

## Not fixed here

`cancel_run_signals_live_running_record` fails for a different reason — `submit_plan_with_runtime_admission(..., |_| Ok(json!({})))` writes an empty `controller_runtime` pin and `validate_pin` rejects it with *"controller runtime pin has no build identity"*. That is the same defect #12700 reverted in `terminal_and_reconcile.rs`; the identical pattern survives in `status_and_recovery.rs` (28 sites). Separate PR — it's a test-content fix, not a config one.

Refs #7505

### DIFF
--- a/docs/internals/test-tiers.md
+++ b/docs/internals/test-tiers.md
@@ -58,10 +58,23 @@ is no thread-local escape hatch. (Rust made `std::env::set_var` `unsafe` in the
 One thread per test binary closes that window on the Cargo fallback. Nextest
 closes it more directly by running each test in its own process.
 
-The Cargo cap is a fallback workaround, not the primary strategy. Sharded
-nextest replay sets `rust_nextest_shard_threads = 0`, selecting nextest's
-process-per-test `num-cpus` parallelism without exposing Cargo's shared-process
-environment mutation.
+The Cargo cap is a fallback workaround for a race that is specific to sharing
+one process across threads. Sharded nextest replay does not have that race — but
+it sets `rust_nextest_shard_threads = 1` anyway, for a second and independent
+reason.
+
+Process-per-test isolates environment. It does not isolate the filesystem. The
+controller-runtime admission store and the rig registry are machine-global
+directories beneath the real `$HOME`, so concurrent test *processes* share them
+just as thoroughly as concurrent threads would. Run `32294476539` caught one
+test reading another's admission lease — the two sides of an "unchanged record"
+assertion differed only in `controller_admission.owner`, which carried a peer
+test's `request_id` — and ten `workspace::tests::prune::*` failures traced to
+shared rig-registry writes.
+
+The two settings are therefore both `1` and clear on different conditions. The
+Cargo cap lifts when `HomeGuard` stops mutating process-global environment; the
+nextest cap lifts when those stores take explicit roots. Both are #7505.
 
 Unsharded nextest runs were previously unmeasurable — they emit no
 `test result:` lines, so the cargo-test adapter matched nothing and produced no
--- a/homeboy.json
+++ b/homeboy.json
@@ -1826,7 +1826,7 @@
       "settings": {
         "rust_test_runner": "nextest",
         "rust_cargo_test_threads": 1,
-        "rust_nextest_shard_threads": 0
+        "rust_nextest_shard_threads": 1
       }
     }
   },
--- a/tests/nextest_shard_parallelism_test.rs
+++ b/tests/nextest_shard_parallelism_test.rs
@@ -1,54 +1,83 @@
-//! Pins nextest shard replay to process-per-test parallelism.
+//! Pins nextest shard replay to serial execution.
 //!
 //! `extensions.rust.settings.rust_nextest_shard_threads` is the `--test-threads`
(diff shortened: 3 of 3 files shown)


chubes4:
## Correction: this fixes ~2 of the failures, not 11

I claimed the parallelism change was the cause of red main. **The serial run disproves most of that**, so here is the actual comparison.

| | parallel (`0`) | serial (`1`) |
|---|---|---|
| total failures | 21 | **19** |
| `status_preserves_existing_terminal_runtime_evidence` | fail | **fixed** |
| `workspace::tests::prune::*` | 10 fail | **10 still fail** |

## What survives, and what it means

**The admission race was real and is fixed.** `status_preserves_existing_terminal_runtime_evidence` failed with a peer test's `request_id: "run_store-contract"` in its record, and under serial execution it passes. Process-per-test genuinely does not isolate the on-disk controller-runtime store, and that reasoning in #12626 was genuinely wrong.

**The prune cluster is not a concurrency bug.** All ten still fail serially, so my attribution was wrong. They use `with_isolated_home` and were failing before #12626 landed — which also means **I did not cause red main.** I added one or two failures to a suite that already had nineteen.

That is a materially different claim from the one in the original description and I would rather correct it than let it stand.

## So is this still worth landing

Honest answer: it buys one test and costs shard wall-time, and it does not make main green. The argument for it is that the race it removes is real, silent, and gets worse as more tests touch controller admission — a test reading a peer's lease produces a failure that points at the wrong code, which is the most expensive kind.

The argument against is that serializing the whole suite to fix two tests is a blunt instrument, and the sharper version is a nextest test-group that serializes only the tests touching controller admission and the rig registry.

I don't have a strong opinion between those and I'd rather you pick. What I am confident about is that the `0` value was justified by reasoning that does not hold.

## The prune cluster is unexplained and is the bigger problem

Ten tests, consistent, not concurrency, not recently modified (`prune.rs` last touched in #11380). `prune.rs:62` asserts one orphan candidate and gets zero:

```rust
fs::remove_dir_all(&source).expect("remove source");
let (output, exit_code) = prune_workspaces("lab-local-prune-preview", ...);
assert_eq!(output.candidates.len(), 1);   // ← left: 0, right: 1
```

The workspace is synced, the source is deleted, and the prune scan then reports nothing to reclaim. That is either a real regression in orphan detection or an environmental dependency of the scan that CI does not satisfy. Either way it is the majority of red main and it needs its own investigation, not a config flag.

Issue #12721, opened 2026-08-20
Isolate live-cancellation test from malformed controller runtime fixture


chubes4:
## Problem

`cancel_run_signals_live_running_record` submits a run with `json!({})` as controller-runtime admission evidence, then calls `mark_running_in_store`. Runtime mutation validation correctly rejects that malformed pin because it has no `/originating/build_identity`.

This reproduces on current `main` and appears across multiple independent PR shard runs that do not modify this test or controller-runtime validation.

## Reproduction

```sh
cargo test -p homeboy-agents cancel_run_signals_live_running_record -- --nocapture
```

Observed failure:

```text
Invalid argument controller_runtime: controller runtime pin has no build identity
```

## Expected

The cancellation test should construct only the lifecycle state its cancellation contract requires, without persisting malformed runtime-admission metadata. The real controller-runtime mutation contract remains fail-closed.

## Evidence

- https://github.com/Extra-Chill/homeboy/actions/runs/32341033189
- https://github.com/Extra-Chill/homeboy/actions/runs/32341033757
- https://github.com/Extra-Chill/homeboy/actions/runs/32341035203
- https://github.com/Extra-Chill/homeboy/actions/runs/32341034028

## AI assistance

GPT-5.6-sol via OpenCode reproduced and isolated the fixture failure and drafted this issue. Chris Huber reviewed the investigation and remains responsible for the proposed direction.

chubes4:
## Root cause confirmed — fixture, not environment, and deterministic

Traced the full call chain. The failure is **not** `validate_for_mutation`. It is legacy-pin **migration**, which runs one line earlier.

`crates/homeboy-agents/src/agent_task_lifecycle/lifecycle_ops.rs:2229-2239`:

```rust
pub(crate) fn mark_running_in_store(...) -> Result<AgentTaskRunRecord> {
    let run_id = sanitize_run_id(run_id);
    let mut record = lifecycle_store.read_record(&run_id)?;
    migrate_record_controller_runtime_in_store(lifecycle_store, &mut record)?;   // 2235  <-- fails here
    homeboy_core::controller_runtime::validate_for_mutation(                     // 2236
        &record.metadata,
        &homeboy_core::build_identity::current().display,
    )?;
```

Migration is skipped only when the key is **absent**. The test persists `{}`, which is present:

```rust
let Some(runtime) = record.metadata.get(CONTROLLER_RUNTIME_METADATA_KEY) else {
[...]

Issue #12715, opened 2026-08-19
Ten workspace prune tests fail on every CI run and are the bulk of red main


chubes4:
## Symptom

Ten `homeboy-lab-runner::workspace::tests::prune::*` tests fail on every CI run, on every branch. They are the majority of red main.

```
prune_workspaces_previews_orphans_without_deleting_by_default
prune_workspaces_apply_passes_drain_until_empty
prune_workspaces_apply_removes_only_metadata_backed_orphans
prune_workspaces_advances_through_thousands_of_mixed_entries
prune_workspaces_preview_reports_synthetic_odd_path_without_deleting
prune_workspaces_reaps_stale_materialized_workspace_with_live_source
prune_workspaces_reports_remaining_bytes_and_drain_command_when_limited
prune_convergence_resumes_durable_receipts_across_more_than_twenty_pages
prune_preserves_process_owned_workspace_in_preview_and_apply
retained_terminal_receipt_reclassifies_and_releases_compacted_workspace
unavailable_job_authority_fails_closed_despite_inactive_process_evidence
```

The shape is consistent — the scan finds nothing to reclaim:

```rust
// prune.rs:62
fs::remove_dir_all(&source).expect("remove source");
let (output, exit_code) = prune_workspaces("lab-local-prune-preview", ...);
assert_eq!(output.candidates.len(), 1);   // left: 0, right: 1
```

## What has been ruled out

**Not concurrency.** Run [32302557365](https://github.com/Extra-Chill/homeboy/actions/runs/32302557365) ran with `rust_nextest_shard_threads: 1` (fully serial). All ten still fail. The parallel run immediately before it failed the same ten. Serialization fixed a different test (`status_preserves_existing_terminal_runtime_evidence`, a real controller-admission race) and changed nothing here.

**Not a recent test edit.** `workspace/tests/prune.rs` was last touched in #11380.

**Not env leakage.** These tests use `with_isolated_home`, and each builds its own `tempfile::tempdir()` runner root.

## Where to look next

`prune_workspaces` scans via a generated shell one-liner, `prune_scan_command` (`workspace/sync/mod.rs:3790`). A directory is dropped from the candidate list at several `continue` points, all silent:

```sh
meta="$dir/$meta_rel"; [ -f "$meta" ] || continue
mtime=$(stat -c %Y "$dir" ... || echo 0); age=$((now-mtime)); [ "$age" -ge "$min_age" ] || continue
[...]

PR #12732, opened 2026-08-20
Root the terminal-notification claim protocol

chubes4:
A slice of #7505. Independent of #12725 — touches no file that PR rewrites.

## The shape

`cook_terminal`, `batch_terminal`, and `run_reconciled` each run one exactly-once protocol:

```
claim  ->  record outcome  ->  confirm   (delivered / queued)
                          \->  release   (dropped / rejected)
```

Every step resolved its own root from the process environment. Four independent resolutions for one transaction.

## Why this is a correctness fix, not a count change

Neither failure mode fails. Both are silent:

| step lands in wrong home | what happens |
|---|---|
| **confirm** | consumes an eligibility nobody holds; the real claim stays merely *held* until its 5-minute lease expires — **the duplicate terminal notification the claim exists to prevent** |
| **release** | `remove_file` treats `NotFound` as success, so it deletes nothing and returns `Ok(())`; the real claim keeps blocking the later observer the release was meant to unblock |

The codebase already made this argument for the claim step, in `claim_cook_notification_in_store`:

> `cook_index_path` as a free function is `default_store()?`, so the store's own method is used here: the claim marker is a bare filesystem create with no record read in front of it, and an ambient reach would land the marker in another home without failing.

The rooted siblings for claim, confirm, and record were written and merged. **The callers were never converted.** This PR converts them.

## What changed

Each entry point resolves one lifecycle store and passes it to the rooted sibling of every step. An unresolvable store is treated exactly as a lost claim — the rule these functions already applied to a failed claim.

The **release step had no rooted form at all**; it reached `store::`, which resolves `default_store()` for itself. Added:
- `lifecycle_store::release_cook_notification_claim_in_store`
- `AgentTaskLifecycleStore::release_cook_notification_claim`
- `agent_task_lifecycle::release_cook_terminal_notification_claim_in_store`

Its path construction is copied verbatim from `claim_cook_notification_in_store`, so the release removes exactly the marker the claim creates.

Deleted, having lost their last callers: `confirm_cook_terminal_notification`, `record_cook_terminal_notification_outcome`. `claim_cook_terminal_notification` and `release_cook_terminal_notification_claim` keep their test callers and stay.

## Count delta — stated plainly, because it goes the "wrong" way

```
AgentTaskLifecycleStore::from_current_environment(   142 -> 143   (+1)
CookRecipeStore::from_current_data_root(              21 -> 21
PathRoots::from_environment(                          40 -> 40
```

**+1.** Two wrapper definitions deleted, three entry-point resolutions added.

The textual constructor count is the wrong measure for this change. What actually changed is resolutions performed **per delivery**:

```
cook_terminal     3 -> 1
batch_terminal    3 -> 1
run_reconciled    2 -> 1
```

11 ambient wrapper call sites replaced by 3 explicit resolutions. The three functions become the "env-resolving entry point that takes no store at all and resolves everything it needs in one place" that `tests/agent_task_store_injection_test.rs` explicitly documents as the legal shape.

I'd rather show a +1 with the real number beside it than convert a caller vacuously to make a counter go down.

## Verification

- `nice -n 10 cargo check --workspace --tests -j2` — **clean, 0 warnings, 0 errors**, 3m23s
- `cargo fmt --check` — clean
- `tests/agent_task_store_injection_test.rs` standalone — **7/7 pass**, both non-vacuity guards and both `KNOWN_*` staleness checks intact
- Dead-wrapper rescan after the edit: **0 dead ambient wrappers, no cascade**
- Verified `store.cook_index_path()` and the free `cook_index_path()` are the same method (the free one is literally `default_store()?.cook_index_path(...)`), so the release path is unchanged when the store is the ambient one

### The tests that actually matter here

Two existing tests in `agent_task_notify_tests.rs` exercise both branches and are **not** vacuous — each asserts a state transition observable from the ambiently-resolved home, so a mis-rooted confirm or release would fail them:

- `terminal_delivery_records_success_and_confirms_the_once_marker` — after `cook_terminal`, asserts `claim_cook_terminal_notification(...)` returns **false** (eligibility consumed ⇒ confirm landed where the ambient reader looks)
- `non_delivery_is_recorded_and_leaves_terminal_delivery_eligible` — after a failed delivery, asserts `claim_...` returns **true** (⇒ release actually removed the marker in that home)

Both run under `with_isolated_home`, so env resolution is consistent between the test and the entry point.

I could not run them locally — a `cargo test` build is ~28G on this box — so CI is the check. `cargo check --tests` compiled them.


chubes4:
## Red CI here was a stale base, not this change

The previous run (32396167682) failed all four test shards with **19 distinct failures**. None of them are in the three files this PR touches (`agent_task_notify.rs`, `lifecycle_ops.rs`, `lifecycle_store.rs`):

```
10x homeboy-lab-runner workspace::tests::prune::*          (#12715)
 2x homeboy::cli_binary output_errors::*
 1x homeboy-agents  status_and_recovery::cancel_run_signals_live_running_record
 1x homeboy-lab-runner connection_daemon::durability_check_*
 1x homeboy-lab-runner evidence::mirroring_lab_job_*
 1x homeboy::homeboy_installer_test installer_rejects_unsafe_*
 1x homeboy-release  release::planner::dry_run_plans_reachable_*
 1x homeboy-cli      cli_surface::reference_docs_tests::*
```

The decisive comparison is #12734, merged from the same day against a newer base:

| run | test shards |
|---|---|
| this PR @ `822f9311e` (4 commits behind main) | **all 4 FAILED** |
| #12734 @ `4605d9c5` (current main) | **all 4 PASSED** |

Its only failing job was `Required Gates Executed`. So the suite is green on current main, and this branch was testing a base where it was not.

Merged `origin/main` in (clean auto-merge, one conflict-free touch in `lifecycle_ops.rs` — #12734 edited ~line 2419, this PR edits ~1836-1890). Re-verified on the merged result:

- `nice -n 10 cargo check --workspace --tests -j2` — clean, **0 warnings, 0 errors**
- `cargo fmt --check` — clean

CI is re-running against the current base.

chubes4:
## Correction, and the actual evidence

My earlier comment blamed a stale base. **That was wrong** — I re-ran CI against current main and it failed again, so the drift was not the cause. Recording the correction rather than quietly deleting it.

The rerun (32406371953) has 21 distinct failures. **None of them are notification tests**, and none are in the three files this PR touches:

```
10x  lab-runner   workspace::tests::prune::*                      -> #12715
 1x  agents       cancel_run_signals_live_running_record          -> #12721
 2x  homeboy      cli_binary output_errors::*
 1x  homeboy      homeboy_installer_test  (tar: '--same-order' cannot be used with '-c')
 1x  lab-runner   connection_daemon::durability_check_*
 1x  lab-runner   evidence::mirroring_lab_job_*
 1x  release      planner (refuses branch 'master' vs default 'main')
 1x  cli          cli_surface::reference_docs_tests
 2x  agents       stalled_provider_is_killed_* / concurrent_first_cooks_*  (absent from the first run — timing)
```

The authoritative datapoint is the repo's own issue title:

> **#12715 — "Ten workspace prune tests fail on every CI run and are the bulk of red main"**

That accounts for 10 of the 21 directly, #12721 for another, and the installer and release-planner failures are environmental (`tar` version, branch naming) rather than code.

Two failures appeared only in the second run (`stalled_provider_is_killed_and_rotates_to_configured_fallback`, `concurrent_first_cooks_elect_one_recipe_creator_without_replacing_its_plan`). Both are concurrency/timeout tests and both passed in the first run on the same code, so they are timing, not this diff.

What this PR changes is the terminal-notification claim protocol in `agent_task_notify.rs`. The two tests that actually gate it — `terminal_delivery_records_success_and_confirms_the_once_marker` and `non_delivery_is_recorded_and_leaves_terminal_delivery_eligible` — pass in both runs.

PR #12736, opened 2026-08-20
Root the detached-Cook handoff protocol

chubes4:
A slice of #7505. Touches `local_detach.rs`, `agent_tasks.rs`, `lifecycle_ops.rs`.

## The protocol

`intercept_local_detached_cook` runs one parent → child → supervisor handoff and resolved a root **nine separate times** to do it:

```
1x  record_detached_cook_handoff_parent    open the fence
1x  record_detached_cook_handoff_child     bind the spawned child
1x  record_detached_cook_supervisor        project daemon ownership
6x  fail_detached_cook_handoff_parent      every compensating failure path
```

A parent opened in one installation and failed in another leaves the first one's fence **standing open forever** — the detached Cook reads as live while nothing owns it — and the compensation that was supposed to release it returns success having touched a different home.

All nine now share one store, resolved immediately before the first durable write so it fails in the same place `record_detached_cook_handoff_parent` used to.

## Why this was never converted — the part worth reading

This is structural, not an oversight. I hit it as a compile error and it explains the shape of the whole remaining problem.

`homeboy-cli` reaches lifecycle operations through `homeboy::agents::agent_tasks::lifecycle`. That is **a curated re-export list, not a glob**:

```rust
pub mod lifecycle {
    pub use super::super::agent_task_lifecycle::{
        aggregate_source, artifacts, cancel, cancel_run, claim_cook_operation,
        ... ~100 names ...
    };
}
```

**Not one of them is an `_in_store` sibling.** Every name is the ambient form. A CLI caller was never choosing ambient — it was the only form the facade exposed.

Underneath sits a second layer of the same shape. Scanning `homeboy-agents` for pairs where the rooted sibling is declared *less visible* than its ambient twin:

```
94 pairs
  pub  submit_plan                    ->  pub(crate)  submit_plan_in_store
  pub  mark_running                   ->  pub(crate)  mark_running_in_store
  pub  record_pre_execution_failure   ->  pub(crate)  record_pre_execution_failure_in_store
  pub  record_promotion               ->  pub(crate)  record_promotion_in_store
  ...
```

So even a glob re-export would not have reached them. **The rooted half of this convention is systematically less reachable than the ambient half**, and the CLI — where the process entry points live, and therefore exactly where roots should be resolved once and threaded down — is on the wrong side of both barriers.

## Three edits, not one

1. 4 `_in_store` siblings widened `pub(crate)` → `pub`, matching the visibility their ambient twins already had
2. Those 4 added to the `agent_tasks::lifecycle` facade
3. The 9 call sites converted

I widened **only the four this caller needs**. Opening all 94 is a far larger blast radius and most of their callers are in-crate, so widening on demand is the smaller correct step. But the asymmetry is the reason the CLI boundary has stayed ambient, and it will keep producing this same finding until it is named.

## Count delta

```
AgentTaskLifecycleStore::from_current_environment(   unchanged
```

One explicit resolution added in the CLI, nine implicit ones removed from the call graph. No wrapper lost its last caller — all four keep other callers, so nothing became deletable here.

## Verification

- `cargo fmt --check` — clean
- Post-edit scan of the function: **0 remaining ambient `agent_task_lifecycle::` calls**

The first gate run failed with 10 `E0425`s — that is how the facade was found. Recorded because "widen to `pub`" looked sufficient and was not.

Not run locally: the full suite (~28G build). `cargo check --tests` compiled it; CI is the check.


chubes4:
## CI red is the documented red main, not this change

20 distinct failures, and they are a **strict subset** of the 21 seen on #12732 — a PR that touches an entirely disjoint set of files (`agent_task_notify.rs` / `lifecycle_ops.rs` / `lifecycle_store.rs` vs this one's `local_detach.rs` / `agent_tasks.rs`). `comm -23` between the two failure sets is empty: this PR introduces nothing that PR did not already show.

```
10x  lab-runner   workspace::tests::prune::*              -> #12715
 1x  agents       cancel_run_signals_live_running_record  -> #12721
 2x  homeboy      cli_binary output_errors::*
 1x  homeboy      homeboy_installer_test   (tar: '--same-order' cannot be used with '-c')
 1x  lab-runner   connection_daemon::durability_check_*
 1x  lab-runner   evidence::mirroring_lab_job_*
 1x  release      planner  (refuses branch 'master' vs default 'main')
 1x  cli          cli_surface::reference_docs_tests
 1x  agents       concurrent_first_cooks_*  (timing; passed on an earlier run of identical code)
```

Per the repo's own issue:

> **#12715 — "Ten workspace prune tests fail on every CI run and are the bulk of red main"**

Worth singling out one, because it is the only failure that even looks related to a CLI change: `cli_surface::reference_docs_tests::generated_command_docs_match_the_live_clap_tree`. It fails identically on #12732, which changes no CLI code at all. This PR alters a re-export list and a call site, not the clap tree.

No `local_detach` test fails, and nothing in the detached-handoff path regressed.

PR #12743, opened 2026-08-20
Root the agent-task status read

chubes4:
A slice of #7505. Follows the facade finding recorded in #12736.

## The read

`status_once` — the `homeboy agent-task status` path — resolved **three** lifecycle roots to answer one question:

```
exact_durable_local_read / durable_local_read   the record
load_plan                                       reject an incompatible budget version
```

Reading the record from one installation and the plan from another reports a run whose budget was never checked, and reports it as though it had been. All three now share one store, resolved once at the top of the read.

`runner_probe_plan` is deliberately left alone: it takes an already-read record and computes a projection from it. It resolves no root and has no rooted sibling, so there is nothing to inject. It is the only `agent_task_lifecycle::` call remaining in the function.

## Doing the facade work demand-driven

You greenlit the facade barrier. I did it **for the three names this conversion needs**, not in bulk, and I want to be explicit about that choice.

#12736 recorded the problem: `agent_tasks::lifecycle` is the curated re-export `homeboy-cli` reaches lifecycle operations through, it names roughly a hundred functions, and before this change **not one was an `_in_store` sibling**. A CLI caller could not root anything, because the rooted half was not exported.

The bulk fix — exporting all ~94 rooted siblings — is API surface with no consumer. This repo does not reward that, and an unused re-export cannot be reviewed against anything. Closing the asymmetry through the conversions that need it means every added name arrives with a caller that proves it works.

All three siblings here were already `pub`; only the facade entry was missing, so **no visibility widening was required**. Some future slices will need it; those can widen what they use.

## Count delta

```
AgentTaskLifecycleStore::from_current_environment(   144 -> 145   (+1)
CookRecipeStore::from_current_data_root(              16 -> 16
PathRoots::from_environment(                          40 -> 40
```

**+1.** One explicit resolution added in the CLI; three implicit ones removed from the call graph.

The counter rises because the resolution moved *out of three wrapper bodies shared with other callers* and *into one entry point that owns it*. Those wrappers keep their other callers, so nothing became deletable here — unlike #12742, where the same shape did free three wrappers one merge later. Which of the two a slice turns out to be is not knowable until its wrappers lose their last caller.

## Verification

- `cargo fmt` — clean
- Post-edit scan of `status_once`: only `runner_probe_plan` remains, which resolves no root

Not run locally: the full suite (~28G build). Expect the documented red-main set (#12715, #12721).


chubes4:
## Revised: this now collapses a pair instead of adding one

The original push rooted `status_once` and left `exact_durable_local_read` standing. That was the wrong stopping point — `status_once` was that wrapper's **only** caller, so rooting it made the ambient half unreferenced. It is now deleted, and the facade re-exports the rooted function in its place.

```
before revision:  25 insertions, 17 deletions   counter 144 -> 145
after revision:   35 insertions, 45 deletions   counter 144 -> 144
```

Net **-10 lines**, one fewer function in `lifecycle_ops.rs`, and the operation now exists once rather than as an ambient/rooted pair.

`durable_local_read` and `load_plan` keep their other callers and are unchanged.

Re-verified after the revision: `cargo check --workspace --tests -j2` clean (0 warnings, 0 errors), `cargo fmt` clean.

PR #12744, opened 2026-08-20
fix: three root causes behind red main (#12715, #12721, #12741)

### LINKED ISSUE #12721: Isolate live-cancellation test from malformed controller runtime fixture

### LINKED ISSUE #12741: Promotion provider tests fail with an opaque "IO error" when the provider exits without draining stdin
## Symptom

`homeboy-agents agent_task_promotion::tests::part_b::provider_response_validation_distinguishes_json_schema_and_required_field_errors` fails at full test scope on `main`:

```
panicked at crates/homeboy-agents/src/agent_task_promotion/tests/part_b.rs:2156:9:
IO error
```

The sibling `provider_failure_surfaces_bounded_stdout_and_stderr_evidence` in the same module failed in one of three observed full-scope runs.

Seen in runs `32291884920`, `32381612499`, `32395721630` (all on `main`).

## `"IO error"` is not an expect message

Grepping the literal `"IO error"` across `part_b.rs` returns **zero** matches. Line 2156 is:

```rust
assert!(error.message.contains(expected), "{}", error.message);
```

`"IO error"` is the canned `Error::message` for `ErrorCode::InternalIoError` — `crates/homeboy-error/src/lib.rs:1513`:

```rust
Self::new(ErrorCode::InternalIoError, "IO error", details)
```

So `run_provider_command` returned an **infrastructure** error instead of the expected validation error. The real OS string and operation context live in `details.error` / `details.context` — **which the assertion never prints.** The CI evidence is structurally insufficient to name the failing site.

## Leading hypothesis: EPIPE writing the request to a provider that already exited

`crates/homeboy-agents/src/agent_task_promotion/apply.rs:694`:

```rust
    .write_all(&request_json)
    .map_err(|error| {
        Error::internal_io(
            error.to_string(),
            Some("write agent-task promotion provider request".to_string()),
        )
    })?;
drop(process.stdin.take());
```

The sequence is spawn-with-piped-stdin → `write_all` → drop. The provider in this test is:

```rust
argv: vec!["sh", "-c", "printf '%s' \"$1\"", "sh", response]
```

It **never reads stdin** and exits almost immediately. Rust sets `SIGPIPE` to `SIG_IGN`, so writing to a pipe whose read end has closed returns `Err(BrokenPipe)` rather than killing the process — yielding exactly `internal_io(...)`. Whether the child wins the race is scheduling-dependent, hence intermittent.

**The discriminator that fits the observed failure set:** `configured_provider_timeout_is_bounded_and_retains_command_evidence` in the same module runs `sh -c "sleep 2"` — the child stays alive, the write always lands — and it is **not** among the failures. The two tests that *do* fail are precisely the two whose provider exits immediately:

| test | provider | rolls per run | observed |
| --- | --- | --- | --- |
| `provider_response_validation_...` | `printf '%s' "$1"` in a 5-case loop | 5 | fails every run |
| `provider_failure_surfaces_bounded_...` | `printf ...; exit 7` | 1 | failed 1 of 3 runs |

## Ruled out

- Neither test writes or execs a fixture binary — both exec `sh` from `PATH`. So `noexec /tmp`, missing stub executables and `ETXTBSY` are inapplicable. (A real `ETXTBSY` bug was fixed in `f6d81546c` on 2026-08-04, but that touched `part_a.rs` and `part_c.rs`, not these.)
- Neither test calls `with_isolated_home`, so `HOME`/temp-dir relocation is not involved.

**Not** ruled out: transient spawn failure (`ENOMEM`/`EMFILE`) under CI parallelism would land in the same `internal_io` bucket.

## Proposed fix

**(a) Product** — treat a provider that exits without draining its request as non-fatal. Its exit code and captured stdout remain the authoritative result:

```rust
match stdin.write_all(&request_json) {
    Ok(()) => {}
    // A provider may exit without draining its request. Its exit code and
    // captured output remain the authoritative result.
    Err(error) if error.kind() == std::io::ErrorKind::BrokenPipe => {}
    Err(error) => {
        return Err(Error::internal_io(
            error.to_string(),
            Some("write agent-task promotion provider request".to_string()),
        ))
    }
}
```

**(b) Evidence** — `part_b.rs:2156`, so the next occurrence names itself:

```rust
assert!(
    error.message.contains(expected),
    "{} {}",
    error.message,
    error.details
);
```

**(b) is worth applying regardless.** It is zero-risk and it is the only way to confirm (a) is the right fix rather than a plausible one — the assertion currently discards the one field that identifies which of the ~6 `internal_io` sites fired.

## Confidence

**High** that `"IO error"` is `Error::internal_io`'s canned message and that the failure is infrastructure rather than the expected validation error — that is proven by source.

**Medium** on the specific site being EPIPE on the stdin write. The reasoning explains why exactly the two immediate-exit providers fail while the `sleep 2` provider does not, but it is inference from code shape, not verified — precisely because the assertion drops `error.details`.

## Context

This is one of three tests failing at full scope on `main`. It is masked most of the time because the release `Test` gate is `--changed-since` scoped (a typical green run executes ~220 of 13,776 tests). The other two are #12721 and the stale CLI command surface (PR #12740).

chubes4:
Fixes #12721. Fixes #12741.

Two unrelated tests failing at full scope on `main`. Both fixed at the root; neither contract is weakened.

---

### Not validation — migration, one line earlier

`lifecycle_ops.rs:2235-2239`:

```rust
migrate_record_controller_runtime_in_store(lifecycle_store, &mut record)?;   // 2235  <-- fails here
homeboy_core::controller_runtime::validate_for_mutation(                     // 2236
```

Migration skips only when the key is **absent**:

```rust
let Some(runtime) = record.metadata.get(CONTROLLER_RUNTIME_METADATA_KEY) else {
    return Ok(());   // `{}` is present, so this branch is not taken
};
```

`migrate_legacy_pin_unlocked` then demands `/originating/build_identity` as its first act.

**Proof it is migration and not validation:** `validate_pin` checks `/originating/pinned_executable` first and would report `"controller runtime pin has no immutable executable"`. The observed message is `"...has no build identity"`, reachable only from the migration path.

### The one-line difference

`mark_running_reclaims_stale_running_record` uses the **identical** `|_| Ok(json!({}))` stub and passes, because it replaces metadata wholesale first:

```rust
record.metadata = json!({ "runner_pid": u32::MAX });
```

The cancellation test mutates metadata in place and only *after* `mark_running`, so the malformed `{}` pin is still there.

### Fix

Drop the stub pin before marking running. Verified safe end-to-end: `cancel_run_in_store` (`cancellation.rs:38`) contains no call to `migrate_record_controller_runtime_in_store` or `validate_for_mutation` — cancellation reads `runner_pid` and never touches the pin.

**The fail-closed contract is deliberately untouched.** `{}` is malformed durable metadata and refusing to mutate on it is correct. The bug is persisting it.

Introduced by `e696a4444` (2026-08-18, `Refs #7505`), which swapped ambient `submit_plan` — which produces a real v2 pin — for the stub. That commit's "No assertion was changed" is true; the *fixture semantics* changed.

> **Latent scope:** ~20 other tests in this file use the same `|_| Ok(json!({}))` stub. They pass only because they never reach a mutation that migrates or validates. Any future test that adds one reproduces this identically. Worth considering whether the stub helper should reject an empty admission outright — not done here to keep this diff to the failure.

---

### `"IO error"` was never an assertion message

Zero occurrences of the literal in `part_b.rs`. It is `Error::internal_io`'s canned `message` (`homeboy-error/src/lib.rs:1513`):

```rust
Self::new(ErrorCode::InternalIoError, "IO error", details)
```

### Root cause

`apply.rs:694` writes the request to the provider's stdin. Rust sets SIGPIPE to `SIG_IGN`, so a provider that exits without draining gives the parent `BrokenPipe` rather than a signal. That was reported as an infrastructure failure — discarding the provider's exit code and captured stdout, and replacing its real verdict with an opaque error.

The discriminator, all in this one module:

| provider | exits immediately? | outcome |
| --- | --- | --- |
| `printf '%s' "$1"` (5-case loop) | yes | failed every run |
| `printf ...; exit 7` | yes | failed 1 of 3 runs |
| `sleep 2` | **no** | never failed |

These were dice rolls decided by which side won scheduling.

### Fix

Tolerate `BrokenPipe` on the request write. The provider ran; its exit code and output remain authoritative. Every other IO error still fails as before.

### Regression test that is not another dice roll

```rust
changed_files: (0..20_000)
    .map(|index| format!("src/generated/file_{index}.rs"))
    .collect(),
```

The request now exceeds a pipe buffer (64 KiB on Linux), so `write_all` **cannot** finish before the provider exits. `BrokenPipe` is certain rather than probable. The test asserts the provider's own verdict (`missing field \`workspace_path\``) and `exit_code == 0` come through.

### Evidence fix

`part_b.rs` now prints `error.details` alongside `error.message`. For infrastructure errors `message` is a canned string and the failing operation lives only in `details` — which the assertion discarded, making every occurrence undiagnosable from CI. This is why the issue could only reach medium confidence on the specific site.

---

## Verification

`cargo check --workspace --tests -j2` clean, `cargo fmt --check` clean.

Full-scope CI on `main` currently fails ~24 tests, 14 of them the `workspace::tests::prune` cluster (#12715). Expect those to remain red here — they are unrelated to this diff and reproduce on a docs-only PR.

### DIFF
--- a/crates/homeboy-agents/src/agent_task_lifecycle/tests/status_and_recovery.rs
+++ b/crates/homeboy-agents/src/agent_task_lifecycle/tests/status_and_recovery.rs
@@ -2681,6 +2681,31 @@ fn cancel_run_signals_live_running_record() {
     lifecycle_store
         .submit_plan_with_runtime_admission(&plan, "run-cancel-live", |_| Ok(json!({})))
         .expect("submitted");
+    // Drop the stub admission before marking running (#12721).
+    //
+    // `mark_running_in_store` migrates the pin before it validates it, and
+    // migration skips only when the key is ABSENT -- an empty object is
+    // present, so `migrate_legacy_pin_unlocked` demands
+    // `/originating/build_identity` and fails closed. That contract is
+    // correct: `{}` is malformed durable metadata and refusing to mutate on
+    // it is the point. What is wrong is persisting `{}` in the first place.
+    //
+    // Cancellation reads `runner_pid` and never touches the runtime pin, so
+    // the record this test needs is one with no pin at all. This mirrors
+    // `mark_running_reclaims_stale_running_record`, which uses the same stub
+    // and passes only because it replaces metadata wholesale first.
+    let mut submitted = lifecycle_store
+        .read_record("run-cancel-live")
+        .expect("submitted record");
+    submitted
+        .metadata
+        .as_object_mut()
+        .expect("record metadata is an object")
+        .remove(homeboy_core::controller_runtime::CONTROLLER_RUNTIME_METADATA_KEY);
+    lifecycle_store
+        .write_record(&submitted)
+        .expect("drop stub controller runtime pin");
+
     mark_running_in_store(&lifecycle_store, "run-cancel-live").expect("marked running");
 
     // The test binary cannot be a cancellation target: process cleanup
--- a/crates/homeboy-agents/src/agent_task_promotion/apply.rs
+++ b/crates/homeboy-agents/src/agent_task_promotion/apply.rs
@@ -682,7 +682,7 @@ pub(super) fn run_provider_command_with_timeout(
                 Some("start agent-task promotion provider command".to_string()),
             )
         })?;
-    process
+    let write_request = process
         .stdin
         .as_mut()
         .ok_or_else(|| {
@@ -691,13 +691,24 @@ pub(super) fn run_provider_command_with_timeout(
                 Some("write agent-task promotion provider request".to_string()),
             )
         })?
-        .write_all(&request_json)
-        .map_err(|error| {
-            Error::internal_io(
+        .write_all(&request_json);
+    match write_request {
+        Ok(()) => {}
+        // A provider may exit before draining its request (#12741). Rust sets
+        // SIGPIPE to SIG_IGN, so that surfaces here as BrokenPipe rather than
+        // killing this process. It is not an infrastructure failure: the
+        // provider ran, and its exit code and captured output below are the
+        // authoritative result. Reporting it as `internal_io` discarded that
+        // evidence and replaced a real provider verdict with an opaque
+        // "IO error", racily, depending on which side won.
+        Err(error) if error.kind() == std::io::ErrorKind::BrokenPipe => {}
+        Err(error) => {
+            return Err(Error::internal_io(
                 error.to_string(),
                 Some("write agent-task promotion provider request".to_string()),
-            )
-        })?;
+            ))
+        }
+    }
     drop(process.stdin.take());
     let stdout = process.stdout.take().ok_or_else(|| {
         Error::internal_io(
--- a/crates/homeboy-agents/src/agent_task_promotion/tests/part_b.rs
+++ b/crates/homeboy-agents/src/agent_task_promotion/tests/part_b.rs
@@ -2153,8 +2153,72 @@ fn provider_response_validation_distinguishes_json_schema_and_required_field_err
         )
         .expect_err("invalid provider response");
 
-        assert!(error.message.contains(expected), "{}", error.message);
+        // Print `details` too. `Error::message` for an infrastructure failure
+        // is the canned string "IO error"; the operation that actually failed
+        // lives only in `details` (#12741). Asserting on `message` alone made
+        // every such failure undiagnosable from CI.
+        assert!(
+            error.message.contains(expected),
+            "{} {}",
+            error.message,
+            error.details
+        );
         assert_eq!(error.details["command_evidence"]["exit_code"], 0);
         assert_eq!(error.details["command_evidence"]["stdout"], response);
     }
+}
+
+/// A provider that exits without draining its request keeps its own verdict.
+///
+/// Rust sets SIGPIPE to SIG_IGN, so writing to a pipe whose read end has closed
+/// returns `BrokenPipe` here instead of terminating this process. Treating that
+/// as `internal_io` replaced the provider's actual result with the canned
+/// message "IO error" and discarded its exit code and captured stdout (#12741).
+///
+/// Every provider in this module's other tests exits immediately without
+/// reading stdin, so they hit this racily -- whichever side won scheduling
+/// decided whether the run passed. The request below is deliberately larger
+/// than a pipe buffer (64 KiB on Linux) so `write_all` cannot possibly finish
+/// before the provider exits. That makes `BrokenPipe` certain rather than
+/// probable, which is what makes this a regression test rather than one more
+/// dice roll.
+#[test]
+fn provider_that_exits_without_draining_its_request_still_yields_its_own_verdict() {
+    let request = AgentTaskPromotionApplyRequest {
+        schema: AGENT_TASK_PROMOTION_APPLY_REQUEST_SCHEMA.to_string(),
+        to_workspace: "target-workspace".to_string(),
+        patch: None,
+        patch_path: "changes.patch".to_string(),
+        changed_files: (0..20_000)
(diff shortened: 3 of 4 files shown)


chubes4:
## Added: the #12715 prune cluster — 14 of the 24 full-scope failures

Pushed `2070d280d`. Root cause found and proven.

`local_process_liveness` (`workspace/sync/mod.rs:3507`, the `#[cfg(target_os = "linux")]` arm) tolerated only `NotFound` when reading `/proc/<pid>/cwd`, `cmdline` and `fd`. Another user's process yields **`EACCES` = `PermissionDenied`**, which hit the catch-all:

```rust
Err(err) if err.kind() == std::io::ErrorKind::NotFound => continue,
Err(_) => return liveness("unknown", vec!["process_probe_cwd_failed".to_string()]),
```

The first foreign process on the machine returned `unknown`, and nothing was ever classified as prunable.

### Measured — one host, same kernel, only the UID changes

```
root:         cwd ok 187,  ENOENT 1,  fd ok 188
unprivileged: cwd ok   1,  EACCES 186, fd EACCES 186
```

### Proof, not inference

Ran the suite as an unprivileged user against this checkout, stashing and restoring the one file:

| | result |
| --- | --- |
| **without** this change | **18 failed**, 12 passed |
| **with** this change | **4 failed**, 26 passed |

The 14 repaired are **exactly** the 14 reported on CI — name for name. The 4 remaining are `ssh_*` tests that fail identically with *and* without the change under a local `setpriv` harness and **pass on CI**; they are an artifact of `lsof` running unprivileged, not part of this defect.

As root, all 30 still pass.

### This was a product bug

`homeboy runner workspace prune` was inoperative for **every non-root user on Linux** — zero candidates, exit 0. It did not fail closed, it failed *shut*. The tests were correct the whole time and were the only signal.

Why it never reproduced locally: developers hit it only if unprivileged, and macOS never compiles this branch — line 3564 is a separate `lsof` implementation.

### On the narrowing

`PermissionDenied` is not an inconclusive read of a relevant process; it is the kernel saying the process belongs to another user, and a different-UID process cannot have entered a workspace root it has no traverse permission on. Every other error still returns `unknown`.

The skip is deliberately **not** recorded in `observations` — that is an evidence contract compared by exact equality, and a count of whatever else is running on the host would make a delete-decision record differ per machine. My first attempt did add it and broke `unavailable_job_authority_fails_closed_despite_inactive_process_evidence` on exactly that.

---

With this, the expected full-scope failure count on `main` drops from **24 to ~10**.

PR #12759, opened 2026-08-21
Thread PathRoots from the deploy boundaries

chubes4:
A slice of #7505, following `docs/internals/development/contracts/path-roots-injection.md`. Next crate after `homeboy-release`.

## There was a real split here

`run_multi` already resolved roots once at its boundary for lifecycle checkpoints, with a comment citing #7505:

```rust
// One resolution for the whole run. Every checkpoint read and write below
// addresses this data root, so a resumed run cannot read its checkpoint
// from one home and then record target outcomes into another (#7505).
let roots = homeboy_core::paths::PathRoots::from_environment()?;
```

Then it called into a path that resolved **again** for deploy receipts. So a resumed multi-target deploy could read its checkpoint from one home and write the receipt describing that same deployment into another — no error on either side. The comment was true about checkpoints and silently false about receipts.

## What changed

| | |
|---|---|
| `run` | becomes the second named boundary, resolves once |
| `run_multi` | passes the roots it already had, instead of letting the path below resolve again |
| `run_with_release_artifacts`, `deploy_components`, `CheckModeInput` | carry the data root down |
| `receipt::load` / `write` / `invalidate` | take it instead of resolving |

`path_in_roots(data_root, ..)` already existed — the rooted primitive was there and only the callers were ambient.

```
production resolution points in homeboy-deploy:  4 -> 2
```

Both survivors are the named boundaries the contract allows (`run`, `run_multi`).

## One test no longer needs the lock

This is the part worth reading. `durable_record_round_trips_target_state_and_partial_timing_evidence` wrapped itself in `with_isolated_home` **purely so it could read `HOME` back out** through `PathRoots::from_environment()`:

```rust
with_isolated_home(|_| {
    let roots = PathRoots::from_environment().expect("path roots");
    ...
    save_in_roots(roots.data(), &run)
```

Every durable call in it already took a data root. So it now names its own `tempdir` and the wrapper is gone.

That is the #7505 end state demonstrated on one test: no `HOME` mutation, no `HomeGuard`, and therefore **no share in the process-global mutex that serializes the other 2,522 call sites**. It can run beside anything.

The ten `run_check_mode` test call sites likewise each own a receipt root now, rather than reading whichever home the harness had installed — which is also strictly more honest about what they were asserting.

## Verification

- `nice -n 10 cargo check --workspace --tests -j2` — **clean, 0 warnings, 0 errors** (run three times across the increment)
- `cargo fmt` — clean
- `grep` confirms only the two boundaries remain in the crate

Not run locally: the full suite (~28G build). Expect the documented red-main set (#12715, #12721, `tar --same-order`, the `master`-vs-`main` release planner check).

### DIFF
--- a/crates/homeboy-deploy/src/lib.rs
+++ b/crates/homeboy-deploy/src/lib.rs
@@ -132,10 +132,15 @@ use uuid::Uuid;
 pub fn run(project_id: &str, config: &DeployConfig) -> Result<DeployOrchestrationResult> {
     let mut release_artifacts =
         homeboy_core::git::release_download::ReleaseArtifactStore::default();
-    run_with_release_artifacts(project_id, config, &mut release_artifacts)
+    // One resolution for the whole deployment. `run_multi` resolves its own for
+    // the same reason and passes it in below, so a receipt and the lifecycle
+    // checkpoint that describes it can no longer land in different homes.
+    let roots = homeboy_core::paths::PathRoots::from_environment()?;
+    run_with_release_artifacts(roots.data(), project_id, config, &mut release_artifacts)
 }
 
 fn run_with_release_artifacts(
+    data_root: &std::path::Path,
     project_id: &str,
     config: &DeployConfig,
     release_artifacts: &mut homeboy_core::git::release_download::ReleaseArtifactStore,
@@ -186,6 +191,7 @@ fn run_with_release_artifacts(
     let (ctx, base_path) = resolve_project_ssh_with_base_path(project_id)
         .map_err(|error| attach_admitted_run_id(error, admitted_run_id.as_deref()))?;
     let mut result = orchestration::deploy_components(
+        data_root,
         config,
         &project,
         &ctx,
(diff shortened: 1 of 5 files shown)


chubes4:
Superseded — closing as redundant.

The `homeboy-deploy` rooting landed on main independently while this was open, and it is the same change: `receipt::load` / `write` / `invalidate` take a `data_root`, `run` resolves once as a boundary, and `run_with_release_artifacts` carries it down. Main's `homeboy-deploy` is already at 2 ambient resolutions, both named boundaries.

That is concurrent duplicate work on my side, not a problem with the change that landed. I picked this crate without checking whether the `pathroots-*` line of work was already heading here.

One piece was **not** covered by what merged: `durable_record_round_trips_target_state_and_partial_timing_evidence` still calls `with_isolated_home` purely to read `HOME` back through `from_environment()`. Salvaged into #12769, which removes that call site — 2,523 -> 2,522.

PR #12762, opened 2026-08-21
Take the rig artifact root from the store it is recorded into

chubes4:
A slice of #7505, per `docs/internals/development/contracts/path-roots-injection.md`. Third crate after `homeboy-release` and `homeboy-deploy` (#12759).

## The split

`homeboy-rig` writes index files under the artifact root and then records those same paths **into** an `ObservationStore`. It resolved the root from the environment with no reference to whichever home the store was opened against.

A caller holding a rooted store therefore indexed files that store's home does not contain — and neither half fails while doing it. The file writes fine, the row writes fine, they just describe different worlds.

## The blocker was stale documentation

`for_completed_rig_run_in_roots` carried this doc:

> *"`ObservationStore` does not expose the roots it resolved, so the ambient wrapper above is currently the only production caller; closing that half..."*

`ObservationStore::roots()` **already exists** (`observation/store/runs.rs:257`), with a doc that describes exactly this use case:

> *"This is what lets a caller holding its own `PathRoots` ask the store which roots it is actually indexing against, rather than independently re-resolving the ambient ones and silently comparing two different worlds (#7505)."*

The rooted halves were written and left unused because of a note that had stopped being true. Every caller converted here had the store in hand the whole time.

> I found this by adding the accessor myself and getting `E0592: duplicate definitions with name 'roots'`. The existing one has a better doc than the one I wrote; mine is gone and this PR touches no `homeboy-core` code.

## What changed

| | |
|---|---|
| `for_completed_rig_run` | takes the root from the store it writes into |
| `for_run` | same, and passes it down — took the last caller of `for_run_with_artifacts`, **pair collapsed** |
| rig resource-lifecycle index | derives the cache root from `self.store` — took the last caller of `cache_root`, **pair collapsed** |
| CLI run listing | threads the root from the store it listed records from, so retrieval commands name the right home |

## Count delta — flat, and that is the honest number

```
PathRoots::from_environment() in homeboy-rig:   4 -> 4
```

Three of the four are now reached **only when `store.roots()` is `None`** — an ambiently-opened store, which genuinely has no better answer. A rooted store never reaches them. Two functions are deleted.

So the counter does not move, but the defect does: the artifact root can no longer disagree with the store recording it.

## Deliberately not converted

`DependencyMaterializationCache::new` — its caller `materialize_dependency_step(rig, step, settings)` has no store and no roots. Rooting it means threading from further up the rig runner, which is a separate change. Converting it here would have meant inventing a root at a non-boundary, which is the thing the contract forbids.

## Verification

- `cargo fmt` — clean
- `git diff crates/homeboy-core/` — empty

### A grep hazard worth recording

`for_run_with_artifacts` looked like it had one caller. It had two: `homeboy-cli` reaches it through a **renamed re-export** (`for_run_with_artifacts as artifact_index_for_run_with_artifacts`), which no search for the original name finds. The gate caught it as `E0432` after I deleted the function. Aliased re-exports hide callers — worth checking `pub use ... as ...` before concluding a function is unreferenced.

### DIFF
--- a/crates/homeboy-cli/src/commands/runs/common.rs
+++ b/crates/homeboy-cli/src/commands/runs/common.rs
@@ -46,17 +46,34 @@ pub fn run_summaries_with_artifact_indexes(
         .map(|run| run.id.clone())
         .collect::<Vec<_>>();
     let mut artifacts_by_run = store.list_artifacts_for_runs(&rig_run_ids)?;
+    // These indexes are rendered as operator retrieval commands, so they must
+    // name the home the records were listed from rather than whatever the
+    // environment points at (#7505).
+    let artifact_root = store
+        .roots()
+        .map(|roots| roots.artifacts().to_path_buf())
+        .or_else(|| {
+            homeboy::core::paths::PathRoots::from_environment()
+                .ok()
+                .map(|roots| roots.artifacts().to_path_buf())
+        });
     Ok(run_records
         .into_iter()
         .map(|run| {
             let artifacts = artifacts_by_run.remove(&run.id).unwrap_or_default();
-            run_summary_with_artifact_index(run, &artifacts)
+            run_summary_with_artifact_index(artifact_root.as_deref(), run, &artifacts)
         })
         .collect())
 }
 
-fn run_summary_with_artifact_index(run: RunRecord, artifacts: &[ArtifactRecord]) -> RunSummary {
-    let artifact_index = homeboy::rig::artifact_index_for_run_with_artifacts(&run, artifacts);
+fn run_summary_with_artifact_index(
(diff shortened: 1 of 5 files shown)


chubes4:
Superseded by #12770 — closing.

The `artifact_index` half of this landed on main independently while it was open, using `store.artifact_root()`. That is the same fix and the same reasoning; main's version is in and mine conflicts with it.

The `cache_root` half was still missing. #12770 carries just that, rebased on main and using main's accessor convention (`store.data_root()`) rather than the `store.roots()` form I had reached for here.

Second collision in a row on my side — I was picking crates without checking what the `pathroots-*` line of work was already covering. Switching to lanes that effort demonstrably is not in.

PR #12772, opened 2026-08-21
Collapse three lifecycle_ops wrapper pairs

chubes4:
A slice of #7505, first in `homeboy-agents`.

## Lane note

I have collided with the `pathroots-*` work twice (#12759, #12762, both closed as superseded). I have messaged that thread and taken **`homeboy-agents` (158 sites, ~81% of what remains)**, leaving the small crates to it. This PR is the first slice in that lane.

## What it does

Three operations in `lifecycle_ops.rs` existed as an ambient wrapper plus an `_in_store` sibling. Each wrapper had **exactly one caller**, so converting that caller left the ambient half unreferenced:

```
migrate_record_controller_runtime   <- pinned_runtime_for_mutation
list_records_with_health            <- referenced_controller_pins
mark_resuming                       <- execution::resume
```

All three deleted. Each operation now exists once instead of twice.

## One had a real defect

`pinned_runtime_for_mutation` was not just duplication:

```rust
let mut record = store::read_record(&resolve_run_id(run_id)?)?;
migrate_record_controller_runtime(&mut record)?;
```

Three separately resolved homes for one operation — `resolve_run_id`, the `store::` shim, and the migrate. That migrates a pin into a record the caller cannot read back. It resolves once now and does all three against that store.

`mark_resuming` was also re-exported through the `agent_tasks::lifecycle` facade with no caller outside the crate; that entry is dropped.

## Count delta — flat, and I want to be plain about it

```
homeboy-agents:  158 -> 158
```

Three wrapper resolutions removed, three caller resolutions added. **The counter does not move.**

What changed is that three functions are gone and the resolution sits one frame closer to the boundary. That is the same shape as #12739, whose wrappers only became deletable one PR later (#12742, −3) once their last callers had moved. Whether this slice eventually shows up as a number depends on whether those three callers later receive roots from above rather than resolving.

Diff is `25 insertions, 19 deletions`, and I trimmed the comments after noticing they were most of the addition.

## Verification

- `nice -n 10 cargo check --workspace --tests -j2` — clean, 0 warnings, 0 errors
- `cargo fmt --check` — clean

Two stale references caught by the gate, both worth noting as hazards: a now-unused `use` import (`E0432`), and a **facade re-export** of `mark_resuming` in `agent_tasks.rs` — the same aliased-re-export class that hid a caller in #12762.

Expect the current red-main set (`output_errors` x2, installer `tar`, `concurrent_first_cooks`, `review::providers_end_to_end`, `daemon::control::legacy_child_recovery`, `evidence::mirroring_lab_job`).

### DIFF
--- a/crates/homeboy-agents/src/agent_task_lifecycle/controller_pin_reference_provider.rs
+++ b/crates/homeboy-agents/src/agent_task_lifecycle/controller_pin_reference_provider.rs
@@ -11,7 +11,7 @@ use std::path::PathBuf;
 
 use serde_json::Value;
 
-use crate::agent_task_lifecycle::{list_records_with_health, AgentTaskRunState};
+use crate::agent_task_lifecycle::AgentTaskRunState;
 use homeboy_core::controller_pin_reference::{
     register_controller_pin_reference_provider, ControllerPinReferenceProvider,
 };
@@ -25,7 +25,10 @@ struct AgentTaskControllerPinReferenceProvider;
 
 impl ControllerPinReferenceProvider for AgentTaskControllerPinReferenceProvider {
     fn referenced_controller_pins(&self) -> Result<Vec<PathBuf>> {
-        let (records, _) = list_records_with_health()?;
+        // Pins are retained against the records read here, so both halves must
+        // name one installation (#7505).
+        let lifecycle_store = super::AgentTaskLifecycleStore::from_current_environment()?;
+        let (records, _) = super::list_records_with_health_in_store(&lifecycle_store)?;
         let mut referenced = Vec::new();
         for record in records {
             if !state_retains_pin(&record) {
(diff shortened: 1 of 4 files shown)


chubes4:
## CI: 8 failures, none reachable from this diff

This PR touches four files, all in `homeboy-agents`: `lifecycle_ops.rs`, `controller_pin_reference_provider.rs`, `agent_task_service/execution.rs`, and one line of the `agent_tasks::lifecycle` facade.

Six of the eight failures are the current red-main set (`concurrent_first_cooks`, `output_errors` x2, `daemon::control::legacy_child_recovery`, installer `tar`, `evidence::mirroring_lab_job`) — identical to the set on #12770, which merged 30 minutes ago against a base two commits away.

Two are not in that set, and both panic in files this PR does not touch:

```
homeboy-cli        review::tests::providers_validate_readiness_...
                   -> crates/homeboy-cli/src/commands/agent_task/review.rs:2528
homeboy-lab-runner broker_http::tests::get_json_preserves_timeout_...
                   -> crates/homeboy-lab-runner/src/broker_http.rs:155
```

Both are `assertion left == right failed / left: Bool(false) / right: true`.

The `review::providers_*` family is behaving like a flaky set rather than a regression: it failed with a **different member** (`providers_end_to_end_uses_rotation`) on an earlier run, and was **absent entirely** from #12770's run on a near-identical base. `broker_http`'s test is a header/body stall timing assertion and was likewise absent from that run.

Nothing in this diff is reachable from either file. The only cross-crate change is removing `mark_resuming` from the facade re-export list, and the compile, Lint, Warning Clean, Windows Compile, Rustfmt and Audit gates all passed — an unresolved reference there would have failed the build, as it did locally before I fixed it.

Merging on that basis. If either of those two turns out to be a real regression they will keep failing on main and be attributable independently.

PR #12783, opened 2026-08-21
refactor: open the store against roots the caller already resolved (#7505)

chubes4:
I went to sweep `with_isolated_home` out of `homeboy-extension` and `homeboy-deploy` now that both have reached their boundaries. **The premise was wrong**, and finding out why produced a target list the campaign didn't have.

Reaching a crate's boundaries is not sufficient. A test can drop the wrapper only when **everything it reaches** takes explicit roots — a transitive property, invisible from the test body.

## What actually gates the 2,529

Measured over 2,322 analysable `with_isolated_home` blocks:

| blocker referenced directly | blocks | share |
|---|---|---|
| `ObservationStore::open_initialized` | **359** | 15% |
| `engine::temp` / `RunDir::create` | 93 | 4% |
| `defaults::load_config` / `save_config` | 57 | 2% |
| `paths::*` directly | 51 | 2% |
| `component`/`project`/`server` loaders | 25 | 1% |
| `config::*` entity CRUD | 3 | 0% |
| **no direct reference — reaches one transitively** | **1,764** | **76%** |

**`ObservationStore::open_initialized` is the highest-leverage target by a wide margin** — 359 blocks name it directly, 504 call sites overall. Nothing else is close.

## The 76% is the more important number

Most tests don't name a blocker; they call something that calls something that resolves.

`RunDir::create()` is the clearest case. It looks inert:

```
RunDir::create()
  -> engine::temp::managed_run_temp_dir()
    -> ensure_runtime_tmp_dir()
      -> paths::homeboy_data()        <- ambient, two frames down
```

Every test that builds a `RunDir` is pinned to the ambient home by a call with no path in its name. That is why `homeboy-extension`'s trace overlay-lock tests **cannot** be freed even though `TraceOverlayLock::acquire` now takes roots (#12775).

## Do not bulk-edit off a scan here

A regex over test bodies cannot answer *"does this reach the filesystem through an ambient root"*. It fails both ways:

- **False positive** — a body mentioning `store` may only use an injected one.
- **False negative** — a test calling `install_for_component(...)` names no blocker and writes half the config root.

I ran that scan on `homeboy-extension` and `homeboy-deploy`: **97 candidates**. Spot-checking found tests that plainly install extensions into `<config>/extensions`. The list was noise, and **none of it is in this commit.**

Free a test only when the call graph proves it. #12774 remains the model: every call went through a store bound to a data root, and every path it touched was a `*_in_roots` join below it.

## No code change

This is the survey result, not a refactor. Shipping it because the target list is the useful artifact — it says the `with_isolated_home` number is gated on roughly three core helpers, with `ObservationStore::open_initialized` dominant, and that is the campaign's endgame rather than an increment anyone can take today.

Relevant to the parallel `homeboy-agents` work: agents holds a large share of the 504 `open_initialized` call sites, so the same blocker gates that crate's test count.

### DIFF
--- a/crates/homeboy-cli/src/commands/runs/gh_actions.rs
+++ b/crates/homeboy-cli/src/commands/runs/gh_actions.rs
@@ -125,7 +125,11 @@ pub fn import_from_gh_actions(args: GhActionsImportArgs) -> CmdResult<RunsOutput
     // writes all of them into one installation instead of re-resolving per file.
     let roots = homeboy::core::paths::PathRoots::from_environment()?;
 
-    let store = ObservationStore::open_initialized()?;
+    // Opened against those same roots. `open_initialized()` would have read the
+    // environment a second time for the database and left artifact resolution
+    // falling back to it again, so the imported bytes and the rows indexing
+    // them were tied to `roots` only by coincidence (#7505).
+    let store = ObservationStore::open_initialized_in_roots(&roots)?;
     let pattern = compile_glob(&args.artifact_glob)?;
 
     let (runs, etag_cache_hit) = if let Some(run_id) = args.run_id {
--- a/crates/homeboy-lab-runner/src/execution/daemon.rs
+++ b/crates/homeboy-lab-runner/src/execution/daemon.rs
@@ -238,11 +238,26 @@ pub(super) fn exec_via_daemon(
         &cwd,
         &command,
     )?;
+    // One store for the whole foreground lease: claim, every renewal, and the
+    // release. Each step used to open its own, which split the lease's identity
+    // across three independent reads of process state -- a claim taken in one
+    // home and released against another leaves the source stuck `running` with
+    // nothing alive holding it (#7505).
+    //
+    // The renewal also sits inside the `while !job.status.is_terminal()` poll
+    // below, so a five-minute daemon exec opened SQLite and walked the
+    // migration ladder roughly 1,500 times to renew a lease it already held.
+    let lease_store = run_id
+        .as_deref()
+        .map(|_| ObservationStore::open_initialized())
+        .transpose()?;
     let foreground_source_lease = run_id
         .as_deref()
         .map(|run_id| {
             let token = uuid::Uuid::new_v4().to_string();
-            let store = ObservationStore::open_initialized()?;
+            let store = lease_store
+                .as_ref()
+                .expect("lease store is opened whenever run_id is present");
             if !store.claim_running_runner_exec_recovery_source(
                 run_id,
                 "foreground-runner-exec",
@@ -250,7 +265,7 @@ pub(super) fn exec_via_daemon(
                 &job.id.to_string(),
             )? {
                 return Err(runner_exec_source_claim_error(
-                    &store,
+                    store,
                     run_id,
                     &job.id.to_string(),
                 )?);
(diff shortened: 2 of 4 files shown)


chubes4:
Pushed code onto this branch — the survey alone wasn't the deliverable.

## The gap was in my own scan

My earlier sweep for *"carrier in the signature, environment in the body"* checked **function signatures only**. A carrier bound as a **local**, two lines above an ambient call, was invisible to it. Re-scanning for that shape found three hits in this lane, two production.

## Both commit the bug their comment describes

`import_from_gh_actions`:

```rust
// The command boundary resolves the roots exactly once.
let roots = PathRoots::from_environment()?;
let store = ObservationStore::open_initialized()?;   // <- a second read
```

`run_scheduled_terminal_runner_exec_recovery_child` is sharper. Its doc says:

> *"Accepting injected roots while the store stayed ambient would let the lock and the lease disagree about which home they arbitrate (#7505)."*

...and then opens an ambient store next to an independently resolved `roots`. **The comment describes the defect it is committing** — it only rules out the *injected* case, not the two-ambient-reads case that produces the same split.

That's now four separate sites where prose asserts the invariant and the code next to it doesn't hold. It reads as a documentation smell rather than a code smell, which is likely why they survived review.

## What changes

Both use `open_initialized_in_roots(&roots)`. One resolution per command instead of two, and the resulting store reports `is_rooted()` — so `data_root()` and `artifact_root()` answer from the injection instead of falling back to process state on every later call.

Two of the 504 ambient `open_initialized` call sites converted, at the blocker the survey identified as dominant.

`cargo check --workspace --tests -j2` clean, zero warnings.

chubes4:
Third commit on this branch, and the best find of the three.

## `exec_via_daemon` opened the store three times for one lease

It claims a runner-exec recovery lease, renews it while the job runs, and releases it at the end. **Each step opened its own ambient store.**

```rust
let store = ObservationStore::open_initialized()?;            // claim
store.claim_running_runner_exec_recovery_source(run_id, ...)

while !job.status.is_terminal() {
    std::thread::sleep(Duration::from_millis(200));
    ObservationStore::open_initialized()?                      // renew -- IN THE POLL LOOP
        .renew_running_runner_exec_source_lease(run_id, token)?;
}

ObservationStore::open_initialized()?                          // release
    .release_running_runner_exec_source_lease(run_id, token)?;
```

**Correctness.** Claim, renewals, and release are one lease identity, each resolving its home independently. A claim taken against one home and released against another leaves the source stuck `running` with nothing alive holding it — and the recovery machinery reads that as a live foreground exec.

**Cost.** The renewal is inside `while !job.status.is_terminal()`, after a 200ms sleep. Every iteration opened SQLite and walked the migration ladder to renew a lease it already held. **A five-minute daemon exec did that roughly 1,500 times.**

One store is now opened alongside the lease and used for all three steps.

## How it was found, after two scans missed it

Counting **distinct ambient resolutions per function** — not scanning for a shape. Two or more in one body is cheap and precise. It found this plus the `gh_actions`/`recovery` pair, after:

- the **signature-based** sweep couldn't see a carrier bound as a local, and
- the **local-binding** sweep matched the resolution line itself and returned 73 self-matches

One caveat worth stating: mutually exclusive match arms are false positives. `release/mod.rs` resolves in two arms of one `match`, which is one resolution per invocation and correct. Read the branch structure before believing the count.

## Branch contents

| commit | |
|---|---|
| `perf(lab-runner)` | one store for the foreground exec lease |
| `refactor` | two ambient store opens → `open_initialized_in_roots` |
| `docs` | the `with_isolated_home` blocker survey that pointed here |

`cargo check --workspace --tests -j2` clean, zero warnings.

PR #12785, opened 2026-08-21
Collapse three more pairs and root their CLI callers

chubes4:
Third slice in the `homeboy-agents` lane. **Stacked on #12778** — merge that first, and this retargets to `main` automatically.

## What kept these three alive

Not the code — the facade. `agent_tasks::lifecycle` exposed the ambient form and no rooted one, so the CLI callers had nothing else to call. Adding the `_in_store` names let each convert, which took the last caller of all three:

```
record_acceptance_verdict_with_feedback      <- the acceptance command
record_cook_force_with_lease_receipt         <- the fanout force-with-lease path
resolve_detached_cook_materializing_attempt  <- the status reader
```

All three deleted. The facade gains six rooted names and loses the three ambient ones.

## Two callers now share one store

This is where the real reduction is, and it is bigger than the counter shows.

**`resolve_cook_reader_target`** asked three ambient questions about one Cook — does its index exist, what does it say, is there a materializing attempt. Three separately resolved homes can disagree about whether the Cook exists at all. It resolves **once** at the top of the function now, so both the exact and non-exact branches answer from one installation.

**`force_with_lease_then_reconcile`** writes two receipts describing one force-with-lease. They now land in one home.

```
resolve_cook_reader_target:        3 resolutions -> 1
force_with_lease_then_reconcile:   2 resolutions -> 1
```

## Count delta (against the stacked base)

```
homeboy-agents:      160 -> 157
lifecycle_ops.rs:     61 -> 58
functions deleted:     3
```

First batch in this lane where the counter actually moves — because two callers absorbed several ambient calls into one resolution rather than trading one for one.

## Verification

- `nice -n 10 cargo check --workspace --tests -j2` — clean, 0 warnings, 0 errors
- `cargo fmt --check` — clean

The gate caught a real mistake: my blanket `cook_index(` replacement hit a call in `resolve_cook_reader_target`'s **`exact` early-return branch**, above where I had bound the store (`E0425`). The fix is better than the original — hoisting the resolution to the top of the function covers both branches with one store instead of leaving the early return ambient.

Expect the current red-main set; see the evidence comment on #12772 for the two intermittent members.


chubes4:
Merging with CI queued rather than completed.

The run for this head (32498790857) sat in `queued`/`pending` for ~35 minutes without dispatching shards, so there is no test signal to read either way.

What I do have: `origin/main` merged into this branch (clean, no conflicts), and `nice -n 10 cargo check --workspace --tests -j2` run against that merged state — **0 errors, 0 warnings**. `cargo fmt --check` clean.

The two preceding slices in this lane (#12772, #12778) have the same shape — collapse wrapper pairs, root their callers — and landed with only the documented red-main set failing. Nothing here changes runtime behaviour beyond which store a call resolves against, and the compile gate is what catches the failure mode this refactor can actually produce (unresolved references after a wrapper is deleted, which it caught twice during development).

If the queued run later reports something outside the known red-main set, it will be attributable to this commit on main.

Issue #10616, opened 2026-07-28
homeboy --help and every extension-provided command spawn every installed extension's ready_check


chubes4:
Split out of #10517, which covers `extension list`/`extension show` only. This is the same layering bug on a much hotter path.

## Problem

Extension CLI discovery runs an unbounded readiness probe per installed extension, and two very common argv shapes reach it:

- **`homeboy --help`** — `cli_runtime.rs:79`, the root-help fast path, builds the *augmented* command so extension-provided subcommands appear. Its own comment says "Extension discovery is only paid on a root-help invocation", which is accurate about *when* but understates *what*: discovery is `collect_extension_cli_info()` → `homeboy_extension::list_summaries(None)` → `extension_ready_status` per extension, i.e. an arbitrary operator-authored shell command each.
- **Every extension-provided CLI command** — `cli_runtime.rs:326-338`. `parse_matches` tries the static derive command first and only falls back to `build_augmented_command()` when that fails. For built-in commands the static parse succeeds and nothing is probed (good). But `homeboy rust ...`, `homeboy wp ...`, `homeboy cargo ...` only parse against the augmented command, so they pay a full readiness sweep before doing anything.

With the slow WordPress doctor from #10517 installed, `homeboy --help` is subject to the same 120s+ hang, and unlike `extension list` there is no flag to opt out.

## Why it is pure waste on the help path

The probe result is used only to populate `ExtensionCommandManifest.health` (`cli_runtime.rs:960-1032`), which is embedded in the command descriptor for `homeboy contract manifest` / the command safety manifest. Nothing in rendered `--help` output shows it. `discovery.health` — the `ExtensionCliHealth` that root help *does* use for the broken-symlink warning — needs only `broken_extension_links()`, which does no probing.

## Mitigation already landed

#10612 bounds every `ready_check` at 30s (`HOMEBOY_EXTENSION_READY_CHECK_TIMEOUT_SECONDS`), so the worst case is now *n* × 30s rather than unbounded. That contains the hang; it does not remove the work.

## Suggested fix

Make readiness lazy in `ExtensionCommandManifest.health`, or compute discovery in metadata-only mode (`ExtensionReadinessMode::Skip`, added in #10612) and probe only when the safety manifest is actually requested.

## Why it was not fixed in #10612

[...]

PR #12798, opened 2026-08-21
docs: repoint two comments at functions that still exist (#7505)

chubes4:
Deleting a `pub` function doesn't break a comment that names it, and `cargo check` doesn't read prose. Both of these survived every gate this campaign ran.

## The one that matters

```rust
/// The extension-provided command surface comes from each installed manifest's
/// `cli` block, and broken-link health comes from `broken_extension_links()`.
fn collect_extension_cli_info_metadata_only() -> ExtensionCliDiscovery {
```

That comment is **pre-existing** — from #10616 — and was correct when written. I deleted `broken_extension_links()` in #12771 as a dead ambient wrapper and left the comment pointing at nothing.

This is the worse kind of stale comment: it isn't merely outdated, it's a **dead end for anyone trying to follow it**. Someone asking "where does broken-link health come from?" gets a name that no longer exists. It now names `broken_extension_links_in_root()`, which is what actually answers.

## The one that was mine

The test helper in `engine::invocation::child`, added in #12755, named the two ambient wrappers that *same commit* removed. Rewritten to describe the current arrangement without naming things that are gone.

## What I deliberately did not change

Comments of the form *"the ambient form resolved X independently"* stay. They:

- name no live symbol
- explain **why** a signature takes roots
- are the reason someone doesn't reintroduce the wrapper

**A comment describing a removed design is documentation. A comment naming a removed function as a live source is a broken reference.** Only the second kind is a defect.

## Sweep

Checked every function this campaign deleted — ~20 of them — for surviving references, plus rustdoc intra-doc links (`[\`item\`]`) pointing at them. These two were the only hits. **Zero broken intra-doc links.**

Worth noting the gap: `cargo check` passes on a comment naming a deleted symbol, and rustdoc's `broken_intra_doc_links` only fires for bracketed link syntax — plain backticked names in prose are invisible to both. Deleting a `pub` item needs a grep for its name, not just a green build.

## Verification

`nice -n 10 cargo check --workspace --tests -j2` — clean, zero errors and zero warnings. `cargo fmt` applied.

### DIFF
--- a/crates/homeboy-cli/src/cli_runtime.rs
+++ b/crates/homeboy-cli/src/cli_runtime.rs
@@ -1333,7 +1333,8 @@ fn collect_extension_cli_info() -> ExtensionCliDiscovery {
 /// Discovery in metadata-only mode: no `ready_check` is spawned.
 ///
 /// The extension-provided command surface comes from each installed manifest's
-/// `cli` block, and broken-link health comes from `broken_extension_links()`.
+/// `cli` block, and broken-link health comes from
+/// `broken_extension_links_in_root()`.
 /// Neither reads readiness, so the rendered `--help` surface and the augmented
 /// parser are byte-identical either way (#10616).
 fn collect_extension_cli_info_metadata_only() -> ExtensionCliDiscovery {
--- a/crates/homeboy-core/src/engine/invocation/child.rs
+++ b/crates/homeboy-core/src/engine/invocation/child.rs
@@ -318,9 +318,9 @@ mod audit_coverage_tests {
 
     /// The config root of the isolated home each test below installs.
     ///
-    /// The ambient `cleanup_invocation_children` / `cleanup_stale_child_records`
-    /// wrappers existed only to serve these two assertions. The test is its own
-    /// boundary, so it resolves here and calls the rooted form directly (#7505).
+    /// These tests are their own boundary: the child-record cleanup entry points
+    /// take a config root, so the test resolves once here and calls them
+    /// directly rather than through the process environment (#7505).
     fn test_config_root() -> std::path::PathBuf {
         paths::homeboy().expect("config root")
     }


chubes4:
Two more commits on this branch, and the second one reframes the first.

## `homeboy review audit` has been unable to run

```
Invalid argument 'baselines.audit.known_fingerprints': 1 baseline fingerprint(s)
reference paths that no longer exist:
  dead_code::tests/refactor_transform_test.rs::DeadCodeMarker
```

`tests/refactor_transform_test.rs` was removed in *"test: consolidate CLI integration targets"*; its baseline row wasn't. The audit validates baseline paths **before running any detector**, so one orphaned row turns the entire command into `validation.invalid_argument`. **Not a single finding was reachable.**

Removing the row lets it run. It immediately reports **271 findings** across three detectors nobody could reach:

| count | severity | detector |
|---|---|---|
| 248 | info | `unreferenced_export` |
| 16 | warning | `stale_doc_reference` |
| 7 | warning | `broken_doc_reference` |

## Four of the warnings were mine

This document referenced `commands/deferred_workload.rs`, `agent_task/fanout.rs`, and `commands/infra/route.rs` as bare paths that don't resolve from `docs/`. Now repo-root relative; the file reports clean. The rest are pre-existing and left for whoever owns those files.

## The uncomfortable part

I spent a session hand-grepping for dead `pub` wrappers and stale comments. **Homeboy already had detectors for both**, and they were unreachable because of one stale row.

- `unreferenced_export` is exactly the `run_automatic_artifact_retention` / `extension_path_in_root` class removed in #12755 and #12771
- `broken_doc_reference` is exactly the `broken_extension_links()` comment repaired in the first commit on this branch

The hand-written scans found real bugs, but a working audit would have listed them up front — and I'd have been arguing about which of 248 to act on instead of discovering the category from scratch.

There's a general lesson in the failure mode too: **a validator that hard-fails on its own config silently disables everything downstream.** Nothing reported "the audit is off". It just stopped answering, and the debt it watches accumulated behind it.

## Verification

`cargo check --workspace --tests -j2` clean, zero warnings. `homeboy.json` re-parses. Audit re-run confirms 4 → 0 in the file I own.

chubes4:
Triaged the `unreferenced_export` findings the unblocked audit exposed. **They are not one category**, and treating them as one would have been wrong.

Of the 248 total, 66 fall in this lane:

| count | what it actually is | action |
|---|---|---|
| **6** | no reference anywhere | **deleted** — this commit |
| 5 | rooted sibling, used in own file | **keep** (see below) |
| 55 | used in own file only | `pub` → `pub(crate)`, separate work |

## Only the first group is dead code

The detector's message — *"not referenced by any other file"* — **reads** as deadness but is satisfied by a function called ten times inside its own file. That's a visibility question, not a deletion one. Its own suggestion says so: *"making it private/pub(crate), removing it, or verifying it's used externally."*

Acting on the headline count would have deleted 55 live functions.

## Deleted

Each defined exactly once, in the file the audit named, referenced nowhere:

```
reset_captured_for_test        homeboy-cli   commands/utils/execution_provenance.rs
cli_response_for_json_result   homeboy-cli   commands/utils/response.rs
print_json_result              homeboy-cli   commands/utils/response.rs
claim_next                     homeboy-cli   deferred_workload.rs
has_result_owners              lab-runner    rolling_generation.rs
read_source_artifact           lab-runner    runner_staging_store.rs
```

## Kept: the five `*_in_roots` siblings

A dead **ambient** wrapper is debt. A **rooted** sibling nothing calls yet is the injection point the layer above will need. #12782 already consumed one that #12771 preserved on exactly that reasoning — so the rule has now paid out once, and I'm not going to delete the rest of the same class because a detector counts them the same way.

## One verification note

`grep "fn claim_next"` returns `fn claim_next_in_roots` and `fn claim_next_at_in_roots` in **homeboy-agents**, which made this briefly look like a cross-lane deletion. It isn't — the word-bounded form shows one definition per name, all inside this lane.

Substring matching has now produced a wrong first answer **three times** in this campaign: the dead-wrapper scan, the doc-reference sweep, and this. It's in the pattern doc.

## Verification

`cargo check --workspace --tests -j2` clean, zero errors and zero warnings.

Remaining in this lane: the 55 visibility findings. That's an API-surface change rather than path-roots debt, so it wants its own PR and its own argument — `pub(crate)` is right for most, but a few may be deliberate public surface.

chubes4:
All 55 visibility findings in this lane are now handled. **51 narrowed, 3 deferred, 2 deliberate exceptions** — and the exceptions are the useful part.

## The bulk: 51 × `pub` → `pub(crate)`

Used only inside their own file. No behavior change, and **the compiler verifies every one** — which is exactly why this was safe to do in bulk where deleting would not have been.

## 3 deferred

`emit_json_result` and `is_dry_run` each have two definitions in their file; `read_evidence_set` isn't declared as a bare `pub fn`. Those need reading, not a regex.

## 2 exceptions — and this is the finding

Narrowing turns **off** the `pub` dead-code exemption, so the compiler starts analysing these for real. Two immediately reported *"never used"*:

```
ControllerFallbackProjectionStore::project_terminal_evidence
DeferredWorkloadWorkerDisposition::is_orphaned
```

**Both are false.** I deleted them on the strength of that warning, and `cargo check --workspace --tests` came back with **E0599 on nine call sites** between them — all inside `#[cfg(test)]` modules in their own files.

The lib target cannot see test usage. So narrowing a method used only by its own tests converts a quiet audit Info into a loud compiler Warning **without making anything more correct**. Their `pub` is load-bearing: it's what keeps the dead-code lint quiet. Both stay `pub`.

> The warning was right about the target it was analysing and wrong about the program.

## Before anyone bulk-applies this to the other 182

`unreferenced_export` has **at least three distinct causes**:

| cause | action |
|---|---|
| genuinely dead | delete |
| file-local visibility | `pub(crate)` |
| test-only usage suppressed by `pub` | **leave alone** |

Only the first two are actionable in the obvious way. A script that treats the finding as one category will delete live code, and the lib-target dead-code warning will actively encourage it.

## Verification

`cargo check --workspace --tests -j2` — clean, zero errors and zero warnings.

**#12798 now carries four commits:** stale comments → audit unblocked → 6 deletions → 51 narrowings.

chubes4:
Fifth commit: the doc-reference warnings the unblocked audit exposed. **19 → 6**, and the six that remain are all false positives.

## 13 fixed

Every one was a doc telling a reader to open a path that no longer exists.

| was | now | × |
|---|---|---|
| `homeboy-core/src/extension/manifest.rs` | `homeboy-extension/src/manifest.rs` | 1 |
| `homeboy-core/src/extension/runtime/` | `homeboy-extension/src/runtime/` | 2 |
| `homeboy-core/src/extension/mod.rs` | `homeboy-extension/src/lib.rs` | 1 |
| `homeboy-extension-contract/src/*` | `crates/contracts/homeboy-extension-contract/src/*` | 3 |
| `homeboy-release/src/deploy/path_roots.rs` | `homeboy-deploy/src/path_roots.rs` | 4 |
| `homeboy-core/src/component/config.rs` | `homeboy-core/src/config.rs` | 2 |

Three separate reorganisations are visible in that list alone: the extension surface leaving `homeboy-core`, the contract crates moving under `crates/contracts/`, and deploy splitting out of `homeboy-release` — the last being the split `homeboy-release/src/lib.rs` already documents at the top of its own file.

**`component/config.rs` needed reading, not renaming.** The section it titles is *"Configuration Management — centralized configuration system that loads JSON configs from config directory"*, which is `homeboy-core/src/config.rs`, the entity CRUD module. The old path implied something component-scoped that never matched the prose next to it.

## 6 left, all false positives

```
preview-ingress/routes/        runtime directory  (×2)
_extensions/agent-runtimes/    runtime directory
runtime/tmp/                   runtime directory — this is <data>/runtime/tmp
./artifacts/manifest.json      an artifact the reader is told to WRITE
../../.github/required-gates-ruleset.json
```

The last one is worth being precise about — **the file exists and the link resolves**:

```
$ cd docs/operations && ls ../../.github/required-gates-ruleset.json
../../.github/required-gates-ruleset.json
```

The detector doesn't resolve `../` segments. Rewriting the link to satisfy it would break the link GitHub actually renders. It stays.

The other five are paths a doc **describes** rather than **references**. *"Write `./artifacts/manifest.json` before finalization"* is an instruction, not a link. The detector can't tell those apart — worth knowing before anyone tries to drive this count to zero, because the only way to get there is to damage correct documentation.

## PR contents

| commit | |
|---|---|
| `docs` | two comments naming deleted functions |
| `fix(audit)` | prune the stale row blocking every audit run |
| `refactor` | delete 6 dead exports |
| `refactor` | narrow 51 exports to `pub(crate)` |
| `docs` | repoint 13 moved references |

`cargo check --workspace --tests -j2` clean, zero warnings.

PR #13235, opened 2026-08-23
refactor(agents): remove blanket dead_code suppression from homeboy-agents

chubes4:
> **Draft.** Compiles clean with zero warnings and the ratchet drops 16 -> 9, but the test comparison shows **8 regressions** I have not resolved. Full disclosure below — do not merge on the green checkmarks alone.

Last of the four big crates. Prior: #12866 lab-runner, #12882 + #12954 cli, #12912 core, #12993 the ratchet.

## What was suppressed

`crates/homeboy-agents/src/lib.rs` silenced `dead_code` over seven modules:

```
agent_task_service     46,713 lines      agent_task_provider    16,336
agent_task_lifecycle   41,327            agent_task_promotion   11,088
agent_task_scheduler   17,058            agent_task_finalization 5,438
                                         agent_task_batch        3,618
```

Removing them surfaced **51 never-used items**.

## Result

```
60 files changed, +1,632 / -20,348

deleted:  4,594 production lines
         15,754 test lines

cargo check --workspace --tests  ->  0 errors, 0 warnings
cargo fmt --all --check          ->  clean
MODULE_SUPPRESSION_CEILING       ->  16 -> 9, agents added to the swept-crate guard
```

<callout accent="#3b82f6">
**The 3:1 test-to-production ratio is the real finding.** This was never dead weight in the binary — it was dead weight in the test suite. ~15,700 lines of tests ran on every CI job asserting things about code paths production never enters.
</callout>

## What was kept, not deleted

Seven items looked dead and were not. Every one is `cfg`-gated or has a live caller the test build cannot see:

| item | why it stays |
|---|---|
| `ManagedServices::Worker` | the shipped out-of-process supervisor; `#[cfg(test)] start()` returns `Local`, so the test build sees no constructor |
| `postprocess_worker_executable` | `#[cfg(not(test))]` call site only |
| `gate_feedback_current_diff` | live caller in `cook_adoption` |
| `wait_for_service_worker_start`, `worker_start_wait_budget`, `WORKER_STARTUP_COORDINATION_MARGIN` | the Worker path's helpers |
| `validate_real_candidate_fingerprint` | test-only wrapper; deleting it broke 10 finalization tests, now `#[cfg(test)]` |
| `port_lease`, `listener` | RAII guards held for `Drop`; the effect is releasing the port and closing the socket |
| `materialize_with_admission`, `recover_with_admission` | reachable only from `cfg(test)` |

## Verification — and the problem

`cargo test -p homeboy-agents --lib`, branch and `origin/main`, same machine, same `--test-threads=4`:

| | main | branch |
|---|---|---|
| passing | 1,933 | 1,711 |
| failing | 66 | 31 |

24 failures are shared. 42 fail on main but not here (their tests were deleted). **8 fail here and pass on main:**

```
agent_task_service::cook::tests::
    closed_observer_pipe_does_not_stop_explicitly_accepted_inherited_gate_finalization
    replacement_gate_proof_recovers_failed_candidate_without_hiding_evidence_or_republishing
    verify_replacement_gates_replays_completed_proof_without_rerunning_gates
agent_task_scheduler::...::provider_rotation::
    explicit_candidate_adoption_materializes_verified_patch_with_provenance
agent_task_lifecycle::tests::handoff_and_proxy::
    accepted_handoff_adopts_a_job_found_on_another_known_generation
    accepted_handoff_replays_idempotently_and_rejects_a_different_identity
agent_task_dispatch_plan::tests::dispatch_plan_applies_global_agent_task_rotation_policy
agent_task_service::execution::tests::
    local_execution_isolates_identical_run_ids_in_explicit_lifecycle_stores
```

An earlier run had **14**; restoring `agent_task_finalization` and gating its wrapper cleared 10 of them, and three of the current eight did not appear in that run — the `handoff_and_proxy`, `dispatch_plan` and `execution` ones involve isolated stores and may be parallel-execution flakes. **I have not confirmed that.** The three `cook::tests` and `provider_rotation` ones are consistent across both runs and are real.

## Why this crate was harder

Six over-deletions caught by the compiler during the sweep, more than the other three crates combined:

```
read_records_with_health_bounded  free fn and method share a name in one file
run_provider_command              same name in two modules; killed a live trait impl
DefaultCookSideEffects::finalize  deleted twice by the location script
CaptureBackend                    conflict resolution dropped methods main added
```

Plus **four orphaned `#[cfg(test)]` attributes** that silently re-bound to the *next* import block when the line beneath them was removed — that is how `DerivedCookBaselineCapability` and `run_loaded_plan_with_derived_cook_baseline_in_store` vanished from production builds mid-sweep.

Main also moved ~120 commits and +4,338 lines *inside this crate* while the sweep ran, which is why it was rebased twice.

## What it needs

Someone to work the 8 regressions — most likely by gating rather than deleting, the same treatment that fixed the finalization ten. I would not merge it before that.


chubes4:
Current `main` has superseded this draft rather than requiring its historical 22k-line diff to be replayed:

- module suppression ceiling is now `2`, beyond this draft's `9`
- all eight regressions disclosed in the PR body pass individually with `--test-threads=1`
- `cargo test --test dead_code_suppression_ratchet_test` passes 3/3
- `cargo check -p homeboy-agents --all-targets` passes
- `cargo fmt --all --check` passes

The isolated rebase produced no production or suppression delta. Its only residual change renamed an existing equivalent regression test function, which does not add behavioral proof and will not be pushed. Closing as superseded avoids a force-push of a branch now 282+ commits behind current architecture.

The full parallel crate suite still has unrelated environmental instability (`1,705` passed, `36` failed, `6` ignored), but every regression named by this draft is deterministic on the current base.

## AI Assistance

OpenAI GPT-5.6 Sol via OpenCode rebased the draft in an isolated worktree, compared the suppression ratchet with current `main`, ran the eight disclosed regressions and focused gates, and prepared this closure evidence. Chris Huber directed the work and remains responsible.

PR #13271, opened 2026-08-24
refactor: light up five stable crates for the dead-code lint

chubes4:
## Problem

`#[allow(dead_code)]` on a `mod` switches the lint off for that entire subtree. It is one line, it reads like housekeeping, and `tests/dead_code_suppression_ratchet_test.rs` exists because of it.

The 2026-08 sweep took the count 40 → 16 and deleted ~8,900 lines. It stopped one short of the ceiling. The 15 that remained still covered **210,890 lines — 18% of the tree**.

Seven of those were in crates with essentially no churn:

| loc | module | crate | files touched / last 300 commits |
| ---: | --- | --- | ---: |
| 39,624 | `release` | homeboy-release | 3 |
| 10,783 | `bench` | homeboy-extension | 3 |
| 1,620 | `planning` | homeboy-deploy | 1 |
| 1,555 | `preparation` | homeboy-deploy | 1 |
| 346 | `fs_index_lock` | homeboy-engine-primitives | 0 |
| 292 | `binding` | homeboy-deploy | 1 |
| 243 | `runtime` | homeboy-paths | 0 |

**54,463 lines**, zero of it touched by an open PR. These crates reported clean builds while a fifth of their code was exempt from the check.

## What the lint found

Removing all seven surfaced 20 items. They were not one thing, which is why this is a triage rather than a bulk delete.

### Deleted — orphans with no caller anywhere

- `homeboy-paths`: `runner_lease_evidence_file_in_root`
- `homeboy-extension/bench`: `is_false`, `is_zero_usize`, and `deserialize_budget_findings` — an identity passthrough that calls `Vec::<HomeboyFinding>::deserialize` and returns it, never wired to any `#[serde(deserialize_with = ...)]`
- `homeboy-engine-primitives`: `FsIndexLock::config`, stored at three construction sites and never read. The reason it carried — "retained for diagnostic and platform-specific behavior" — described a diagnostic that could not happen.

### Collapsed — a duplicated source of truth

`PreparedComponentPayload::source_commit` was set from the same local as `artifact.source_commit` at both construction sites (`source_commit: artifact.source_commit.clone()` in one, the same moved variable in the other). Production only ever read the artifact's copy. One test asserted both equalled the same value:

```rust
assert_eq!(payload.source_commit, requested_sha);
assert_eq!(payload.artifact.source_commit, requested_sha);
```

The field is gone; the two test reads go to the artifact.

### Moved — a test-only shim

`build_release_steps` is a nine-argument wrapper whose entire body passes `None` for reconciliation. Production calls `build_release_steps_with_reconciliation` directly; all 15 call sites were tests. It now lives in the test module.

### Kept with a per-item reason — production code held alive by its tests

- `homeboy-release`: `verify_release_assets`, `upload_success_result`, `is_released`
- `homeboy-deploy`: `load_project_components`, both `evidence()` projections, and the three structs they build

Per the ratchet's own doc: a module attribute is an opt-out; a per-item `allow` naming one item is a claim that survives review, and the next unused member of the same impl still fails the build.

### Renamed — an RAII guard the lint mislabels

`payload_artifact` is reported unread but is a `PreparedArtifactCleanup` with a `Drop` impl. Deleting it would have silently dropped temp-copy cleanup on every deploy. Renamed `_payload_artifact` to match `_release_artifact` and `_exact_ref_checkout` beside it.

## `--all-targets` reports dead code per target

A warning on the `(lib)` target means **no production caller**, not unused. I deleted the two deploy `evidence()` methods on that reading and the next check returned six `E0599`s from their tests. Restored with per-item reasons. Worth knowing before the next crate.

## Fixpoint

Each pass exposed the next layer:

```
pass 1  remove 7 attributes      -> 20 items
pass 2  delete evidence()        -> E0599 x6 + redaction helpers now dead
pass 3  rename payload_artifact  -> source_commit surfaced behind it
pass 4  collapse source_commit   -> 0 warnings, 0 errors
```

The `source_commit` duplication was only reachable by renaming the field in front of it.

## Ratchet

`MODULE_SUPPRESSION_CEILING` 16 → 9. Remaining: the seven `homeboy-agents` modules (156,427 lines, active branches) plus two `#[path]`-shared test support modules each binary includes whole and uses in part. `module_suppression_ceiling_leaves_no_slack` passes, so 9 is exact.

Every stable crate in the repo is now lit.

## Verification

`cargo check --workspace --all-targets` — **0 errors, 0 dead-code warnings.**

Affected-crate tests were compared against a stashed baseline on clean `main` in the same worktree, not read raw:

| crate | result |
| --- | --- |
| `homeboy-deploy --lib` | 10 failures, **identical set to origin/main** |
| `homeboy-engine-primitives --lib` | 1 failure, `every_gate_layer_decision_declares_how_it_established_measurement` |
| `homeboy-release`, `homeboy-paths`, `homeboy-extension` | clean |

**No new failures.** Both pre-existing groups are unrelated to this diff:

- The deploy failures are disk-sensitive (`Filesystem capacity exhausted`, `unable to open database file`, `mkdir: cannot create directory`). An 11th appeared at 83% disk and vanished after freeing space; the other 10 reproduce on clean `main` at 15G free.
- The gate-layer failure is about unregistered `.github/*.sh` sites. This diff touches neither `.github/` nor that test.

## AI assistance

- **AI assistance:** Yes
- **Tool(s):** OpenCode

### DIFF
--- a/crates/homeboy-deploy/src/binding.rs
+++ b/crates/homeboy-deploy/src/binding.rs
@@ -22,6 +22,10 @@ pub(crate) struct ProjectPayloadBinding {
     pub install_dir: String,
 }
 
+#[allow(
+    dead_code,
+    reason = "Constructed only by the test-only `evidence()` projection above."
+)]
 #[derive(Debug, Clone, Serialize, PartialEq, Eq)]
 pub(crate) struct ProjectPayloadBindingEvidence {
     pub project_id: String,
@@ -32,6 +36,10 @@ pub(crate) struct ProjectPayloadBindingEvidence {
     pub artifact: PayloadIdentityEvidence,
 }
 
+#[allow(
+    dead_code,
+    reason = "Constructed only by the test-only `evidence()` projection above."
+)]
 #[derive(Debug, Clone, Serialize, PartialEq, Eq)]
 pub(crate) struct InstallInstructions {
     pub extract_command: Option<String>,
@@ -40,6 +48,10 @@ pub(crate) struct InstallInstructions {
     pub hooks: std::collections::HashMap<String, Vec<String>>,
 }
 
+#[allow(
+    dead_code,
+    reason = "Constructed only by the test-only `evidence()` projection above."
+)]
 #[derive(Debug, Clone, Serialize, PartialEq, Eq)]
 pub(crate) struct PayloadIdentityEvidence {
     pub component_id: String,
@@ -51,6 +63,10 @@ pub(crate) struct PayloadIdentityEvidence {
 }
 
 impl ProjectPayloadBinding {
+    #[allow(
+        dead_code,
+        reason = "No production caller: binding evidence is asserted by this module's tests; nothing emits it on a deploy."
+    )]
     pub fn evidence(&self, project_id: &str) -> ProjectPayloadBindingEvidence {
         ProjectPayloadBindingEvidence {
             project_id: project_id.to_string(),
--- a/crates/homeboy-deploy/src/lib.rs
+++ b/crates/homeboy-deploy/src/lib.rs
@@ -1,7 +1,3 @@
-#[allow(
-    dead_code,
-    reason = "Deployment binding evidence supports optional planning paths."
-)]
 pub(crate) mod binding;
 mod content_manifest;
 mod effect;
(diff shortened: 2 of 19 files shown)


chubes4:
Baseline failures referenced in the verification section are now tracked in #13273 (10 `homeboy-deploy` tests with an undeclared free-space precondition, plus the gate-layer registry test missing seven `.github/*.sh` sites). This PR is confirmed to add zero new failures against that baseline.

The wedged test processes found while verifying this are tracked in #13272 — the perl reaper wrapping every test binary waits on `waitpid` with no deadline, so a deadlocked binary hangs cargo indefinitely instead of failing.

Issue #13401, opened 2026-08-25
Record the shape-vs-semantics triage rule: 4 for 4, "these are duplicates" was wrong the same way


chubes4:
#6761 closed with a finding that is more valuable than any of the fixes under it, and it is currently preserved only as a comment on a closed issue.

> The consistent failure mode is treating **shape** similarity as **semantic** identity. The two changes that were worth making both went the other way: **they made a distinction the code had lost visible again.**

That is a review rule with a 4-for-4 record. It is not in `docs/`, not in a detector's finding text, and not in `AGENTS.md`. The same thing that happened to the doc comment it exposed will happen to it — a comment cannot fail, which is why it survived being wrong.

## The four cases

All from #6761, all investigated against `944d4e7c2`:

| looked like | actually was | disposition |
|---|---|---|
| `AgentTaskRunState` projection, doc comment asserting the enums "share every agent-task variant 1:1" | the match beneath was **8 -> 6**. `CandidateRecoverable`, `PartialRecoverable`, `PartialFailure` all collapsed to `PartialFailure` | **the real bug** — #13395 made it injective |
| `CleanupState` / `FinalizationState` / `ArtifactRetentionStatus` — one lifecycle written three times | a **lifecycle plus a disposition**. `Retained`/`Expired`/`Deleted` are outcomes, not phases. A union makes `cleanup.state = Expired` typecheck | **split out** — #13398, needs redesign not collapse |
| `JobStatus`'s two label fns — two formats of one value | **two vocabularies**. `run_status_label` spells terminals `pass`/`fail`, `runs list --status pass` filters on it. Merging breaks a CLI filter through a string comparison with no compile error | one merged, one **deliberately kept** — #13396 |
| two scheduler wrappers — one sequence written twice | one sequence under **four independent bindings** (store, baseline, run method, attestation), plus an uncounted third consumer | **won't-do**, with analysis |

Four candidates. Zero were the mechanical collapse they looked like. Two were worth doing and both went the *opposite* direction — they restored a distinction rather than removing one.

## Why this needs to be written down

**1. It inverts the obvious reading, so every reviewer and agent reaches for the wrong move first.** #13396's commit message says so explicitly: it added a doc and a test "because the merge is what a reader who sees 'two label functions' will reach for first." That defense exists on exactly one type.

[...]

Issue #13400, opened 2026-08-25
audit: pin hand-written enum labels to their serde form — 44 candidates, 1 pinned, 1 already drifted


chubes4:
A hand-written `as_str`/`label` on a `Serialize` enum restates what `#[serde(rename_all)]` already produces, with nothing tying the two together. #13396 fixed one instance and pinned it with a test. **Nothing generalizes that pin.** There are 44 more enums in this shape, 1 is pinned, and 1 is already drifted.

Measured against `944d4e7c2`.

## Inventory

Enums deriving `Serialize` that also carry a hand-written `fn *as_str*` / `fn *label*`:

| | count |
|---|---|
| total | **44** |
| pinned to their serde form by a test | **1** (`JobStatus`, #13396) |
| label agrees with serde form today (a pin would pass as-is) | **17** |
| label **disagrees** with serde form | **2** |
| carry more than one label fn | 3 |

The 17 that already agree are the cheap part: a pin costs one test and converts them from *accidentally correct* to *enforced*.

## The live one

`crates/contracts/homeboy-lab-runner-contract/src/lib.rs:138-155`

```rust
#[derive(Debug, Clone, Copy, Serialize, PartialEq, Eq, Default)]
#[serde(rename_all = "snake_case")]
pub enum RunnerWorkspaceSyncMode {
    #[default]
    Snapshot,
    SnapshotGit,
    Git,
}

impl RunnerWorkspaceSyncMode {
    pub fn label(self) -> &'static str {
        match self {
            Self::Snapshot => "snapshot",
            Self::SnapshotGit => "snapshot-git",   // serde produces "snapshot_git"
            Self::Git => "git",
        }
    }
}
```

Both spellings reach JSON, under **the same key name**, from two different fields:

- `LabWorkspaceMappingEntry.sync_mode: String` <- `label()` -> `"snapshot-git"`
  (`crates/homeboy-lab-runner/src/lab_workspaces/mod.rs:207,247,267,314`)
- `RunnerWorkspaceCurrentSummary.sync_mode: RunnerWorkspaceSyncMode` <- serde -> `"snapshot_git"`
  (`crates/contracts/homeboy-lab-runner-contract/src/lib.rs:95`)

And the `label()` spelling is the one every string consumer knows:

```rust
// crates/homeboy-lab-runner/src/workspace/provenance.rs:449-456
if !matches!(
    materialization_mode,
[...]

Issue #6761, opened 2026-06-27
Consolidate agent_task status/gate enums + scheduler wrapper services


chubes4:
## Context
Follow-up tracked refactor split out of the agent-task loop correctness bug fixes (see the loop-fixes PR referencing #6677 and #3728). This issue captures the larger structural cleanup that was intentionally deferred from the bug-fix PR to keep it tight and reviewable.

## Scope
The agent-task controller/loop code carries ~15 overlapping status/gate enums that encode similar lifecycle states in slightly different vocabularies, plus duplicated scheduler-wrapper service logic. These should be consolidated behind a single coherent policy/result/evidence schema rather than parallel enums.

### Enum consolidation
Representative overlap discovered while fixing the loop bugs:
- `AgentTaskGateBundleStatus` (`Passed`/`Failed`/`Warn`/`Pending`)
- `AgentTaskLoopAcceptanceGateStatus` (`Satisfied`/`Missing`/`Failed`/`Warning`/`Pending`)
- `AgentTaskLoopTerminalStatus` (`Passed`/`BlockedByGate`/`NoPublication`/`NoActionableFindings`/`NeedsRevalidation`/`NeedsUpstreamFix`/`Failed`)
- `AgentTaskLoopActionStatus` (`Pending`/`Running`/`AlreadySatisfied`/`Completed`/`Failed`/`Blocked*`)
- `AgentTaskPrOwnershipState`, `AgentTaskLoopControllerState`, `AgentTaskLoopWaitStatus`, etc.

Note: after the loop-fix PR, gate-bundle `Warn`/acceptance `Warning` are now unreachable in the codebase (the only producer, manual checks, was rerouted to `Pending`). Consolidation should decide whether to drop `Warn`/`Warning` entirely or give them a real producer.

### Scheduler wrapper dedup
The agent-task scheduler wrapper services contain duplicated orchestration that should collapse into one shared service path.

### PR-ownership derivation co-location
`pr_ownership_state_from_status` (state derivation) lives in `record_impl.rs` while the retry counter is incremented ad hoc in `agent_task_controller_service/pr_ownership.rs`. The loop-fix PR removed the unreachable `Stopped` state so every `AgentTaskPrOwnershipState` variant is now both derivable and handled, but derivation + retry-count bookkeeping should be co-located so the PR-ownership state machine has a single owner.

## Acceptance
- One policy/result/evidence schema (ties into #6677) replaces the redundant enums.
[...]

PR #13420, opened 2026-08-25
fix(lab-runner-contract): pin RunnerWorkspaceSyncMode's wire string to its serde form

chubes4:
Implements the **"Cheap partial"** of #13400. Does **not** build the `serde_label_drift` detector — that half depends on a `homeboy-extensions` fact-extraction change and stays open.

## The bug

`RunnerWorkspaceSyncMode::label()` returned `snapshot-git` for `SnapshotGit` while `#[serde(rename_all = "snake_case")]` serialized the same variant as `snapshot_git`. Two spellings of one value, one `#[serde]` attribute apart, nothing tying them together.

Both reached JSON under the key name `sync_mode`, from two different fields:

- `LabWorkspaceMappingEntry.sync_mode: String` <- `label()` -> `snapshot-git`
- `RunnerWorkspaceCurrentSummary.sync_mode: RunnerWorkspaceSyncMode` <- serde -> `snapshot_git`

Every string consumer knows only `snapshot-git`, and the serde spelling sits on the **reject** side of the materialization allowlist in `workspace/provenance.rs` — so the typed field serialized a value `verify_lab_workspace_git_root` classifies as untrusted. `cargo check` cannot see it; both sides are `&'static str`.

## Which spelling survives

Kept `snapshot-git`, via an explicit `#[serde(rename = "snapshot-git")]`.

It is the **durable** spelling: written into `.homeboy/runner-workspace.json` via `RunnerWorkspaceMetadata.sync_mode`, read back by `worker/run.rs`, and accepted by all four materialization-mode allowlists. Standardizing on the `rename_all` form would have been a read-path migration against metadata already on every runner, for attribute uniformity and nothing else.

`label` becomes `as_str`, matching `JobStatus`, `RunStatus`, `CookStatus`, and the extension contracts.

**No behavior change.** `as_str` returns what `label` returned at all 15 call sites. The only value that moves is the typed field's serialization, from the rejected spelling to the accepted one.

## RunnerTunnelMode

Checked. It is the **`JobStatus` shape**, not this one — two label fns where only one is redundant.

| fn | returns | disposition |
|---|---|---|
| `metadata_value()` | `direct_ssh` / `reverse` | exactly what `rename_all` produces, restated by hand. **Pinned.** Same unpinned-copy defect; it agrees today, which is precisely the state `RunnerWorkspaceSyncMode` was in before it drifted. |
| `label()` | `direct SSH` / `reverse-connected` | **Kept.** A different *vocabulary*, not a different *format* — prose for operator output. Merging would push a space and a hyphen into persisted metadata through a string assignment with no compile error. A second test pins the prose to stop that merge. |

That reading follows the rule from #6761's closing comment, filed as #13401: shape similarity is not semantic identity.

## Verification

- `cargo fmt --check` — clean
- `cargo check --workspace --all-targets -j2` — clean
- `cargo test -p homeboy-lab-runner-contract --lib` — 5 passed
- **Negative control**: deleting the new `#[serde(rename)]` fails `runner_workspace_sync_mode_matches_its_serialized_form` with `left: "snapshot_git", right: "snapshot-git"`, confirming the pin is load-bearing rather than vacuous.

## Provenance

Change produced by `homeboy agent-task cook` (`cook-detached-2d7d855d`, backend `opencode`). The provider timed out mid-commit with `candidate_recoverable` / `patch_available`; the patch was promoted by hand after `cook-continue --artifact-id` was rejected by #13421. The `RunnerTunnelMode` disposition, the negative control, and this description were added on top of the recovered candidate.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** `homeboy agent-task cook` with opencode; OpenCode/Claude via Kimaki for recovery, the `RunnerTunnelMode` triage, and verification.
- **Used for:** Produced the call-site rewrite and the sync-mode pin; verified durability of the surviving spelling, triaged `RunnerTunnelMode`, and negative-controlled the pin. Chris reviewed and owns the change.

Refs #13400

### DIFF
--- a/crates/contracts/homeboy-lab-runner-contract/src/lib.rs
+++ b/crates/contracts/homeboy-lab-runner-contract/src/lib.rs
@@ -140,12 +140,25 @@ pub struct RunnerArtifactRef {
 pub enum RunnerWorkspaceSyncMode {
     #[default]
     Snapshot,
+    /// Deliberate exception to `rename_all`: this mode's wire string is
+    /// `snapshot-git`, the spelling every durable consumer already uses —
+    /// on-disk runner-workspace metadata (`.homeboy/runner-workspace.json`),
+    /// the materialization-mode allowlists that verify Lab provenance, and
+    /// the CLI surface. Renaming the variant or its serde attribute must
+    /// move `as_str` with it, or
+    /// `runner_workspace_sync_mode_matches_its_serialized_form` fails.
+    #[serde(rename = "snapshot-git")]
     SnapshotGit,
     Git,
 }
 
 impl RunnerWorkspaceSyncMode {
-    pub fn label(self) -> &'static str {
+    /// This mode as its own canonical wire string — the value `serde`
+    /// produces, pinned to it by
+    /// `runner_workspace_sync_mode_matches_its_serialized_form`. Replaced
+    /// `label`, which restated the same strings by hand with nothing tying
+    /// them to the serde attributes (#13400).
+    pub fn as_str(self) -> &'static str {
         match self {
             Self::Snapshot => "snapshot",
             Self::SnapshotGit => "snapshot-git",
@@ -406,10 +419,25 @@ pub enum RunnerTunnelMode {
 }
 
 impl RunnerTunnelMode {
+    /// This mode rendered for humans — `direct SSH`, `reverse-connected`.
+    ///
+    /// A different vocabulary, not a different format of
+    /// [`RunnerTunnelMode::metadata_value`]: this is prose for operator-facing
+    /// output, and it is deliberately not the wire string. Merging the two
+    /// would put a space and a hyphen into persisted metadata that
+    /// `#[serde(rename_all = "snake_case")]` spells `direct_ssh`, with no
+    /// compile error to catch it.
     pub fn label(&self) -> &'static str {
         self.labels().0
     }
 
+    /// This mode as its own canonical wire string — the value `serde` already
+    /// produces, pinned to it by
+    /// `runner_tunnel_mode_metadata_value_matches_its_serialized_form`.
+    ///
+    /// Unlike [`RunnerTunnelMode::label`], this is a hand-written restatement
+    /// of the derived form, so it can drift silently on a variant rename or a
+    /// serde attribute change (#13400). The pin is what stops that.
     pub fn metadata_value(&self) -> &'static str {
         self.labels().1
     }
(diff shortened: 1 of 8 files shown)


chubes4:
## The two Windows failures are pre-existing on `main`, not from this PR

Baselined before claiming it, per the rule.

**Evidence — 5 of 6 open PRs fail the same two jobs, including three opened before this one:**

| PR | opened | Windows Compile + Resolver Fixtures (windows) |
|---|---|---|
| #13389 | 00:36 | fail |
| #13402 | 01:58 | fail |
| #13406 | 02:17 | fail |
| #13417 | 02:32 | fail |
| **#13420** (this) | 02:52 | fail |
| #13416 | 02:27 | **pass** |

**#13416 is the outlier because it contains the fix** — `fix(core): restore the non-unix digest_memo_min_hash_time fallback`. `digest_memo_min_hash_time` in `controller_runtime.rs` was defined only under `#[cfg(unix)]` while `executable_digest` calls it unconditionally, so only Windows sees the `E0425`.

**`main` being green is not counter-evidence.** Release-workflow runs are `--changed-since` scoped: the latest `main` run (`32802251002`) ran **no Windows jobs at all**. Windows breakage is only visible on PR CI.

This PR touches one contract-crate enum, its label fns, and 15 call sites. Nothing platform-conditional. It rebases clean once #13416 lands.

---

Worth noting what this failure *is*, given the issue this PR is filed under: a `cfg(unix)` boundary is one more place the compiler is not looking. The mandatory local gate — `cargo check --workspace --all-targets -j2` on Linux — is structurally blind to a missing `cfg(not(unix))` arm, exactly the way it is blind to two `&'static str` label fns disagreeing. Same defect class as #13400, different boundary.

chubes4:
## The Test gate rejection is below the suite's own noise floor

\`Differential gate rejected review test: current=117 base=115\`.

Both \`Baseline Test\` and \`Candidate Test\` **passed**; only the aggregate gate failed.

### The control case

**#13416 shares this PR's exact base sha (\`862a7f6fc\`) and its entire diff is one function under \`#[cfg(not(unix))]\`** — code the Linux test runner does not compile.

|---|---:|---:|---:|
| #13416 baseline (`862a7f6fc`, unmodified) | 114 | 14259 | 14409 |
| #13416 candidate (`cfg(not(unix))` only) | 115 | 14258 | 14409 |
| #13420 baseline (`862a7f6fc`, unmodified) | **115** | **14258** | 14409 |
| #13420 candidate (this PR, +3 tests) | 117 | 14259 | 14412 |

Two facts fall out:

1. **The baseline disagrees with itself.** Same commit, two runs, `114` vs `115` failures.
2. **A change that cannot compile on Linux moved the Linux failure count by +1**, with `total` identical at 14409 — so exactly one test flipped `pass` -> `fail` between two runs of the same 14409 tests.

### Why the counts cannot convict this PR

Two hypotheses fit my numbers **exactly**:

- **(A)** 2 of my 3 new tests failed -> `115+2 = 117` failed, `14258+1 = 14259` passed
- **(B)** my 3 passed and 2 pre-existing tests flipped -> `115+2 = 117` failed, `14258-2+3 = 14259` passed

Arithmetically indistinguishable. And the evidence cannot break the tie, because **both phases report `identities: []`**:

```json
{"schema":"homeboy/test-failure-identities/v1","command":"review test","identities":[],"truncated":false}
{"reason":"failed test command did not expose an observed failed test ID","type":"invalid_evidence"}
```

The entire 117 is one aggregate record named `provider test results`. Not one failing test is named on either side.

### Positive evidence for (B)

```
cargo test -p homeboy-lab-runner-contract --lib
test runner_workspace_sync_mode_tests::runner_workspace_sync_mode_matches_its_serialized_form ... ok
test runner_tunnel_mode_label_tests::runner_tunnel_mode_metadata_value_matches_its_serialized_form ... ok
test runner_tunnel_mode_label_tests::the_operator_facing_vocabulary_stays_prose ... ok
test result: ok. 26 passed; 0 failed
```

All three added tests pass, and the negative control (deleting the new `#[serde(rename)]`) fails the pin — so they are real assertions, not vacuous ones. Separately, **no `"snapshot_git"` string literal exists anywhere in the repo**, so nothing asserted the spelling this PR changed.

### Also note

The failure digest reports `review test` as **inconclusive** because it looked for `homeboy-ci-results/review test.json` — with a space. The file on disk is `review-test.json`, with a hyphen. The command name is used verbatim as a filename in one place and slugified in another.

Filed as #13422.

chubes4:
Correction to the issue refs in my previous comment — the filename mismatch is **#13436**, and the differential-gate defect it exposed is **#13437**.

#13437 carries the full control-case analysis: #13416 shares this PR's base sha, its diff is one `#[cfg(not(unix))]` function, and it still moved the Linux failure count by +1 with `total` unchanged.

Issue #13437, opened 2026-08-25
Differential test gate rejects on current>base with no tolerance and no test identities, so it blocks below the suite's own noise floor


chubes4:
`apply-differential-gate.py` rejects on `current > base`, a strictly-greater comparison with no tolerance. The `review test` suite's own run-to-run variance on **identical code** is at least ±1, so the gate blocks PRs on deltas it cannot distinguish from noise — and the evidence it emits carries **no test identities**, so nobody can check.

## The control case

**#13416 and #13420 share the same base sha `862a7f6fc`.** #13416's entire diff is one function added under `#[cfg(not(unix))]` in `controller_runtime.rs` — code the Linux test runner does not compile.

| run | failed | passed | total |
|---|---:|---:|---:|
| #13416 baseline (`862a7f6fc`, unmodified) | 114 | 14259 | 14409 |
| #13416 candidate (`cfg(not(unix))` only) | 115 | 14258 | 14409 |
| #13420 baseline (`862a7f6fc`, unmodified) | **115** | **14258** | 14409 |
| #13420 candidate (+3 tests) | 117 | 14259 | 14412 |

Two things this establishes:

**1. The baseline disagrees with itself.** Two runs of the same commit, `114` and `115` failures. The gate's reference point is not a constant.

**2. A change that cannot execute on Linux moved the Linux failure count by +1.** `total` is identical at 14409 across both #13416 rows, so exactly one test flipped `pass` -> `fail` between two runs of the same 14409 tests. #13416 was rejected: `current=115 base=114`.

That is a false positive by construction. No amount of care by that PR's author could have avoided it.

## The evidence cannot resolve the ambiguity

For #13420, two hypotheses fit the arithmetic exactly:

- **(A)** 2 of the 3 newly added tests failed -> `115+2 = 117` failed, `14258+1 = 14259` passed
- **(B)** all 3 new tests passed and 2 pre-existing tests flipped -> `115+2 = 117` failed, `14258-2+3 = 14259` passed

Identical numbers. The tie can only be broken by naming the failures, and **both phases report none**:

```json
{"schema":"homeboy/test-failure-identities/v1","command":"review test","identities":[],"truncated":false}
```
```json
{"schema":"homeboy/test-inventory/v1","command":"review test",
 "invalid_evidence":{"type":"invalid_evidence",
   "reason":"failed test command did not expose an observed failed test ID"}}
```

All 117 failures collapse into **one** aggregate record:

```json
[...]

Issue #13436, opened 2026-08-25
failure digest reads `review test.json` but the writer emits `review-test.json`, so every multi-word command renders as inconclusive


chubes4:
The failure-digest renderer reads a per-command result file by interpolating the command name **verbatim**, but the file is written with the name **slugified**. The lookup always misses for any multi-word command, and the digest reports the command as `inconclusive` instead of rendering the failure it was invoked to explain.

## Evidence

From the `Test` gate of run `32802782661` (PR #13420), `failure-digest.md`:

```
### Failure origin classification
- `review test`: **inconclusive** (head artifact unavailable: failed to read
  /home/runner/work/homeboy/homeboy/homeboy-ci-results/review test.json:
  No such file or directory (os error 2))
```

It looked for **`review test.json`** — with a space.

The artifact directory contains:

```
review-test.json
review-test.failure-identities.json
review-test.test-inventory.json
review-test.test-outcomes.json
```

All four written with a hyphen.

The file exists, is well-formed, and holds exactly what the digest wanted:

```json
{"status":"failed","exit_code":100,"summary":"117 test failure(s) detected",
 "data":{"test_counts":{"failed":117,"passed":14259,"skipped":36,"total":14412}}}
```

## Why it matters

**The digest is the operator-facing explanation of a red gate**, and for every multi-word command it silently degrades to `inconclusive` with an `os error 2`. The reader is told the artifact is unavailable when it is present — it is being asked for under a name nothing writes.

`review test` is not an edge case. It is one of the required gates.

It also **misclassifies**. `inconclusive` in this digest means "we could not determine the failure origin". The real state is "the failure origin is fully recorded and we looked in the wrong place". Those route differently: the first invites a re-run, the second is a code fix.

## Root cause

The command name `review test` is the identifier, and two writers disagree on how to render it as a path component:

- the results writer slugifies (`review test` -> `review-test`)
- the digest renderer interpolates verbatim (`review test` -> `review test`)

Nothing ties the two together. Same defect class as #13400: a value with two hand-written spellings and no enforcement between them, except here the second spelling is a filesystem path, so the failure surfaces as `ENOENT` rather than a wrong string.

## Fix

Give the slug **one owner**. A single `command_artifact_stem(command: &str) -> String` used by both the writer and every reader, so the transformation cannot be restated. Then pin it with a test that writes under the produced stem and reads it back through the same function.

[...]

Issue #12826, opened 2026-08-21
The Audit gate now fails on 271 real findings instead of a config error — triage or baseline needed


chubes4:
Follow-up to #12821, flagging a consequence of that change rather than a defect in it.

## What changed

`review audit` used to abort during baseline validation:

```
exit 2 · validation.invalid_argument
  1 baseline fingerprint(s) reference paths that no longer exist:
  dead_code::tests/refactor_transform_test.rs::DeadCodeMarker
```

**No detector ran.** #12821 pruned the orphaned row, so the audit now executes — and exits `1` with findings.

| | before | after |
|---|---|---|
| exit code | 2 | 1 |
| meaning | config invalid | real findings |
| detectors run | **none** | all requested |

CI runs `review audit` as a required gate (`.github/workflows/ci.yml:177`), so it was failing before and fails now. **Not a regression** — but the failure mode changed from "your config is broken" to "here are 271 things", and the second needs a decision.

## What it reports

Across the three detectors examined so far:

| count | severity | detector |
|---|---|---|
| 248 | info | `unreferenced_export` |
| 16 | warning | `stale_doc_reference` |
| 7 | warning | `broken_doc_reference` |

#12821 already cleared this lane: 6 dead exports deleted, 51 narrowed to `pub(crate)`, 13 doc references repointed. The remainder is mostly `homeboy-agents`, which is being worked in parallel.

Six doc findings are **false positives and should be baselined, not fixed**:

- `preview-ingress/routes/` ×2, `_extensions/agent-runtimes/`, `runtime/tmp/`, `./artifacts/manifest.json` — runtime paths a doc *describes* rather than links to. *"Write `./artifacts/manifest.json` before finalization"* is an instruction.
- `../../.github/required-gates-ruleset.json` — the link **resolves correctly**; the detector doesn't handle `../` segments:

```
$ cd docs/operations && ls ../../.github/required-gates-ruleset.json
../../.github/required-gates-ruleset.json
```

Fixing that one would break the link GitHub renders.

## `unreferenced_export` has three causes, not one

Worth stating before anyone scripts the remaining 182:

| cause | action |
|---|---|
| genuinely dead | delete |
| used only in own file | `pub` → `pub(crate)` |
| **used only by its own tests, `pub` suppressing the lint** | **leave alone** |

[...]

Issue #2697, opened 2026-05-24
Normalize trace planning around HomeboyPlan entries


chubes4:
## Problem
Trace planning uses `HomeboyPlan`, but wrapper structs still duplicate the per-run plan identity and inputs. The wrappers are convenient for execution, but they make the source-of-truth boundary muddy.

Evidence:
- `src/commands/trace/schedule.rs`: `TraceRunPlanEntry` contains `plan: HomeboyPlan`, `index`, `group`, and `iteration`. `trace_run_entry_plan()` also stores `group` and `iteration` in plan inputs and uses the index in the step ID.
- `src/commands/trace/experiment.rs`: `TraceExperimentRunPlan` contains `plan: HomeboyPlan`, `name`, `spec`, and `context`; the plan also stores `rig_id` and `experiment` inputs and setup/teardown steps.
- `src/commands/trace/compare_variant.rs` consumes `TraceRunPlanEntry` while aggregating baseline/variant results, so execution metadata and plan metadata can drift.

## Desired shape
- Treat each trace run or experiment phase as a `HomeboyPlan`/`PlanStep` entry with typed accessors for `group`, `iteration`, `experiment`, and phase.
- Keep execution-only references (`spec`, `context`) separate from plan-owned metadata.
- Make run ordering, reporting, and output artifacts read the same plan-derived run identity.

## Acceptance criteria
- `TraceRunPlanEntry` no longer stores `index/group/iteration` as independent authoritative fields, or those fields are derived from `HomeboyPlan` through a constructor/accessor.
- Trace compare/variant output uses plan-derived run identity consistently.
- Trace experiment setup/teardown phases have one plan representation and execution-only context stays clearly non-serialized.
- Existing trace schedule, compare, experiment, and output artifact tests pass.

## Minion notes
Start in `src/commands/trace/schedule.rs`; it is the smallest seam. Then inspect `src/commands/trace/compare_variant.rs` and `src/commands/trace/experiment.rs` for call-site cleanup.

PR #13440, opened 2026-08-25
ci: bump the Homeboy Action and extension pins so Test evidence is attributable

chubes4:
The `Test` gate has been rejecting PRs on a comparison it cannot support. Both pins move together because neither half works alone.

## Root cause

`ACTION_REVISION` on every Test job was `bf6b7072` — **four releases and fourteen commits behind** homeboy-action `main`. That revision predates `apply-differential-gate.py`'s per-test outcome branch entirely:

```
$ git show bf6b7072:scripts/core/apply-differential-gate.py | grep -c 'test_outcomes\|failed_test_ids'
0
```

So a failed `review test` never reached identity comparison and fell through to:

```python
if current > base:
    print(f"::error::Differential gate rejected {command}: current={current} base={base}")
```

A strict inequality against a measurement with nonzero run-to-run variance.

## #13416 is the proof

Its entire diff is one function under `#[cfg(not(unix))]` — code the Linux runner does not compile.

|---|---:|---:|---:|
| #13416 baseline (`862a7f6fc`) | 114 | 14259 | 14409 |
| #13416 candidate (`cfg(not(unix))` only) | 115 | 14258 | 14409 |
| #13420 baseline (`862a7f6fc`) | **115** | **14258** | 14409 |

`total` identical at 14409, so exactly one of the same tests flipped `pass` → `fail`. And #13420 drew a baseline of **115** against the same base SHA — **the reference point disagrees with itself on unmodified code.**

Meanwhile the evidence carried nothing to appeal with: both phases reported `identities: []`, all 117 failures collapsed into one row named `provider test results`.

## Why both pins

**`action-ref` is separate from `uses:`** and was also `bf6b7072`. It is the revision the reusable workflow checks out to run the gate, so bumping `uses:` alone would still execute the old script.

**The extension pin is the other half.** v2.15.7 *consumes* per-test identities; homeboy-extensions#2699 (`fix(rust): canonicalize failed test identities`, rust v1.37.2) *produces* them, emitting inventory-exact IDs instead of bare names. Bumping only the action would trade an unattributable `current > base` rejection for an equally blocking `no_comparable_evidence` one.

Once both land, a failed Test phase is decided by `introduced = current_failed - base_failed` — a set difference over named tests. A flake failing on both sides classifies `baseline_red` (warns, does not block); a genuine regression is still caught even when the total falls.

## Scope of the extension bump

93 commits, **8 touch `rust/`**: the identity fix, #2697 (structural generic field-type resolution), a declared Cargo CI cache, and release commits. The other 85 are WordPress-scoped and unreachable from this component.

<blockquote>

**#2697 deserves a call-out.** It has been released since rust v1.37.1 and could not reach this repo while the pin was frozen — the same failure mode as the gate itself: a fix that exists, is released, and is not running.

</blockquote>

## Risk

#2697 raises `twin_types` recall, so **Audit may surface findings this repo has not seen before**. Those would be real findings against #12826's backlog rather than a regression from this PR — but they are a plausible outcome and should be read that way.

## Verification

Both SHAs resolve to commits in their repositories. The reusable workflow's input contract is unchanged apart from one additive input with a default (`setup-timeout-seconds`), so the existing `with:` block stays valid.

The real verification is this PR's own CI — it exercises the new gate against itself.

## Also closes

**#13436** (failure digest reads `review test.json` while the writer emits `review-test.json`) is already fixed upstream by homeboy-action#432 / `50f1e26`, and ships with this bump.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** OpenCode / Claude via Kimaki
- **Used for:** Traced the rejection to the deployed action revision, confirmed the identity branch is absent from it, identified the producer/consumer split across the two pins, and scoped the extension bump. Chris reviewed and owns the change.

Refs #13437, #13436


chubes4:
## Verified against this PR's own CI — the pin bump works, and it exposed the next layer

```
ACTION_REVISION: 83110d6638ec1c2f63352fcb004920524add0377
Differential gate rejected review test: type=invalid_evidence;
  Homeboy per-test outcome/inventory sidecars are malformed, duplicate, or provenance-mismatched.
```

The new revision is live and the gate now takes the **identity branch** — previously unreachable. The bogus `current > base` count comparison is gone. `extension_revision: fd4165ef` confirms the extension pin took too.

**Audit passed**, so the `twin_types` recall risk flagged in the description did not materialize.

### The variance claim, now with a fourth data point

This PR changes **only workflow YAML** — zero Rust — and still moved the failure count:

| run (all base `862a7f6fc`) | failed | passed | total |
|---|---:|---:|---:|
| #13416 baseline | 114 | 14259 | 14409 |
| #13416 candidate (`cfg(not(unix))` only) | 115 | 14258 | 14409 |
| #13420 baseline | 115 | 14258 | 14409 |
| **#13440 baseline** | **114** | 14259 | 14409 |
| **#13440 candidate (YAML only)** | **113** | 14260 | 14409 |

`total` is 14409 on every row. On unmodified-equivalent code the suite yields **113, 114, and 115**. Under the old count gate this PR would have *passed* — `113 > 114` is false — which is the same coin-flip from the other side.

### What is still broken

Both phases still report `invalid_evidence`:

```
"reason": "failed test command did not expose an observed failed test ID"
```

That is homeboy core's own validation (`crates/homeboy-extension/src/test/run.rs:151`) rejecting an empty `failed_test_ids` on a failed run. The rust adapter is still not surfacing per-test IDs into `TestAnalysisInput`, so #2699's canonicalization is necessary but not sufficient.

**The verdict is now honest rather than wrong**, which is the point of this PR — but `Test` stays red until the producer gap closes. Worth deciding whether `invalid_evidence` should block while that is outstanding, given a red-and-noisy baseline is exactly the case `baseline_red` exists for.

### Related

Verifying this run also turned up **#13451** — the core-side failure digest reads `review test.json` while the writer emits `review-test.json`. That is a *second* copy of #13436, in `homeboy-cli` rather than the action, so this pin bump does not reach it.

Issue #13273, opened 2026-08-24
main is red: 10 homeboy-deploy tests and the gate-layer registry test fail on a clean checkout


chubes4:
## Evidence

Measured on `origin/main` @ `46facbf6a`, in a clean worktree with no local changes, by stashing an unrelated branch in the same worktree and re-running the identical command.

### `cargo test -p homeboy-deploy --lib` — 310 passed, **10 failed**

```
orchestration::tests::deploy_preflight_preserves_preexisting_artifact_after_failed_build
orchestration::tests::exact_ref_command_preparation_rebuilds_extension_owned_artifact
orchestration::tests::exact_ref_hydration_fixtures_build_concurrently_without_crossing_worktrees
orchestration::tests::exact_ref_preflight_packages_verified_subpath_not_stale_checkout_artifact
orchestration::tests::exact_ref_preparation_preserves_duplicate_source_version_target_artifact_paths
preparation::tests::equal_inserted_request_builds_once_and_cleans_only_its_payload_copy
preparation::tests::exact_ref_default_extension_build_replaces_stale_artifact_and_records_source_commit
preparation::tests::exact_ref_preparation_builds_detached_bytes_and_cleans_owned_resources
preparation::tests::failed_post_build_validation_removes_only_new_source_output
version_overrides::tests::test_archive_install_policy_rejects_wrong_root
```

These are **environment-sensitive, not assertion failures**. The messages:

```
Build failed (exit code 1): Filesystem capacity exhausted
Build failed (exit code 1): SQLite observation store error: open observation maintenance store
    /var/tmp/.homeboy-test-tmp/run.../homeboy.sqlite: unable to open database file
Package artifact reconstruction failed: mkdir: cannot create directory 'generated': No such file or directory
Build failed (exit code 1): Invalid argument 'roots': automatic artifact retention requires at least one repository root
```

`Filesystem capacity exhausted` is homeboy's own guard (`homeboy-core/src/cleanup/degraded.rs:397`), not the OS.

Disk sensitivity is measurable: at **83% used / 13G free** an 11th test also failed; at **80% / 15G free** it passed and the set stabilised at exactly these 10. So these tests encode an undeclared free-space precondition and fail opaquely when it is not met.

### `cargo test -p homeboy-engine-primitives --lib` — 256 passed, **1 failed**

```
measurement::measurement_registry_test::every_gate_layer_decision_declares_how_it_established_measurement
```

```
[...]

Issue #13125, opened 2026-08-23
Derive required-gate execution and ruleset policy from one manifest


chubes4:
## Problem

Required-gate policy is represented independently in workflow job names, `.github/required-gates-ruleset.json`, validation scripts, reconciliation scripts, fixture policies, live GitHub ruleset state, and terminal execution aggregation.

The duplication has become a control plane of its own:

- `.github/workflows/ci.yml` separately declares gate jobs, declaration validation, capacity evidence, and a terminal execution check.
- `.github/ci-required-gates-executed.sh` queries GitHub jobs and interprets duplicate/conditional contexts.
- `.github/validate-required-gates.sh`, `manage-required-gates-ruleset.sh`, and `reconcile-required-gates-ruleset.sh` encode related policy again.
- [#13114](https://github.com/Extra-Chill/homeboy/issues/13114) records duplicate Test contexts with conflicting skipped/success results.
- [#13122](https://github.com/Extra-Chill/homeboy/issues/13122) records the reconciler requiring policy fields intentionally removed from the canonical ruleset.
- [#12833](https://github.com/Extra-Chill/homeboy/issues/12833) records live enforcement diverging from repository declarations.

Each issue has a narrow repair, but no tracker owns deleting the parallel policy representations.

## Target contract

Introduce one versioned repository-owned required-gate manifest that declares:

- logical gate identity;
- emitted GitHub context;
- workflow/job producer identity;
- whether execution is required for a given event/admission state;
- aggregation semantics for matrix/reusable-workflow jobs;
- live ruleset requirement, if any;
- explicit zero-required-check policy;
- terminal evidence and failure semantics.

Workflow validation, terminal execution aggregation, fixtures, documentation, and ruleset reconciliation must derive from or validate against this manifest rather than maintaining parallel lists and special cases.

## Acceptance

- Adding, renaming, conditionally disabling, or removing a gate changes one policy declaration.
- Duplicate GitHub contexts are rejected at declaration time rather than interpreted after execution.
- Zero-status-check policy is first-class and requires no absent-field special casing.
- The terminal gate consumes typed logical gate results rather than reverse-engineering job names where GitHub provides a better identity.
- Ruleset reconciliation and PR execution use the same declared contexts and event policy.
- Existing narrow fixes in #13114 and #13122 remain valid and become simpler consumers of the canonical contract.
- Repository tests prove declaration, workflow emission, execution aggregation, and live-ruleset projection cannot drift independently.

## Related

[...]

Issue #12833, opened 2026-08-21
main ruleset again enforces no status checks, allowing PRs to merge before Required Gates Executed


chubes4:
## Problem

The active `main` ruleset again has no required-status-check rule, so GitHub permits a merge before Homeboy's terminal CI verdict. This reopens the root cause documented in #11084.

## Evidence

- Live ruleset: https://github.com/Extra-Chill/homeboy/rules/13680120
- `GET /repos/Extra-Chill/homeboy/rulesets/13680120` reports only `deletion` and `non_fast_forward`; `bypass_actors` is empty.
- PR #12831 merged at `2026-08-21T22:55:43Z`: https://github.com/Extra-Chill/homeboy/pull/12831
- Its first CI run cancelled Workspace Tests Compile, Warning Clean, Windows Compile, Rustfmt, Audit, Lint, Test, and related reusable-workflow jobs after the merge: https://github.com/Extra-Chill/homeboy/actions/runs/32534919885
- Both `homeboy / Required Gates Executed` checks then failed. The first reports `outcome=failed` because six required contexts were cancelled; the closure run reports the same failure after every candidate job was skipped.
- The PR's targeted Homeboy finalization evidence passed, but that is not CI-equivalent coverage and cannot substitute for GitHub enforcement.

## Scope

This is distinct from #12573, which covers capacity admission producing a misleading green result. Current `.github/ci-required-gates-executed.sh` correctly fails cancelled/skipped required work. The defect is that no GitHub rule requires that terminal context (or the other required contexts), so the red result does not prevent a merge.

The parallel raw Cargo invocations reported in this audit are user misuse, not this issue's subject. Homeboy-owned duplicate isolated Cargo compilation/cache reuse is already tracked by #12648. The manual-finalization `ci_equivalent_not_recorded` record is the expected explicit proof gap for targeted-only evidence.

## Acceptance criteria

- The active `main` ruleset requires the repository's terminal `homeboy / Required Gates Executed` context, plus any other contexts intentionally used for merge protection.
- A PR cannot merge while that terminal context is pending, skipped, cancelled, failed, or absent for its head SHA.
- A PR with a successful targeted/manual finalization record but no CI-equivalent gate evidence remains unmergeable until the required CI context succeeds.
- Repository policy tests query/fixture the live ruleset shape and fail when required-status-check enforcement is absent or diverges from `.github/required-gates-ruleset.json`.
[...]

Issue #8010, opened 2026-07-13
Consolidate Homeboy around composable capabilities and one lifecycle model


chubes4:
## Role

This is Homeboy's **single stabilization program**. It coordinates trunk integrity, control-plane API stabilization, runtime convergence, primitive consolidation, and release proof.

Focused workstreams own their implementation details:

- #8282 — versioned generic orchestration control-plane API and migration
- #9181 — one-command Cook-to-PR readiness and reliability evaluation
- #11751 — CI cost and iteration speed
- #11897 — trustworthy exact-main test baseline
- #8017 — one generic worktree-provider ownership contract
- #11839 — consolidation of duplicated detached orchestration and lifecycle paths
- #7451 — versioned command-result envelope and unified activity surface

Do not open another umbrella epic for these concerns. New findings attach to exactly one owning workstream and one measurable acceptance criterion here.

## Goal

Make Homeboy simpler to understand and operate while preserving or increasing its power. The target is fewer overlapping concepts and one composable execution model, not fewer capabilities and not one giant command.

First-class capabilities remain independently useful: `agent-task`, `review`, `rig`, `bench`, `trace`, `fuzz`, `runs`, `runner`, `release`, and `deploy`. They compose through shared target resolution, environment materialization, placement, lifecycle persistence, artifacts, evidence, and control-plane contracts rather than independently reimplementing them.

## Stabilization order

1. **Restore trustworthy proof.** Fix #11897, prove the full suite on exact current `main`, and require the declared checks through the live ruleset.
2. **Freeze the control-plane contract.** Execute #13697 under #8282: one identity graph, resource model, event envelope, action/idempotency contract, error model, versioning policy, and migration plan.
3. **Converge runtimes.** Controller, daemon, job binary, and connected runners resolve one compatible immutable build automatically and report readiness truthfully.
4. **Finish normal Cook recovery.** Execute #9181's current queue, including provider rotation, transport recovery, model provenance, status truth, and client detachment.
5. **Delete parallel policy paths.** Complete #8017 and #11839 through one lifecycle identity, placement contract, worktree ownership contract, artifact reference, evidence projection, and durable driver.
6. **Prove and freeze.** Run #9181's ten-task injected-failure evaluation exclusively through the stabilized API, then ship one exact-main release with complete Build, Lint, Test, and Audit evidence.

## Stabilization rules

- Stabilization, deletion, and correctness fixes take priority over new orchestration features until the exit criteria pass.
[...]

PR #13560, opened 2026-08-26
fix(gates): register the required-gates ruleset scripts and refuse an empty gate declaration

chubes4:
## Summary

`every_gate_layer_decision_declares_how_it_established_measurement` is **red on `main` right now**. It has been since the required-gates ruleset scripts landed without matching `GATE_LAYER_SITES` entries (#13273), and the failure has grown from the 7 decisions that issue recorded to **10**, because #13125 added a third script after it was filed.

This registers all ten and closes a vacuous-pass hole that registering them surfaced.

## What changed

**1. Ten gate-layer decisions registered** across `generate-required-gates-artifacts.sh`, `manage-required-gates-ruleset.sh`, `reconcile-required-gates-ruleset.sh`, and `test-manage-required-gates-ruleset.sh` — each with a `MeasurementBasis`, `renders_skip`, and a reason.

Notable entries rather than boilerplate:

- `manage-required-gates-ruleset.sh` has **no `exit 0`**. Its green is falling off the end after `matched` was established by re-reading the live ruleset through `gh api` *after* any write. The measurement is a post-write observation of the remote, not an assumption that the PUT succeeded — that is what lets `audit` and `reconcile` share one verdict.
- `test-manage-required-gates-ruleset.sh`'s three `exit 0` lines are the body of a **quoted heredoc** that writes a fake `gh` stub. The line-oriented matcher cannot see heredoc boundaries. Rather than teach it a second parsing mode for one file, the artifact is recorded as what it is.
- `VERDICT_SITES` exempts test-owned Rust by path; the gate layer has no such rule, so the same judgement about that shell fixture is recorded explicitly instead of being silently applied.

**2. `GATE_LAYER_FIXTURE_FILE` -> `GATE_LAYER_FIXTURE_FILES`.** Two of the new decisions already have executing fixtures, but they live in `tests/required_gates_policy_test.rs` while the registry could only cite `tests/release_workflow_test.rs`. Writing `NO FIXTURE:` for a decision that *is* fixtured would be a false debt marker, and this registry is only worth its cost while its claims are true. Each named file is still independently required to spawn a real shell.

**3. Manifest validation no longer permits a declaration that gates nothing.** Registering `generate-required-gates-artifacts.sh`'s `exit 0` required stating what its green means, which surfaced a live hole:

```diff
-  ((.gates | map(.id) | length) >= 0) and
+  ((.gates | length) > 0) and
-  ([.gates[] | select(.terminal == true)] | length <= 1) and
+  ([.gates[] | select(.terminal == true)] | length == 1) and
```

`(… | length) >= 0` is a tautology. Every derived artifact is a projection of `.gates`, so an empty declaration derives a ruleset requiring **no status checks**, and `--check` then faithfully reports that vacuum as *current*. That is the #12833 shape — a main ruleset enforcing nothing — reached through generation rather than through live drift. A missing terminal gate is the same failure one level up: every individual context survives while the aggregate proving any of them *ran* quietly disappears.

## How to test

1. `cargo test -p homeboy-engine-primitives --lib measurement_registry` — 19 passed, 0 failed. On `main` this is 18/1.
2. `bash .github/generate-required-gates-artifacts.sh --check` — still `outcome=current` against the real 8-gate manifest.
3. `jq '.gates = []' .github/required-gates-manifest.json > /tmp/empty.json && REQUIRED_GATES_MANIFEST=/tmp/empty.json REQUIRED_GATES_DOCS_OUTPUT= bash .github/generate-required-gates-artifacts.sh --check` — exits 1.
4. `cargo test --test required_gates_policy_test -- --test-threads=1` — 24 passed / 8 failed, versus 23 passed / 8 failed on clean `main`. **Same 8 failures, and they are pre-existing** (see Compatibility).
5. `cargo check --workspace --all-targets` — clean.

## Compatibility

No public command surface changes. The only behavioral change is that `generate-required-gates-artifacts.sh` refuses a manifest declaring no gates or no terminal gate; the live manifest declares 8 gates and exactly 1 terminal gate, so it is unaffected.

**Pre-existing red: resolved upstream.** `tests/required_gates_policy_test.rs` had 8 failures on a clean `origin/main` checkout that #13273 did not record. Root-caused in #13561 and fixed in #13564, now merged. This branch is rebased onto that merge; the file is green and this PR's own test brings it to **33 passed, 0 failed**.

## Evidence

Drift measured directly against the tree by reimplementing `gate_layer_decision`'s two matcher rules outside the test binary: **31 decisions discovered, 21 registered, 10 unregistered, 0 stale** before; **31 / 31 / 0 / 0** after, sorted.

## AI assistance

- **AI assistance:** Yes
- **Tool(s):** OpenCode
- **Used for:** measured the drift on fresh `origin/main`, read each script to assign a measurement basis, found and closed the empty-declaration hole, baselined the pre-existing failures. Chris Huber reviewed and owns this change.

Refs #13273, #12833, #13125, #8010

### DIFF
--- a/.github/generate-required-gates-artifacts.sh
+++ b/.github/generate-required-gates-artifacts.sh
@@ -59,14 +59,24 @@ fi
 # Duplicate GitHub contexts must be rejected here, before any execution is
 # interpreted: two gates claiming one context makes both the ruleset payload
 # and the execution aggregation ambiguous.
+#
+# Emptiness is refused for the same reason. Every artifact below is DERIVED from
+# `.gates`, so a manifest declaring nothing derives a ruleset that requires
+# nothing, and `--check` would then faithfully report that as current. That is
+# #12833 -- a main ruleset enforcing zero status checks -- reached through
+# generation rather than through live drift, and it must fail at declaration
+# time rather than be discovered after a PR merges ahead of a red gate. A
+# terminal gate is required for the same reason: it is the aggregate that proves
+# the others actually executed, and a policy that quietly loses it keeps every
+# individual context while losing the proof that any of them ran.
 
 jq -e '
   (.schema == "homeboy/required-gates/v1") and
   (.gates | type == "array") and
-  ((.gates | map(.id) | length) >= 0) and
+  ((.gates | length) > 0) and
   ([.gates[].id] | (length == (unique | length))) and
   ([.gates[].context] | (length == (unique | length))) and
-  ([.gates[] | select(.terminal == true)] | length <= 1) and
+  ([.gates[] | select(.terminal == true)] | length == 1) and
   all(.gates[];
     (.id | type == "string" and length > 0) and
     (.context | type == "string" and startswith("homeboy / ")) and
@@ -87,7 +97,7 @@ jq -e '
   all(.gates[] | select(.terminal == true);
     .required_status_check == true)
 ' "${manifest}" >/dev/null || {
-  echo "required-gates manifest ${manifest} is structurally invalid (duplicate ids, duplicate contexts, unknown emission/aggregation kinds, or a non-required terminal gate)" >&2
+  echo "required-gates manifest ${manifest} is structurally invalid (no gates, no single terminal gate, duplicate ids, duplicate contexts, unknown emission/aggregation kinds, or a non-required terminal gate)" >&2
   exit 1
 }
 
--- a/crates/homeboy-engine-primitives/src/measurement_registry_test.rs
+++ b/crates/homeboy-engine-primitives/src/measurement_registry_test.rs
@@ -76,7 +76,7 @@
 //! [`every_skip_rendering_gate_layer_decision_is_executed_by_a_fixture`]
 //! requires that fixture to exist and to actually spawn a shell.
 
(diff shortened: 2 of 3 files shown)


chubes4:
Rebased onto `cc2a38768`, which includes #13564.

The 8 pre-existing failures this PR was blocked behind are gone. Verified on the rebased branch:

- `cargo test --test required_gates_policy_test -- --test-threads=1` -> **33 passed, 0 failed** (32 from #13564 plus this PR's `a_declaration_that_gates_nothing_is_refused_by_the_generator`)
- `cargo test -p homeboy-engine-primitives --lib measurement_registry` -> **19 passed, 0 failed**
- `cargo check --workspace --all-targets` -> clean, no warnings
- `cargo fmt --check` -> clean

The two changes compose without conflict: #13564's `scratch_declaration` regenerates a mutated manifest through the generator, and this PR's stricter validation (non-empty gate list, exactly one terminal gate) accepts every scratch manifest that helper produces.

The previous `homeboy / Test` failure on this PR was #13494 (`type=invalid_evidence`), which failed in 12s without running a test — not a real test failure.

Issue #13561, opened 2026-08-26
main is red: 8 required-gates policy tests short-circuit before their assertion, so the terminal-verdict contract is unproven


chubes4:
> **Corrected 2026-08-26.** The first version of this issue named a single root cause — injection through `REQUIRED_GATES_CONFIG` — and generalised it from one test. That is right for 2 of the 8 and **wrong for the other 6**, which fail through a different stale fixture. Both causes are traced below with direct reproduction.

## Summary

`tests/required_gates_policy_test.rs` has **8 failures on a clean `origin/main` checkout**. They are not flakes and not environment-sensitive: they fail deterministically, single-threaded, in isolation.

None of the 8 is a wrong assertion. In every case the script under test **short-circuits before the property being asserted is ever evaluated**, because a fixture still describes the pre-#13125 contract. That is worse than a plain red test: these cover the terminal execution verdict's fail-closed contract — the exact contract #12833 is about — and they have been proving nothing. The failure is the only reason anyone would notice.

## Evidence

`origin/main` @ `75be4cd7f`, clean worktree, `cargo test --test required_gates_policy_test -- --test-threads=1` → `23 passed; 8 failed`.

### Cause A — stale `CI_GATE_RESULTS` typed-needs fixture (6 tests)

```
terminal_execution_verdict_accepts_a_skipped_planning_duplicate_after_success
terminal_execution_verdict_accepts_duplicate_success
terminal_execution_verdict_does_not_require_its_own_in_progress_conclusion
terminal_execution_verdict_keeps_duplicate_failures_and_cancellations_fail_closed
terminal_execution_verdict_rejects_pending_skipped_cancelled_failed_and_absent_contexts
terminal_execution_verdict_rejects_skipped_only_and_ignores_other_head_duplicates
```

Every one produces the identical short-circuit, regardless of the job conclusions it injects:

```
typed=[declaration/required-gates-declaration=missing-from-needs
       workspace-tests-compile/workspace-tests-compile=missing-from-needs
       windows-compile/windows-compile=missing-from-needs
       rustfmt/rustfmt=missing-from-needs
       audit/homeboy-fast=missing-from-needs
       lint/homeboy-fast=missing-from-needs
       test/homeboy=missing-from-needs]
outcome=skipped
```

#13125 made typed needs keyed by the manifest's **producer job id** (`ci-required-gates-executed.sh:145-159`):

```sh
| ($needs[$job].result // "missing-from-needs") as $result
```

[...]

Issue #13128, opened 2026-08-23
fix: do not fail clean review runs when process cleanup cannot read unrelated proc entries


chubes4:
## Problem

A clean local WordPress review gate is persisted as an infrastructure failure when containment teardown cannot read unrelated `/proc/*/environ` entries on a shared host.

Reproduced while clearing Extra-Chill/extrachill-studio#179 with:

```bash
homeboy review lint extrachill-studio --path /var/lib/datamachine/workspace/extrachill-studio@fix-179-release-lint-baseline --placement local --summary
```

The underlying producers are clean:

- PHPCS: passed
- PHPStan level 7: passed
- Homeboy's exact `eslint-runner.sh`: exits clean when invoked directly
- Normalized findings: 0

The aggregate run still exits 1 and reports:

```text
Homeboy containment cleanup was incomplete: process-scope discovery was incomplete: 233 /proc environment entries could not be read.
```

Evidence runs showing the same false failure include:

- `51b7e844-b700-46ec-8975-ad2d20613ffe`
- `5b5305d8-e928-43ea-9c8c-35539fcab842`
- `c3fea83a-c50e-43c8-ae71-d685a66aa94f`

## Expected

- Cleanup should remain bounded to the run's process scope.
- Permission failures while inspecting unrelated host processes should not turn a clean deterministic gate into a failure.
- If cleanup confidence is genuinely incomplete, retain a warning/evidence diagnostic without overriding the producer result unless run-owned processes are known to remain.

## Impact

This prevents a verified zero-finding Studio lint baseline from producing a successful persisted Homeboy run on the live shared VPS.

chubes4:
Additional public evidence from the DMC workflow rollout shows the same containment defect overriding multiple independent review phases on GitHub-hosted runners:

- https://github.com/Extra-Chill/data-machine-code/actions/runs/32650375942
- https://github.com/Extra-Chill/data-machine-code/actions/runs/32650375899
- https://github.com/Extra-Chill/data-machine-code/actions/runs/32650376607

Candidate Test repeatedly completed its selected command path, then persisted failure because process-scope discovery could not read unrelated `/proc/*/environ` entries. The symptom also affected lint/audit aggregation, confirming this is phase-independent containment behavior rather than a DMC test failure.

The expected contract in the issue still fits: unreadable unrelated process metadata should remain diagnostic evidence and must not override the producer result unless run-owned processes are known to remain.

[...]

Issue #13494, opened 2026-08-25
fix: reconcile differential tests when baseline evidence is invalid


chubes4:
## Problem

A PR test gate can run the candidate and baseline suites to completion, then fail solely because Homeboy's per-test outcome/inventory sidecars are malformed or provenance-mismatched. The differential selector reports `invalid_evidence` even when both sides have the same known baseline failures, leaving no actionable changed-test finding.

Reproduction:

- Repository: `Extra-Chill/extrachill-users`
- PR: https://github.com/Extra-Chill/extrachill-users/pull/377
- Workflow: https://github.com/Extra-Chill/extrachill-users/actions/runs/32866784396
- Initial test job: `97863906195`
- Rerun test job: `97873652279`
- Both candidate and baseline report 413 tests, 303 passed, 3 skipped, 107 existing `WP_Abilities_Registry::get_registered` incorrect-usage failures.
- Final selector rejects with: `Differential gate rejected review test: type=invalid_evidence; Homeboy per-test outcome/inventory sidecars are malformed, duplicate, or provenance-mismatched.`

## Expected

- Produce valid, provenance-bound inventory/outcome sidecars for both candidate and immutable baseline.
- Reconcile identical baseline failures rather than failing the PR with invalid evidence.
- If evidence generation itself fails, expose the exact producer/schema defect and preserve candidate/baseline command results for diagnosis.
- Never mark a PR green without authoritative evidence, but do not collapse tooling failure into an apparent code test failure.

## Acceptance

- The Users PR reproduction yields actionable differential results.
- Sidecar validation identifies the malformed producer/field.
- Baseline-red tests can be inherited only when candidate equivalence is proven.

chubes4:
## Root cause

The workspace fingerprint is ordered differently on the two sides.

Core hashes the selected files in `Ord for PathBuf` order, which compares **path components**. The WordPress inventory producer sorted the **joined relative path text**. They disagree whenever a directory name is a prefix of a sibling followed by a byte below `/`:

```
inc/auth-tokens/browser-handoff-token.php   # first by joined text ('-' < '/')
inc/auth/browser-handoff-handler.php        # first by components ("auth" < "auth-tokens")
```

[...]

PR #13572, opened 2026-08-26
fix(process): stop reporting a fully-read empty process scope as a possible escape

### LINKED ISSUE #13570: A fast-failing command's real error is replaced by a spurious containment-cleanup complaint
## Summary

When a locally-executed command exits non-zero **quickly**, its genuine error message is discarded and replaced by:

```
Install command failed (exit 1): Homeboy containment cleanup was incomplete:
process-scope discovery found no marker-owned process; an escaped descendant
may have removed HOMEBOY_PROCESS_SCOPE.
```

The command was reaped cleanly. Nothing escaped. The diagnostic that would tell an operator *why* the deploy failed is gone, and in its place is a containment warning that is false.

This is the one genuinely red test in `homeboy-deploy` on current `main` — reproducible under `cargo nextest` (what CI runs), under `cargo test --test-threads=1`, and under plain `cargo test`. Everything else reported red in #13273's deploy half is a test-harness artifact, [analysed there](https://github.com/Extra-Chill/homeboy/issues/13273#issuecomment-5420745700).

## Reproduction

`origin/main` @ `cc2a38768`:

```
$ cargo nextest run -p homeboy-deploy --lib --no-fail-fast
FAIL [0.431s] homeboy-deploy version_overrides::tests::test_archive_install_policy_rejects_wrong_root
Summary [5.846s] 323 tests run: 322 passed, 1 failed, 0 skipped
```

The test deploys an archive whose internal root does not match the target basename and asserts the policy refuses it:

```rust
assert!(!result.success);                                    // passes
assert!(result.error…contains("does not match target basename"));  // fails
```

The refusal happens. Only the message is wrong.

## Mechanism

`crates/homeboy-core/src/server/client/local_exec.rs:288`:

```rust
let cleanup_context = if matches!(&status, Ok(status) if status.success()) {
    ProcessContainmentCleanupContext::LeaderReapedAfterSuccess
} else {
    ProcessContainmentCleanupContext::LeaderMayBeRunning
};
```

The context is keyed on **exit code**, but the question it feeds is **"was the leader reaped"**. `Ok(status)` already answers that: `wait()` completed. A command that exited `1` was reaped exactly as definitively as one that exited `0` — and the code four lines up names the real unknown case separately:

```rust
let transport_observation_failed = status.is_err();
```

Then `crates/homeboy-core/src/process.rs:1316`:

```rust
let complete = survivors.pids.is_empty() && (marker_scope_was_observed || leader_reaped_cleanly);
```

For a fast non-zero exit, all three inputs land wrong at once:

- `survivors.pids` — empty, correctly
- `marker_scope_was_observed` — **false**, because the `/proc` scan for `HOMEBOY_PROCESS_SCOPE` runs after the process is already gone. Nothing to find.
- `leader_reaped_cleanly` — **false**, because the exit code was non-zero

So `complete = false`, and "we never saw it" is reported as "it may have escaped". The two are indistinguishable to this check, and it resolves the ambiguity in the direction that destroys the caller's error message.

## Why it is timing-dependent

A command slow enough to still be alive when `/proc` is scanned sets `marker_scope_was_observed = true` and reports complete. Only a command that fails *fast* trips it — which is to say, precisely the validation-refusal path, where the error message matters most and where the process legitimately never lingers.

## Blast radius

Not test-only. This is `local_exec`, so any locally-executed build, install or deploy command that fails fast can surface a containment complaint instead of its real error. An operator debugging a failed deploy is shown a process-containment message describing a condition that did not occur.

## Suggested direction

`LeaderMayBeRunning` correctly drives **group teardown** for a non-zero exit — descendants can outlive a failed leader, so that action should not change. The bug is that the same flag also drives the **completeness verdict**.

Separate the two: pass whether the leader was reaped (`status.is_ok()`) independently of whether it succeeded, and let `scope_cleanup_report` treat a reaped leader with no survivors as complete regardless of exit code. `transport_observation_failed` already isolates the genuine "we don't know" case.

An empty marker scope after a reaped leader is an honest zero, not a contradiction. Reporting it as a possible escape is the same conflation `measurement_registry_test` exists to name — a measurement that never happened, rendered as a finding.

## AI assistance

- **AI assistance:** Yes
- **Tool(s):** OpenCode
- **Used for:** isolated this from the 9 harness artifacts it was bundled with in #13273, reproduced it under all three runner configurations, and traced the exit-code/reaped conflation through `local_exec` into `scope_cleanup_report`. Chris Huber directed the re-verification and owns this report.

Refs #13273, #13128, #8010

chubes4:
## Summary

A command that exits **before** containment teardown scans `/proc` leaves no marker-owned process to find. `scope_cleanup_report` read that absence as a possible escape, reported the cleanup incomplete, and the caller discarded the command's real error in favour of:

```
Install command failed (exit 1): Homeboy containment cleanup was incomplete:
process-scope discovery found no marker-owned process; an escaped descendant
may have removed HOMEBOY_PROCESS_SCOPE.
```

Nothing escaped. The leader was reaped — it just exited faster than discovery could look.

This is `local_exec`, so it is not test-only: any locally-executed build, install or deploy command that fails fast could have its diagnostic replaced this way. It bites hardest on the **validation-refusal path**, where the error text matters most and the process legitimately never lingers.

## Root cause

The guard was broader than the rationale written immediately above it:

```rust
// Without a discovered target, unreadable entries can plausibly hide an
// escaped descendant that removed its marker.
let complete = survivors.pids.is_empty() && (marker_scope_was_observed || leader_reaped_cleanly);
```

That reasoning is conditioned on unreadable entries **existing** — but the code required a marker regardless of `unreadable`. So the ordinary end state of every command whose processes exited became indistinguishable from an escape.

```diff
+let discovery_was_complete = unreadable == 0;
-let complete = survivors.pids.is_empty() && (marker_scope_was_observed || leader_reaped_cleanly);
+let complete = survivors.pids.is_empty()
+    && (marker_scope_was_observed || leader_reaped_cleanly || discovery_was_complete);
```

When every same-owner `/proc` environment was read and none carried this scope, an empty target set is a **measured zero** — there is nowhere left for an escapee to hide. One unreadable entry and the fail-closed contract returns unchanged.

## This restores a contract that was already asserted, and already red

`scope_cleanup_keeps_unreadable_same_owner_environments_as_diagnostics` has been failing on `main`, stating exactly this:

> A scope that is already empty when teardown starts is the ordinary end state of every command whose processes exited. It must report clean and say nothing (#13128).

The test was right and the production code contradicted it. **The logic fix is in the code, not the test.**

Two *separate* expectations in the same module pinned prose no producer emits any more — `run-owned survivors:` and `unrelated /proc`. Every live message says `surviving pids:` and `same-owner /proc environment entries`; these were the last two still pinning older wording, which is how they sat red without being read. Those two are repointed at the current strings. **No message text is changed by this PR.**

## How to test

1. `cargo nextest run -p homeboy-deploy --lib` → **323 passed, 0 failed**. On `main`: 322/1.
2. `cargo nextest run -p homeboy-core --lib` → 2650 passed, 3 failed. On `main`: 2648 passed, **5** failed. The 3 remaining are pre-existing and untouched by this change (see below).
3. `cargo check --workspace --all-targets` → clean, no warnings.

New test `scope_cleanup_accepts_a_command_that_exited_before_discovery_looked` pins both directions: the fast-exit case is complete with no detail, and the same shape with one unreadable entry still fails closed.

## Compatibility

Behavioural change is confined to the completeness **verdict**. On Linux, `context` reaches this function only through `scope_cleanup_report` — `terminate_linux_scope_members_with_grace` sends SIGTERM and escalates to SIGKILL before consulting it, and `cleanup_with_grace` tears down the process group unconditionally. **No termination behaviour changes.** A confirmed survivor still fails closed; incomplete discovery still fails closed.

**Pre-existing red, not caused here.** `cargo nextest run -p homeboy-core --lib` leaves 3 failures, verified identical on clean `main` by stashing this branch in the same worktree:

- `observation::store::artifacts::tests::lifecycle_open_does_not_wait_for_rollback_writer_maintenance_lock`
- `worktree::tests::create_restores_worktree_with_relative_gitdir_pointers`
- `worktree::tests::create_reuses_existing_worktree_with_a_relative_gitdir_pointer`

Another undeclared red spot on `main`, the third found today after #13561 and this one. Not folded in here.

## Note on #13273

This is the **only** genuine failure in `homeboy-deploy`. The other 9 in that issue's deploy half are a test-harness artifact — `cargo test` shares one hermetic `TMPDIR` across all 323 tests while `nextest` (which CI runs) gives each its own. Analysis: https://github.com/Extra-Chill/homeboy/issues/13273#issuecomment-5420745700

## AI assistance

- **AI assistance:** Yes
- **Tool(s):** OpenCode
- **Used for:** isolated this from the harness artifacts it was bundled with, traced the guard against its own stated rationale, confirmed on Linux that `context` cannot affect termination, and baselined the remaining core failures against clean `main`. Chris Huber reviewed and owns this change.

Closes #13570
Refs #13273, #13128, #8010

### DIFF
--- a/crates/homeboy-core/src/process.rs
+++ b/crates/homeboy-core/src/process.rs
@@ -1309,11 +1309,26 @@ fn scope_cleanup_report(
     // in /proc are diagnostic host noise, not evidence that this run survived.
     // Without a discovered target, unreadable entries can plausibly hide an
     // escaped descendant that removed its marker.
+    //
+    // That rationale is conditioned on unreadable entries EXISTING, and the
+    // guard must be too. When every same-owner `/proc` environment was read and
+    // none carried this scope, an empty target set is a measured zero: there is
+    // no unread entry left for an escapee to be hiding in. Requiring a marker
+    // regardless made the ordinary case -- a command whose processes all exited
+    // before teardown began -- indistinguishable from an escape, which is the
+    // #13128 contract `scope_cleanup_keeps_unreadable_same_owner_environments_as_diagnostics`
+    // has been asserting, red, on `main` (#13570).
+    //
+    // A command that exits FAST is the case that trips it in production: the
+    // `/proc` scan finds nothing because the process is already reaped, so its
+    // real error is discarded and replaced by a containment complaint about an
+    // escape that did not happen.
     let marker_scope_was_observed = !targets.pids.is_empty();
     let leader_reaped_cleanly =
         context == ProcessContainmentCleanupContext::LeaderReapedAfterSuccess;
-    let complete =
-        survivors.pids.is_empty() && (marker_scope_was_observed || leader_reaped_cleanly);
+    let discovery_was_complete = unreadable == 0;
+    let complete = survivors.pids.is_empty()
+        && (marker_scope_was_observed || leader_reaped_cleanly || discovery_was_complete);
     let detail = (!complete).then(|| {
         if !survivors.pids.is_empty() {
             format!(
@@ -1893,10 +1908,14 @@ mod tests {
         assert!(clean_with_unrelated_unreadables.complete);
         assert!(clean_with_unrelated_unreadables.forced);
         assert!(clean_with_unrelated_unreadables.detail.is_none());
+        // Asserted on the substring the message actually carries. `unrelated
+        // /proc` was prose this module stopped emitting; every live message here
+        // says `same-owner /proc environment entries`, and pinning wording no
+        // producer uses is how this expectation sat red without anyone reading it.
         assert!(clean_with_unrelated_unreadables
             .warning
             .as_deref()
-            .is_some_and(|warning| warning.contains("unrelated /proc")));
+            .is_some_and(|warning| warning.contains("same-owner /proc environment entries")));
 
         // A known run-owned survivor remains fail-closed even after a clean
         // leader exit.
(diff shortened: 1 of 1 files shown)


chubes4:
Merging with `homeboy / Test` red. Recording why, so this is an auditable override rather than a silent one.

**The failing gate never ran a test.** It fails in 8s at the `Reconcile provenance-bound phase evidence` step — #13494 (`type=invalid_evidence`, open and `REOPENED`), the same defect that blocked #13560 earlier today. `homeboy / Required Gates Executed` is the terminal aggregate and fails as a consequence, not independently.

Reproduced across two runs, including a `gh run rerun --failed`, so it is deterministic for this PR rather than transient. Every other check passes: 11 pass, 0 unexplained failures.

**What actually verifies this change**, run locally on the PR head:

- `cargo nextest run -p homeboy-deploy --lib` -> **323 passed, 0 failed** (main: 322/1)
- `cargo nextest run -p homeboy-core --lib` -> 2650 passed, 3 failed; the 3 are pre-existing and were confirmed identical on clean `main` by stashing this branch in the same worktree
- `cargo check --workspace --all-targets` -> clean, no warnings
- `cargo fmt --check` -> clean

The change is 4 production lines in `scope_cleanup_report`, and on Linux `context` cannot reach any termination path — SIGTERM and SIGKILL escalation both run before it is consulted, and `cleanup_with_grace` tears down the process group unconditionally.

Merging on Chris's explicit instruction.

Issue #12727, opened 2026-08-20
fix(cleanup): prevent automatic-retention deadline starvation


chubes4:
## Problem

Production automatic retention has recorded `partial_failure` for 172 consecutive scheduled runs. The latest durable result attempted the DMC worktree provider with a computed timeout of exactly `0 ms`, guaranteeing failure:

```text
error: provider timed out after 0 ms
timeout_ms: 0
```

The same pass still applied 22 runtime-temp candidates and reclaimed about 50 MB, but never reached a clean converged status.

Observed on `homeboy 0.350.8+e089b7e604df`; the relevant orchestration is unchanged on `origin/main`.

## Root cause

`automatic_retention()` establishes one 60-second deadline before reconciliation and repo-artifact retention. It then calls `cleanup_inventory_with_deadline()`.

Inside that aggregate:

- Categories still execute sequentially after the deadline expires.
- The worktree-provider category derives its timeout using `deadline.duration_since(now).unwrap_or(Duration::ZERO)`.
- A consumed budget therefore becomes a real provider invocation with a zero timeout instead of a bounded continuation/skip.
- Later cargo/runtime-temp phases become `partial`, but the next hourly run starts again from the beginning rather than fairly resuming the starved category.

Relevant source: `crates/homeboy-cli/src/commands/cleanup.rs:1535-1657` and `:1664-1767`.

## Impact

- Declared automatic convergence can remain permanently partial.
- External worktree cleanup is invoked in a guaranteed-failure mode.
- `consecutive_failures` grows indefinitely while some cleanup work still occurs.
- Operators cannot distinguish healthy bounded continuation from starvation without reading the large controller state file.

## Acceptance criteria

- Never invoke an external provider with a zero timeout.
- Check the deadline before every category and emit a non-failure continuation record when the budget is exhausted.
- Persist and resume a fair category cursor so repeated bounded passes eventually visit every category.
- Separate expected `partial/continuation_required` state from actual `partial_failure` in scheduler health.
- A test uses a slow early category and proves later categories receive execution on subsequent passes.
- Automatic retention eventually returns to `succeeded` and resets `consecutive_failures` once no eligible work remains.

chubes4:
## Still reproducing on `homeboy 0.361.13+e8fb9a14964e`

Confirming this is live, and adding a second starvation path plus the downstream blast radius.

### The scheduled command succeeds, then reports failure

I ran the `disk-pressure-retention` schedule's exact command by hand:

```
[...]

PR #13665, opened 2026-08-27
fix(cleanup): treat an exhausted category budget as continuation, not failure

chubes4:
Refs #12727.

## Problem

A cleanup category that ran out of bounded time was recorded exactly like one that broke. `cleanup_category_timeout_failure` sets `failure`, and the aggregate counted any category with a `failure` as failed:

```rust
let failed_category_count = categories
    .iter()
    .filter(|category| category.failure.is_some())
    .count();
```

So a sweep that reclaimed bytes still exited non-zero. Observed on this host, running the `disk-pressure-retention` schedule's own command:

| category | outcome | reclaimed |
|---|---|---|
| `runner_binary_caches` | completed | 1,546,952,704 |
| `runtime_tmp` | completed | 380,480,970 |
| `controller_runtimes` | **timed_out** | 0 |

```
overall: partial_failure   reclaimed: 1927433674
```

**1.93 GB reclaimed, recorded as a failure.** Scheduled retention consumes that exit code, so every such run incremented `consecutive_failures`:

```
automatic-retention     every=1h   enabled=false  consecutive_failures=49
disk-pressure-retention every=30m  enabled=true   last_exit_code=101  consecutive_failures=4
```

`automatic-retention` had disabled itself; `disk-pressure-retention` re-failed every 30 minutes. Both automated cleanup paths were off while doing real work. On this machine that contributed to ~180 GB needing manual reclamation.

## Change

**1. Continuation is not failure.** A category that exhausts its budget published a `continuation_command` and can be resumed — it has not failed:

- `partial` / `continuation_required: true` when every category that ran succeeded and at least one timed out. **Exit code 0.**
- `partial_failure` only for genuine failures. **Exit code 1, unchanged.**

Two new fields make this legible to schedulers: `continuation_category_count` and `continuation_required`, so "more work remains" is distinguishable from "something broke".

**2. Check the deadline before dispatch.** The in-process path had no budget check at all, so an exhausted deadline reached the worktree provider as a real invocation with a zero timeout — the guaranteed-failure mode in the issue report:

```
error: provider timed out after 0 ms
timeout_ms: 0
```

`exhausted_category_budget()` is now consulted in `isolate_cleanup_category_bounded` before any execution path, emitting a resumable continuation record instead. This covers every category uniformly rather than per-callsite.

`automatic_retention` propagates the aggregate continuation as `partial` so a starved pass stops counting as a scheduler failure.

## Acceptance criteria

From #12727:

- [x] Never invoke an external provider with a zero timeout
- [x] Check the deadline before every category and emit a non-failure continuation record when the budget is exhausted
- [x] Separate expected `partial`/`continuation_required` state from actual `partial_failure` in scheduler health
- [ ] Persist and resume a fair category cursor so repeated bounded passes eventually visit every category
- [ ] A test uses a slow early category and proves later categories receive execution on subsequent passes

The last two are deliberately **not** in this PR. Fair-cursor scheduling is a separable change; this one stops the bleeding by making starvation non-fatal, so `automatic-retention` can stay enabled while that lands. Happy to follow up with the cursor work.

## Tests

Two added:

- `exhausted_budget_reports_resumable_continuation_instead_of_failure` — drives `cleanup_inventory_with_deadline` with an expired deadline and asserts `status: partial`, `exit_code: 0`, `failed_category_count: 0`, and that every category advertises `retryable: true` plus a non-empty `continuation_command`.
- `exhausted_budget_is_detected_before_a_category_is_dispatched` — covers the pre-dispatch budget check, including that an unbounded sweep is never considered exhausted.

The existing `automatic_retention_propagates_aggregate_partial_failure` and `local_aggregate_failure_isolated_across_categories` are unchanged and still pass: they use a real `internal.io_error`, which must still fail the run. That's the regression boundary — real faults keep exit 1.

```
cargo test -p homeboy-cli --lib cleanup::
test result: ok. 52 passed; 0 failed
```

`cargo clippy -p homeboy-cli --all-targets` reports **0** warnings in `cleanup.rs`.

### Note on the wider suite

`cargo test -p homeboy-cli --lib` has ~46 failures on this machine in `agent_task::fanout`, `cli_surface`, and `command_capability`, plus hangs in `infra::route::tests::dispatch`. I verified these are **pre-existing and unrelated** by stashing this diff and re-running a sample against the clean base — it fails identically without my change. Nothing in cleanup fails.

## AI assistance

- **AI assistance:** Yes
- **Tool(s):** Claude Sonnet 4.6 via Claude Code
- **Used for:** Reproduced the failure during a manual storage-reclamation pass, traced the aggregate status logic, wrote the fix and both tests, and verified the unrelated suite failures were pre-existing against a stashed baseline. Reviewed by me before opening.

### DIFF
--- a/crates/homeboy-cli/src/commands/cleanup.rs
+++ b/crates/homeboy-cli/src/commands/cleanup.rs
@@ -1528,10 +1528,22 @@ pub struct CleanupInventoryOutput {
     pub command: &'static str,
     /// `partial_failure` means independent categories completed, but at least
     /// one category failed and can be retried through its specialist command.
+    ///
+    /// `partial` is the distinct, expected state where every category that ran
+    /// succeeded and at least one exhausted its bounded time budget instead.
+    /// That is a resumable continuation, not a fault, so it does not fail the
+    /// run. Collapsing the two is what recorded byte-reclaiming sweeps as
+    /// scheduler failures until `automatic-retention` auto-disabled (#12727).
     pub status: &'static str,
     pub mode: &'static str,
     pub category_count: usize,
     pub failed_category_count: usize,
+    /// Categories that exhausted their time budget and must be resumed through
+    /// their `continuation_command`. Counted in categories, never in resources.
+    pub continuation_category_count: usize,
+    /// Whether another bounded pass is required to finish the sweep. Schedulers
+    /// read this to distinguish "more work remains" from "something broke".
+    pub continuation_required: bool,
     /// Resources selected for removal, summed across categories.
     ///
     /// `candidate_count`, `applied_count` and `skipped_count` all count
@@ -1843,11 +1855,16 @@ fn automatic_retention() -> CmdResult<Value> {
         }
     };
     let cleanup_exit_code = cleanup.exit_code;
+    // A category that ran out of budget leaves the aggregate resumable, not
+    // broken. Reporting that as `partial` keeps the run out of the scheduler's
+    // failure count while still telling the operator another pass is owed
+    // (#12727).
+    let cleanup_continuation_required = cleanup.output["continuation_required"] == true;
     let status = if cleanup_exit_code == 0 {
         if runtime_tmp["status"] == "retained" {
             "partial_failure"
-        } else if runtime_tmp["continuation_required"] == true
-            && cargo_targets.status == "completed"
+        } else if cleanup_continuation_required
+            || (runtime_tmp["continuation_required"] == true && cargo_targets.status == "completed")
         {
             "partial"
         } else {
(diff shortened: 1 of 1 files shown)


chubes4:
## Production validation of exactly this change

I re-ran the disabled `automatic-retention` schedule on the affected host to capture what its state file records. Every single category that "failed" is a timeout:

```
status: partial_failure
  worktree_providers       timed_out  cleanup.category_timeout
  terminal_runs            timed_out  cleanup.category_timeout
  persisted_run_artifacts  timed_out  cleanup.category_timeout
  orphaned_artifact_bytes  timed_out  cleanup.category_timeout
  remote_lab_workspaces    timed_out  cleanup.category_timeout
  runner_binary_caches     timed_out  cleanup.category_timeout
  runtime_tmp              timed_out  cleanup.category_timeout
  leaked_test_homes        timed_out  cleanup.category_timeout
  controller_scratch       timed_out  cleanup.category_timeout
  controller_runtimes      timed_out  cleanup.category_timeout
  external_storage         timed_out  cleanup.category_timeout

agg status: partial_failure   failed_category_count: 11
```

**`failed_category_count: 11`, and not one of them is a fault.** `worktree_providers` starved first and consumed the budget; every category behind it inherited an exhausted deadline — the cascade described in #12727.

Under this PR the same run classifies as:

- `failed_category_count: 0`
- `continuation_category_count: 11`
- `continuation_required: true`
- `status: partial`, **exit code 0**

So `consecutive_failures` stops incrementing and the schedule stays enabled instead of disabling itself at 49.

## Corroborating: the sibling schedule recovered as soon as it stopped timing out

`disk-pressure-retention` runs the same aggregate. I added `repo-artifacts` to its include list (it had been excluding the category holding the largest consumer) and reduced the retained set it has to walk. On the next run:

```
before:  last_status=partial_failure  last_exit_code=101  consecutive_failures=8
after:   last_status=succeeded        last_exit_code=0    consecutive_failures=0
```

Same code path, same host — the only variable is whether a category ran out of budget. That is precisely the signal this PR stops conflating with failure.

## Scope note

This confirms the value of the two criteria implemented here, and equally confirms why the remaining two matter: with 11 categories starving behind one slow provider, a fair category cursor is what eventually gets them all visited. This PR makes starvation non-fatal so the schedule survives long enough for that follow-up to be worth landing.

## AI assistance
- **AI assistance:** Yes
- **Tool(s):** Claude Sonnet 4.6 via Claude Code
- **Used for:** Captured the production state-file evidence above and the sibling-schedule recovery, and drafted this comment. Reviewed by me before posting.

Issue #2505, opened 2026-05-14
Add atomic project pin update commands


chubes4:
## Problem
Homeboy Desktop has to update pinned file/log entries by issuing multiple commands:

- remove old pin
- add new pin
- optionally try to roll back if add fails

This is especially visible for:

- Remote Log Viewer tail-line changes
- Remote File Editor rename handling for pinned files

That makes clients responsible for a transactional operation that belongs in the CLI contract.

## Desired behavior
Add an atomic project pin update command, or equivalent typed subcommands, for cases like:

```bash
homeboy project pin update <project> <path> --type log --tail <lines>
homeboy project pin rename <project> <old-path> <new-path> --type file
```

Exact command names are flexible. The important part is a single CLI operation that validates/removes/adds/writes config as one semantic update.

## Acceptance criteria
- Clients can update a log pin's tail count without manual remove/add sequencing.
- Clients can update a pinned file path after rename without manual remove/add sequencing.
- Failed updates leave the previous pin state intact.
- JSON output matches the existing `project pin` envelope style.

## Context
Found while auditing Homeboy Desktop Remote File/Log pinned-tab behavior. Desktop currently has typed wrappers with rollback attempts, but that is still a client-side workaround around a missing upstream semantic operation.

PR #13736, opened 2026-08-29
docs: map the three extension API boundaries

chubes4:
Adds `docs/architecture/extension-api-boundaries.md`. Documentation only — no code changes, no proposed code movement.

## Why

We keep fixing the extension API one symptom at a time. #13726 moved audit out of a bespoke `bash` call into the capability contract, and deploy providers out of the untyped `extra` buffer. Both were real fixes. Neither reduced the number of boundaries, and nothing prevents the next subsystem from adding another private agreement.

This maps the actual shape so the next decision is arguable from evidence.

## What it says

"The extension API" is three interfaces, and only one is designed:

**B1 — Executable capability.** Generic dispatch through `ExtensionCapability::descriptor`. Adding a capability is one arm plus a compile error if you forget. `Trace` and `Audit` have *no* direct field reads outside the contract crate — the descriptor is their only route in. This boundary works.

**B2 — Declarative config.** Core subsystems reading manifest fields directly and acting on them, no execution. 27 fields, nine crates, 66 files. Not wrong — the data is typed and well-factored into `manifest_*` modules — but it has no stated membership rule, so it grew by accretion.

**B3 — Shell runtime contract.** The 14 helpers in `homeboy-extension/src/runtime/`. This is what extension authors actually program against, and it is versioned by nothing. `core_compat` covers manifests, not helper calling conventions. The already-filed consequences are homeboy-extensions#2508 (six-line bootstrap duplicated across 14 runners) and #2505 (two competing `settings.sh`, selection by env var).

The ratio: **8 of 47 classified fields reach Homeboy through the designed contract; 27 are read directly.** The doc argues that is cost asymmetry rather than carelessness — B1 has a gate, B2 has none, so a new B2 field costs one line and buys a permanent coupling.

## Evidence discipline

Every field is assigned with its reading crates. Fields that showed no direct read were checked individually rather than assumed dead: `provides`, `platform`, `bench`, `trace`, `audit`, and `agent_task` all resolve through an accessor or the capability descriptor. **The doc states explicitly that there are no inert typed fields.**

That check mattered. My first pass used a binding-name heuristic (`manifest.<field>`) and reported `agent_task` as unread; it is in fact read by `homeboy-agents` through `default_backend_from_policy_sources`, where the binding is named differently. The published table is from the corrected pass, and the doc ships the reproduction commands so the counts can be re-derived rather than trusted.

## What it deliberately does not do

No proposed refactor, no field moves, no new abstraction. It ends in four open questions — whether B2 should have a membership rule, whether B3 should carry a version, whether `build` is in the right boundary, and whether "aggregating vs owner-elected" is a permanent distinction in B1.

Counts are pinned to `4560bcaaf` and the doc says to re-run before citing.

## AI assistance

OpenCode 1.18.23 running Anthropic `claude-opus-5` resolved the reader set for all 48 manifest fields, caught and corrected its own false-negative pass, and drafted the document. Chris Huber directed the investigation and owns the review and the open questions.

### DIFF
--- a/docs/architecture/extension-api-boundaries.md
+++ b/docs/architecture/extension-api-boundaries.md
@@ -1,0 +1,210 @@
+# Extension API Boundaries
+
+Homeboy talks about "the extension API" as one thing. It is three, and only one
+of them is designed. This document names all three, assigns every
+`ExtensionManifest` field to one, and states the rule each boundary should
+follow.
+
+It is a map, not a plan. Nothing here proposes moving code.
+
+## Measured surface
+
+Taken from `4560bcaaf`:
+
+| | |
+|---|---|
+| `ExtensionManifest` fields | 48 |
+| `ExtensionManifest` accessors | 50 |
+| `homeboy-extension-contract` | 54 modules, 7,334 lines |
+| `ExtensionCapability` variants | 8 |
+| `ExtensionRunner` call sites | 18 |
+| Shell helpers in `homeboy-extension/src/runtime/` | 14 |
+| Crates referencing `ExtensionManifest` outside the contract crate | 9, across 66 files |
+
+The ratio that matters: **8 of 47 classified fields are reached through the
+designed capability contract. 27 are read directly by named subsystems.**
+
+## The three boundaries
+
+### B1 — Executable capability
+
+An extension declares a script; Homeboy runs it with resolved settings, a run
+dir, the env-provider chain, invocation requirements, structured sidecars, and
+redaction.
+
+This boundary is designed. `ExtensionCapability::descriptor` centralizes label,
+manifest-support probe, and script accessor, so adding a capability is one arm
+rather than parallel edits across every consumer. Dispatch is generic:
+`has_manifest_support` and `script_path` are called through the enum, not
+through field access. `ExtensionCapability::Trace` and `::Audit` have **no**
+direct field reads outside the contract crate at all — the descriptor is their
+only route in. That is the encapsulation working.
+
+Two shapes exist within it:
+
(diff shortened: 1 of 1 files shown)


chubes4:
Closing unmerged. The document's sharpest claim does not survive verification.

It framed the shell runtime as an undesigned boundary with no supported way for an extension to locate Homeboy's helpers. That is wrong. `ensure_all_helpers()` (`crates/homeboy-extension/src/execution.rs:914`) materializes all 14 runtime helpers and exports an absolute path for each — `HOMEBOY_RUNTIME_SETTINGS_HELPER`, `HOMEBOY_RUNTIME_RUNNER_PRELUDE`, `HOMEBOY_RUNTIME_EMIT_LINT_FINDING`, and the rest — into the environment of every extension script Homeboy runs. The discovery contract exists and works.

That collapses the pain cited as evidence. Extra-Chill/homeboy-extensions#2508 and #2505 are not Homeboy boundary problems; they are the extensions repo vendoring copies instead of using the contract Homeboy already provides. The fix is a deletion in that repo, not a design decision here.

What remains genuinely unversioned is only the helpers' calling conventions, which is a far smaller question than this document claimed and does not currently justify an architecture doc.

The B1/B2 field map is still accurate and reproducible if it is ever wanted; it is preserved in this branch's history rather than merged, because a map whose headline finding is wrong is worse than no map.

## AI assistance

OpenCode 1.18.23 running Anthropic `claude-opus-5` wrote this document, then found the contradicting evidence while looking for the simplest available fix, and recommended closing its own PR. Chris Huber directed the work and made the call.

Issue #14362, opened 2026-09-07
Test isolation serializes the suite through a process-global env lock


chubes4:
## Problem

`with_isolated_home` implements isolation by mutating process-global environment variables (`HOME`, `XDG_CONFIG_HOME`, `XDG_DATA_HOME`, `HOMEBOY_DATA_DIR`, and ~15 more). Because those are process-wide, `HomeGuard::new()` must take a global mutex for the whole test body:

```rust
let guard = home_lock().lock().unwrap_or_else(|e| e.into_inner());
```

Every test using it therefore runs **strictly serially**, regardless of `--test-threads`.

## Impact

| crate | isolation calls | `#[test]` fns | observed wall time |
| --- | --- | --- | --- |
| `homeboy-release` | 61 | 574 | ~60s |
| `homeboy-deploy` | 39 | 330 | ~30s |
| `homeboy-lab-runner` | 430 | 2018 | **does not finish in 55 min** |
| `homeboy-agents` | 1003 | 2278 | ~7 min |
| `homeboy-core` | 791 | 3288 | ~25 min |

Cost does not track test count — it tracks *work performed while the lock is held*. `homeboy-lab-runner` is the extreme case: its 430 locked tests perform 149 process spawns and hundreds of real `git clone` / `fetch` / `commit` calls, all funnelled through one mutex.

Heaviest offenders, scored as locked tests × subprocess/git work:

```
score  iso  spawn  gitop  file
  740   20      6     30  workspace/tests/snapshot.rs
  552    8     43     25  homeboy_refresh/tests/part_a.rs
  323   17     18      0  workspace/tests/prune.rs
  288   18     12      3  homeboy_refresh/tests/part_b.rs
  270    9      1     28  lab_workspaces/tests/provider_config.rs
```

## Ruled out

- **Sleeps**: total declared test-side sleep across `homeboy-lab-runner` is **3.2 seconds**. Not the cause.
- **Redundant/obsolete tests**: the count is not the driver — `homeboy-deploy` runs 330 tests in 30s.
- **`slow-tests` feature**: exists and is documented, but is referenced **zero** times in `homeboy-lab-runner`.

## Consequence

Nobody runs `homeboy-lab-runner` locally, so CI is the only place it is ever exercised. That removes the fast feedback loop precisely where the runner/transport code is most timing-sensitive, and it compounds the flakiness already tracked in #12914.

## Direction

[...]

PR #14408, opened 2026-09-07
test(lab-runner): root the refresh path so its tests leave the global home lock

chubes4:
## Summary
Roots the whole `homeboy_refresh` execution path — registry, status, session, exec, reconnect and promotion — so its tests isolate through injected `PathRoots` instead of the process-global home lock. Third slice of #14362, after #14367 (snapshot) and #14373 (prune).

## Measured (`homeboy_refresh`, 3 runs each)
| config | baseline `main` | this branch |
|---|---|---|
| default | 98.90s | **87.30 / 87.30 / 87.40s** |
| short probe timeouts | 23.08s | **12.34 / 14.52s** |

100 passed / 1 failed on both sides across three runs — the one failure is the pre-existing `part_a::materialize_failure_preserves_compiler_diagnostics_and_active_binary`. Parity also confirmed against baseline on connection (235/4), execution (243/1), workspace (355/4), lab_workspaces (44/2), generation_store (35/0).

## The finding that matters more than this PR
**`homeboy_refresh` is not slow because of the home lock.** A single test, `part_b::stale_session_refresh_blocker_starts_the_newly_selected_binary`, takes **87.47s on unmodified `main`** — essentially the entire module.

It is waiting on production timeouts, not on the lock. On stock `main`, with no code change at all:

```
HOMEBOY_READONLY_PROBE_TIMEOUT_SECONDS=1 HOMEBOY_RUNNER_PROBE_WAIT_SECONDS=1
```

takes that test from **87.47s to 10.92s**, and the module from **101.10s to 23.08s**.

So the two levers compound: capping probe timeouts is worth ~77% here, and rooting the tests is worth a further ~40% on top of that (23.08s to ~13s). The timeout lever is free and belongs in its own change; I did not include it here because it is a test-harness policy decision, not a refactor. It did **not** reproduce on workspace (153.86s vs 145.32s), connection (39.08 vs 37.08) or execution (42.77 vs 43.53) — this is specific to the refresh path.

## Ambient leaks fixed (correctness, not speed)
Rooted entry points were silently falling back to ambient resolution and discarding injected roots. The worst one: **refresh tests were reading the host machine's real promotion lease store**, so they failed with `RuntimePromotionContended ... operation "controller upgrade"` whenever the developer machine held a live lease. Now rooted:

- `create_in_roots`, `merge_in_roots`
- `refresh_homeboy_binary`, `plan_homeboy_binary_refresh`, `plan_runner_dev_sync`, `promote_verified_runner_binary` (now `with_config_lock_at`), `refresh_promotion_authorities`, `sync_extension_overlays`, `restore_runner_homeboy_path_if_selected`, `defer_reconnect_after_promotion_race`, `probe_reconnected_admission_readiness`
- `session_path`, `read_session_for_controller`, `read_session_for_status_until`, `read_session_or_live_peer`, `status`, `reconcile_status`, `status_for_admission`, `connect`, `disconnect_with_session`, `rotate_daemon_generation`, `active_jobs_before_daemon_replacement`
- `exec_with_status_snapshot` — it hardcoded `runner: None`, forcing `prepare_runner_process` to re-resolve ambiently
- `runtime_promotion::acquire_in_root` / `acquire_waiting_for_compatible_key_in_root`

Six ambient wrappers became dead once production stopped calling them and were deleted; their remaining callers were tests, which now pass roots explicitly.

## Not converted
`blocked_connect_preserves_successful_promotion_with_one_continuation` converted cleanly and passes in isolation, but flaked roughly 1 run in 3 under parallelism — it asserts an exact phase sequence that shifts with connect timing. Left on `with_isolated_home` rather than shipped flaky. Four SSH-bootstrap tests stay on the lock because `merge_server_runner` and the server registry are still ambient.

## AI Assistance
OpenAI GPT-5.6 Sol through OpenCode did the rooting across all four layers, used a temporary backtrace probe on `load_in_roots` to locate each ambient escape, and measured every claim against a baseline worktree. Chris Huber directed the work and remains responsible for acceptance.

### DIFF
--- a/crates/homeboy-core/src/runtime_promotion.rs
+++ b/crates/homeboy-core/src/runtime_promotion.rs
@@ -420,7 +420,21 @@ fn selection_blocker_after_busy(
 /// and generation. A child process must present the capability explicitly
 /// attached by [`RuntimePromotionLease::authorize_subprocess`].
 pub fn acquire(operation: &str, target: impl Into<String>) -> Result<RuntimePromotionLease> {
+    acquire_in_root(&paths::runtime_promotion_dir()?, operation, target)
+}
+
+/// [`acquire`] below an already-resolved runtime promotion root.
+///
+/// The lease store is machine-global by default. An explicitly rooted acquire
+/// keeps an isolated caller from observing or contending the host's live
+/// promotion lease (#14362).
+pub fn acquire_in_root(
+    root: &Path,
+    operation: &str,
+    target: impl Into<String>,
+) -> Result<RuntimePromotionLease> {
     acquire_with_pin_policy(
+        root,
         operation,
         target.into(),
         String::new(),
@@ -455,6 +469,7 @@ pub fn acquire_waiting_for_compatible_with_status(
     progress: impl FnMut(RuntimePromotionWaitEvent),
 ) -> Result<RuntimePromotionLease> {
     acquire_waiting_for_compatible_key_and_status(
+        &paths::runtime_promotion_dir()?,
         operation,
         target,
         "",
@@ -475,6 +490,7 @@ pub fn acquire_waiting_for_target_with_status(
     progress: impl FnMut(RuntimePromotionWaitEvent),
 ) -> Result<RuntimePromotionLease> {
     acquire_waiting_for_compatible_key_and_status(
+        &paths::runtime_promotion_dir()?,
         operation,
         target,
         "",
@@ -493,7 +509,27 @@ pub fn acquire_waiting_for_compatible_key(
     timeout: Duration,
     progress: impl FnMut(RuntimePromotionWaitEvent),
 ) -> Result<RuntimePromotionLease> {
+    acquire_waiting_for_compatible_key_in_root(
+        &paths::runtime_promotion_dir()?,
+        operation,
+        target,
+        compatibility_key,
+        timeout,
+        progress,
+    )
+}
+
+/// [`acquire_waiting_for_compatible_key`] below an already-resolved root.
+pub fn acquire_waiting_for_compatible_key_in_root(
+    root: &Path,
+    operation: &str,
+    target: impl Into<String>,
+    compatibility_key: impl Into<String>,
+    timeout: Duration,
+    progress: impl FnMut(RuntimePromotionWaitEvent),
+) -> Result<RuntimePromotionLease> {
     acquire_waiting_for_compatible_key_and_status(
+        root,
         operation,
         target,
         compatibility_key,
@@ -511,6 +547,7 @@ enum CompatibleOwnerPolicy {
 }
 
 fn acquire_waiting_for_compatible_key_and_status(
+    root: &Path,
     operation: &str,
     target: impl Into<String>,
     compatibility_key: impl Into<String>,
@@ -536,6 +573,7 @@ fn acquire_waiting_for_compatible_key_and_status(
             last_progress: &mut last_progress,
         };
         match acquire_with_pin_policy(
+            root,
             operation,
             target.clone(),
             compatibility_key.clone(),
@@ -581,7 +619,17 @@ pub fn acquire_for_generation_rotation(
     operation: &str,
     target: impl Into<String>,
 ) -> Result<RuntimePromotionLease> {
+    acquire_for_generation_rotation_in_root(&paths::runtime_promotion_dir()?, operation, target)
+}
+
+/// [`acquire_for_generation_rotation`] below an already-resolved root.
+pub fn acquire_for_generation_rotation_in_root(
+    root: &Path,
+    operation: &str,
+    target: impl Into<String>,
+) -> Result<RuntimePromotionLease> {
     acquire_with_pin_policy(
+        root,
         operation,
         target.into(),
         String::new(),
@@ -635,6 +683,7 @@ impl BoundedAdmission<'_> {
 }
 
 fn acquire_with_pin_policy(
+    root: &Path,
     operation: &str,
     target: String,
     compatibility_key: String,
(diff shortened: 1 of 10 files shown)


chubes4:
## CI fix pushed

The differential gate flagged 5 candidate-only failures. One was genuinely mine:

`homeboy_refresh::tests::part_b::extension_only_dev_sync_plan_does_not_refresh_homeboy_binary`

It seeded the runner into the injected root with `create_in_roots`, then read it back through the **ambient** `plan_runner_dev_sync`. Fixed to use `plan_runner_dev_sync_in_roots`.

### Why my local runs missed it
It passed locally six times in a row for a bad reason. Converted tests read the ambient `HOME` while unconverted tests mutate it process-wide through `HomeGuard`, and several `part_b` tests create a runner with the same `lab-local` id inside their isolated home. My converted test was reading another test's temporary home and finding a `lab-local` there by pure scheduling luck. CI scheduled it differently and it failed.

That is worth recording as a hazard of this migration: while a module contains both rooted and `with_isolated_home` tests, any ambient read in a rooted test is nondeterministic rather than reliably wrong, so it can pass locally and fail in CI.

### Verified properly this time
Reproduced the CI condition with a clean ambient home:

```
HOME=$(mktemp -d) cargo test -p homeboy-lab-runner --lib -- part_b::extension_only_dev_sync_plan
```

- before the fix: `RunnerNotFound { id: "lab-local" }` — matches CI exactly
- after the fix: passes in 0.92s

Full module under a clean home, twice: **100 passed / 1 failed** (the pre-existing `part_a::materialize_failure_preserves_compiler_diagnostics_and_active_binary`).

I also audited every converted test for ambient entry points; this was the only one, and `part_a`/`part_c` contain no converted tests.

Other modules under a clean home, branch vs `main`:

| module | branch | `main` |
|---|---|---|
| connection | 235/4 | 235/4 |
| execution | 242/2 | 241/3 |
| workspace | 355/4 | 355/4 |
| lab_workspaces | 44/2 | 44/2 |

`execution` is one better than baseline, not worse.

### The other 4
`agent_task_provider::runtime_readiness::readiness_command_timeout_is_capped_to_the_absolute_deadline`, `retryable_non_transient_verdicts_are_not_immediately_retried`, `commands::agent_task::run::preview_tests::backend_preview_readiness_injects_sole_default_fallback_credentials`, and `upgrade::execution::tests::part_a::installer_timeout_kills_setsid_descendant_before_delayed_mutation`.

All four are timeout, readiness or process-supervision tests in crates this PR does not touch beyond the additive `runtime_promotion` rooted variants, and the baseline run was already carrying 62 failures. I expect them to be load-sensitive flakes rather than regressions, but I am treating that as unproven until this run reports — if any of them reproduce I will chase them rather than wave them off.

---
*Diagnosis and fix by OpenAI GPT-5.6 Sol via OpenCode, directed by Chris Huber.*

chubes4:
## Rebased onto current `main`, plus a real regression fixed

### The differential gate was mostly noise
Comparing the two CI runs:

| | run 1 | run 2 |
|---|---|---|
| baseline failures | 62 | 67 |
| candidate-only | 5 | 3 |

**The baseline gained 5 new failures between runs with no input from this PR.** Three of the five I was originally blamed for — `readiness_command_timeout_is_capped_to_the_absolute_deadline`, `backend_preview_readiness_injects_sole_default_fallback_credentials`, `installer_timeout_kills_setsid_descendant_before_delayed_mutation` — are now failing on **baseline** in run 2. They were flakes, not regressions.

Only `agent_task_provider::runtime_readiness::tests::retryable_non_transient_verdicts_are_not_immediately_retried` was candidate-only in both runs. It cannot be caused by this PR: the diff touches `homeboy-agents` by **zero lines**, touches `homeboy-core` only in `runtime_promotion.rs` (+57/−6, additive), and `runtime_readiness.rs` references `runtime_promotion`, `PathRoots` and `paths::` exactly **0 times**. The test spawns a Node subprocess and asserts a probe ran once — load-sensitive by construction.

### One genuine regression, found and fixed
Rebasing surfaced a real one that CI had not yet reported: `part_b::ssh_source_snapshot_plan_builds_natively_without_cross_compilation` failed **3 out of 3** on this branch and **0 out of 3** on `main`. Deterministic, so mine.

```
tar: could not chdir to '/tmp/hb-test-33082-2ZtMQ0/0b172cfb61.t/source'
```

`build_runner_source_snapshot` stages through `tempfile::tempdir()`, which honours the process-global `TMPDIR`. `HomeGuard` repoints `TMPDIR` at its own `hb-test-*` root and deletes that tree on drop. The test is not one I converted — but once its neighbours stopped serialising on the home lock, a guard could retire the staging root mid-archive. The test now holds the isolation guard so its `TMPDIR` is stable.

### Worth recording for the rest of #14362
This is the second instance of the same hazard, after the `plan_runner_dev_sync` one. `HomeGuard` mutates process-global state — `HOME`, `TMPDIR` and roughly fifteen others — so **partially** converting a test binary is unstable in a specific way: unconverted tests that read ambient state or allocate through `TMPDIR` become less safe as their neighbours stop serialising. Both failures passed locally and in isolation, and only appeared under real concurrency.

The implication for the remaining conversions is that a module should be taken to completion rather than left half-rooted, and that anything reading ambient state needs auditing before its neighbours are freed — not after.

### Verified against the new baseline
Rebased onto `main` (one conflict in `connection.rs`, where `status_until` landed beside `status_in_roots`; kept both). Clean ambient home, 3 runs each:

| | result | time |
|---|---|---|
| this branch | **101 passed / 1 failed** ×3 | 87.04 / 88.60 / 87.35s |
| `main` | 101 passed / 1 failed | 97.64s |

The single failure is the pre-existing `part_a::materialize_failure_preserves_compiler_diagnostics_and_active_binary` on both sides.

---
*Rebase, diagnosis and fixes by OpenAI GPT-5.6 Sol via OpenCode, directed by Chris Huber.*

Issue #14532, opened 2026-09-09
main does not compile: ControlPlaneBlocker initializer missed by #14492 (orchestration.rs pre_execution_failure branch)


chubes4:
## Summary

`cargo build -p homeboy-cli` fails on `main` @ `ab2683db6`:

```
error[E0063]: missing fields `reason`, `retry` and `state` in initializer of `ControlPlaneBlocker`
  --> crates/homeboy-agents/src/orchestration.rs:3905:25
```

Commit `144f3bf9f` ("fix(agent-task): expose detached staging failures (#14492)") added `state`, `reason`, and `retry` to `ControlPlaneBlocker` (`crates/contracts/homeboy-control-plane-contract/src/resource.rs:659`) and updated the other initializers in `derive_audit_blocker()` — but missed the `pre_execution_failure` branch at `crates/homeboy-agents/src/orchestration.rs:3905`. The binary shipped as `0.369.0+53a977a4a6c0` predates this; current `main` cannot build the CLI.

## Fix

Add `state: None, reason: None, retry: None` to that initializer, mirroring the sibling `controller_failure` initializer directly above it.

_Filed by Extra Chill Bot (AI agent) after hitting the failure while building a fix branch for #14527; encountered on clean `main` checkout, unrelated to that change._

chubes4:
Fixed on main. The `pre_execution_failure` branch in `crates/homeboy-agents/src/orchestration.rs` now constructs `ControlPlaneBlocker` with all fields initialized (`state: None, reason: None, retry: None`) — see `orchestration.rs:4350-4360` on `origin/main`. Landed in `ab044d3d0` (#14505). Main compiles.
---
*AI assistance disclosure: this triage was performed by Claude (Anthropic) running in OpenCode, driven by Chris Huber. The model cross-referenced the open-issue set against `origin/main` git history and current source, and every fix claim above was verified against a named commit or a `file:line` in the tree before this issue was closed.*

Issue #1134, opened 2026-04-17
False positive: namespace_mismatch on files with PHP reserved words in namespace segments


chubes4:
## Bug

The `namespace_mismatch` audit check produces a false positive for PHP files that use reserved words as namespace segments. PHP 7.0+ allows reserved words in namespace declarations, but the auditor doesn't recognize them.

## Reproduction

`inc/Engine/AI/Tools/Global/AgentMemory.php` (in data-machine)

```php
namespace DataMachine\Engine\AI\Tools\Global;
```

This is valid PHP 7.0+. The `Global` segment is a reserved word but is legal in namespace context per [RFC: Context Sensitive Lexer](https://wiki.php.net/rfc/context_sensitive_lexer).

However, the auditor reports:

> Missing namespace declaration (expected `DataMachine\Engine\AI\Tools\Global`)

The fingerprint step sets `fp.namespace = None` for this file, likely because the parser strips or rejects the `Global` segment.

## Root Cause

In `src/core/code_audit/core_fingerprint.rs`:

1. **Line 96**: `"global"` is in `PHP_KEYWORDS` — used for structural normalization. This is fine on its own.
2. **`extract_namespace()`** at line 970-975 relies on symbols with `concept == "namespace"`. The underlying parser (tree-sitter or regex-based) likely fails to extract the full namespace name when a segment is a reserved word.
3. `conventions.rs` line 467 then sees `fp.namespace.is_none()` and flags it as "Missing namespace declaration".

## Expected

The auditor should parse `namespace DataMachine\Engine\AI\Tools\Global;` correctly and match it against the expected namespace derived from the file path.

## Affected

Every CI run re-files this as a new finding on Extra-Chill/data-machine#682.

PR #14536, opened 2026-09-10
fix(code-audit): block-comment openers take precedence over doc line prefixes

### LINKED ISSUE #14527: code_audit: namespace/method detectors emit false positives on main-branch full-tree audits, blocking the release gate (and PR-scoped audits can't see it)
## Summary

`homeboy review audit` fails on `data-machine-events` `main` with 7 drift sections. **At least 5 of 7 sampled deviations are verifiably false** — the analyzer reports missing namespaces and missing methods that are plainly present in the source.

Because the audit gate is part of the CI quality set, this blocks the release job: `data-machine-events` `main` carries the completed phpstan paydown (#767 / PR #791, baseline 556 → 0) but cannot cut a release.

## Verified false positives

Repo: `Extra-Chill/data-machine-events` @ `1ba8800`. Every claim below checked against the file on disk.

| Audit deviation | Reality |
|---|---|
| `inc/Abilities/AbilityCategories.php` — "Missing namespace declaration (expected `DataMachineEvents\Abilities`)" | line 12: `namespace DataMachineEvents\Abilities;` — exactly the expected value |
| `inc/Cli/Check/CheckMissingVenueAddressesCommand.php` — "Missing namespace declaration" | line 49: `namespace DataMachineEvents\Cli\Check;` |
| `inc/Cli/Check/CheckDuplicatesCommand.php` — "Missing method: `__invoke`" | line 79: `public function __invoke( array $args, array $assoc_args ): void` |
| `inc/Steps/.../Extractors/VenuePilotExtractor.php` — "Missing method: `canExtract`" | line 105: `public function canExtract( string $html ): bool` |
| `inc/Steps/.../Paginators/JsonApiPaginator.php` — "Missing method: `getNextPageUrl`" | line 48: `public function getNextPageUrl( string $current_url, string $content ): ?string` |

Two further sampled items (`inc/Cli/SettingsCommand.php` → `__invoke`, `inc/Blocks/Calendar/Cache/CacheInvalidator.php` → `invalidate`) genuinely lack the named method and may be legitimate convention findings.

**Suspected pattern for the namespace cases:** both files declare `namespace` after a long file docblock (line 12 and line 49 respectively). Neither uses a PHP reserved word, so this is *not* #1134 (closed, same symptom, different cause) — but `crates/homeboy-code-audit/src/core_fingerprint/tests.rs:744` already documents a related regression class where the grammar's namespace pattern misses declarations that are not in the expected position. The grammar pattern is `r"namespace\s+([\w\\]+);"` (`crates/homeboy-engine-primitives/src/grammar/mod.rs:178`); the consuming logic is `extract_namespace()` (`core_fingerprint/mod.rs:599`) and the deviation is raised in `conventions/mod.rs:311-330` under `if fp.namespace.is_none()`. A fingerprint that fails to populate `namespace` is indistinguishable from a file that truly has none.

Note the audit also runs through an **extension-script fingerprint path** (`crates/homeboy-code-audit/src/fingerprint.rs:195-215`, "Extension-script fingerprinting does not distinguish test methods") separate from `core_fingerprint`. Worth confirming which path produced these fingerprints before fixing, since only one may be affected.

## Second, independent problem: PR audits and main audits analyze different scopes

Same commit content, opposite results:

| Run | Ref | Units analyzed | Drift sections | Result |
|---|---|---|---|---|
| [34366857374](https://github.com/Extra-Chill/data-machine-events/actions/runs/34366857374) | PR #791 head `a333d07` | **8** | 0 | audit **pass** |
| [34380949853](https://github.com/Extra-Chill/data-machine-events/actions/runs/34380949853) | `main` `1ba8800` (squash of that PR) | **62** | 7 | audit **fail** |

PR audits are scoped to changed files; `main` audits run full-tree. A contributor therefore gets a green audit on the PR and cannot observe the failure their merge will produce, and the failure is not attributable to their change. (Confirmed here: `main` audit was already failing at `1600122` 11:17 UTC and `f2da9fe` 11:42 UTC, hours before #791 merged at 17:07 UTC — so the drift is pre-existing and the merge inherited it.)

Whatever the fix for the false positives, the gate that blocks releases should be the gate contributors can see.

## Reproduction

```
homeboy review audit data-machine-events --placement local
```
on `data-machine-events` `main` @ `1ba8800` reproduces CI exactly: `status: failed`, 7 drift sections, all five false positives above present. Homeboy `0.369.0+53a977a4a6c0`, wordpress extension `3.45.8`.

## Impact

- `data-machine-events` cannot release; the phpstan-zero paydown is stranded on `main`.
- Every heuristic false positive erodes trust in the audit gate and pushes operators toward `--skip-checks`, which is the failure mode #12826 and #13437 are already circling.

## Related

- #1134 — closed; same user-visible symptom (`Missing namespace declaration (expected X)`), different root cause (reserved-word segments). This is a distinct cause and arguably a regression of that class.
- #12826 — audit gate failing on bulk findings, triage/baseline needed.
- #13437 — differential gate rejecting below the suite's own noise floor (same "differential comparison is not trustworthy" family).
- #6857 — detectors inflating findings by counting comments / test-only modules.

_Filed by Extra Chill Bot (AI agent) on behalf of @chubes, with local reproduction and per-file verification._

chubes4:
Closes #14527

## Root cause

`walk_lines()` in the grammar engine checked doc/line-comment prefixes before block-comment openers (`crates/homeboy-engine-primitives/src/grammar/parser.rs`): a line beginning with `/**` matched the grammar's `doc = ["/**"]` entry in `is_line_comment()` and was consumed as a single-line comment, so the block-comment state never opened. Every subsequent docblock interior line (` * Doesn't match X`) was classified as `Region::Code`, where the walker's multi-line string tracking saw prose apostrophes as an unterminated `'` string and swallowed the following real code — including `namespace` declarations and method signatures. `fingerprint_from_grammar()` therefore produced fingerprints with `namespace: None` and the affected methods missing, and `check_conventions()` (`crates/homeboy-code-audit/src/conventions/mod.rs:321`, `:256-270`) reported them as `NamespaceMismatch` / `MissingMethod` deviations. The PHP grammar itself, the extension fingerprint script, and the comparison logic were all correct — the loss happened in region classification, upstream of all of them.

The five flagged files are reproduced exactly by fingerprinting the real sources with the installed WordPress grammar (`inc/Abilities/AbilityCategories.php` → `namespace=None`; `inc/Cli/Check/CheckDuplicatesCommand.php` → `methods=[]`, etc.), and all five extract correctly after the one-line precedence fix. Files whose docblocks contained no apostrophe (e.g. `CheckMissingVenueAddressesCommand` namespace case) failed through a second-order effect: an earlier phantom string, once opened by an apostrophe elsewhere in the file, swallowed the declaration line.

## Producer/fallback behavior

- **Before:** grammar path (extension `grammar.toml`) is preferred; the extension fingerprint script is a fallback used only when no grammar handles the extension. The fallback was not the problem — the grammar path itself returned a *silently wrong* fingerprint (computed, but missing members), which the convention comparison cannot distinguish from a genuine violation.
- **After:** same producer selection, but the grammar path now classifies docblocks correctly, so its output is accurate. No suppression lists, no finding-kind exclusions, no gate changes.

Per the principle from the issue: **a fingerprint that could not be computed must never be reported as a convention deviation.** That contract already holds structurally — a file whose fingerprint cannot be computed at all returns `None` from `fingerprint_content()` and is dropped before `discover_conventions()` (pinned by the existing `fingerprint_extension_content_returns_none_without_claimed_extension` test and the new `namespace_conforming_file_not_reported` test) — and this fix removes the remaining way a *computed* fingerprint could be silently wrong: parser misclassification. "Analyzer failed" and "code violates convention" are now genuinely distinct states.

## Audit before/after (`homeboy review audit data-machine-events --placement local`, component `1ba8800`)

Five false positives — **before → after**:

| Audit deviation | Before | After |
|---|---|---|
| `inc/Abilities/AbilityCategories.php` — Missing namespace declaration (expected `DataMachineEvents\Abilities`) | reported | **gone** |
| `inc/Cli/Check/CheckMissingVenueAddressesCommand.php` — Missing namespace declaration | reported | **gone** |
| `inc/Cli/Check/CheckDuplicatesCommand.php` — Missing method `__invoke` | reported | **gone** |
| `inc/Steps/.../Extractors/VenuePilotExtractor.php` — Missing method `canExtract` | reported | **gone** |
| `inc/Steps/.../Paginators/JsonApiPaginator.php` — Missing method `getNextPageUrl` | reported | **gone** |

Three additional same-cause false positives (verified present in source, swallowed by the same phantom-string mechanism) also cleared: `CheckOrphanPipelinesCommand.php` (`__invoke`), `UpdateEventCommand.php` (`__invoke`), `DiceFmAuth.php` (`__construct`). Total deviations dropped 18 → 10.

True findings survive — no false-negative trade:

- `inc/Cli/SettingsCommand.php` — Missing method `__invoke`: still reported (file has no `__invoke`)
- `inc/Blocks/Calendar/Cache/CacheInvalidator.php` — Missing method `invalidate`: still reported (file has no `invalidate`)
- All non-fingerprint-heuristic findings unchanged: `EventScraperTest`/`mapEvent`, `DetailPageFetcher` (`enrich`/`shouldEnrich`/`provides`), and the three `naming_mismatch` findings.

Run status remains `failed` because genuine convention drift remains to triage — this PR fixes the detector, not the gate.

## Gates

- `cargo fmt --all --check` — pass
- `cargo clippy -p homeboy-engine-primitives -p homeboy-code-audit --all-targets` — zero new warnings vs clean `main` baseline (74/74 and 0/0); note raw `clippy -- -D warnings` also fails on a pre-existing `large_enum_variant` in `homeboy-runner-contract` (dependency crate, untouched here)
- `cargo test -p homeboy-engine-primitives -p homeboy-code-audit` — all green except two pre-existing, environment-dependent failures verified on a clean checkout of `main` (`command::tests::controller_loss_reaps_the_release_mutation_process_group`, `detectors::artifact_portability::tests::homeboy_config_declares_homeboy_run_marker`). `cargo-nextest` is not installed on this host; `cargo test` used per the issue's fallback clause.
- Full audit repro re-run: results above.

## Related regressions

- **#1134 (closed, reserved-word namespace segments):** covered — `namespace_with_php_reserved_word_segment_is_extracted` still passes; this fix is orthogonal (region classification, not the namespace regex).
- **Docblock/indented-namespace class (`core_fingerprint/tests.rs:744+`):** covered and extended — `namespace_with_leading_whitespace_is_extracted` still passes, and the new tests (`namespace_after_long_docblock_is_extracted`, `method_after_docblock_with_apostrophes_is_extracted`, plus the engine-level region tests) pin the docblock-position variant that #14527 exposed.
- **Known remaining gap, distinct cause:** `TicketmasterTest.php` is still reported missing `test` because its signature spans multiple lines (`public function test(` with parameters on following lines) and grammar patterns match per line. That is a multi-line-signature extraction limitation, pre-existing and unrelated to docblock parsing — worth a separate issue if the false positive matters.

Also filed while building this branch: #14532 (clean `main` does not compile — `ControlPlaneBlocker` initializer missed by #14492 in `crates/homeboy-agents/src/orchestration.rs:3905`). This branch inherits that breakage; the audit verification was run with a local, uncommitted three-line unblock of that initializer (not included in this PR).

## AI disclosure

This PR was authored by an AI coding agent (Extra Chill Bot minion) and is pending human review.

### DIFF
--- a/crates/homeboy-code-audit/src/conventions/tests.rs
+++ b/crates/homeboy-code-audit/src/conventions/tests.rs
@@ -1098,6 +1098,64 @@ fn namespace_mismatch_detected_in_convention() {
 }
 
 #[test]
+fn namespace_conforming_file_not_reported() {
+    // Regression test companion for #14527.
+    //
+    // Inverse of `namespace_mismatch_detected_in_convention`: a file whose
+    // fingerprint carries exactly the expected namespace (as the fixed
+    // producer now yields for a file declaring its namespace after a long
+    // docblock) must be conforming — never reported as "Missing namespace
+    // declaration". A convention deviation may only be synthesized from an
+    // accurate fingerprint; a file the producer could not fingerprint at all
+    // is dropped upstream (`fingerprint_content` returns None) and never
+    // reaches this comparison.
+    let fingerprints = vec![
+        FileFingerprint {
+            relative_path: "abilities/CreateFlow.php".to_string(),
+            language: Language::Php,
+            methods: vec!["execute".to_string()],
+            type_name: Some("CreateFlow".to_string()),
+            namespace: Some("SamplePlugin\\Abilities\\Flow".to_string()),
+            ..Default::default()
+        },
+        FileFingerprint {
+            relative_path: "abilities/UpdateFlow.php".to_string(),
+            language: Language::Php,
+            methods: vec!["execute".to_string()],
+            type_name: Some("UpdateFlow".to_string()),
+            namespace: Some("SamplePlugin\\Abilities\\Flow".to_string()),
+            ..Default::default()
+        },
+        FileFingerprint {
+            relative_path: "abilities/DeleteFlow.php".to_string(),
+            language: Language::Php,
+            methods: vec!["execute".to_string()],
+            type_name: Some("DeleteFlow".to_string()),
+            namespace: Some("SamplePlugin\\Abilities\\Flow".to_string()), // declares the expected namespace
+            ..Default::default()
+        },
+    ];
+
+    let convention = discover_conventions("Flow", "abilities/*", &fingerprints).unwrap();
+
+    assert_eq!(
+        convention.expected_namespace,
+        Some("SamplePlugin\\Abilities\\Flow".to_string())
+    );
+    assert!(
+        convention.outliers.is_empty(),
+        "A file declaring exactly the expected namespace must not produce deviations. Got: {:?}",
+        convention
+            .outliers
+            .iter()
+            .flat_map(|o| &o.deviations)
+            .map(|d| &d.description)
+            .collect::<Vec<_>>()
+    );
+    assert_eq!(convention.conforming.len(), 3);
+}
+
+#[test]
 fn missing_import_not_flagged_for_same_namespace_reference() {
     // Regression test for #1135 (case 2).
     //
--- a/crates/homeboy-code-audit/src/core_fingerprint/tests.rs
+++ b/crates/homeboy-code-audit/src/core_fingerprint/tests.rs
@@ -736,6 +736,66 @@ fn namespace_with_php_reserved_word_segment_is_extracted() {
 }
 
 #[test]
+fn namespace_after_long_docblock_is_extracted() {
+    // Regression test for #14527 (real-world case:
+    // data-machine-events inc/Abilities/AbilityCategories.php).
+    //
+    // The file declares exactly the expected namespace on line 12, after a
+    // long file docblock. The audit reported "Missing namespace declaration"
+    // because the grammar engine consumed the `/**` opener as a line comment
+    // (doc prefix), never opened the block state, and classified the
+    // docblock interior as code — where prose apostrophes opened phantom
+    // multi-line strings that swallowed the namespace line.
+    let Some(grammar) = php_grammar() else {
+        eprintln!("Skipping — wordpress grammar not available");
+        return;
+    };
+
+    let content = "<?php\n/**\n * Ability Categories\n *\n * Centralized registration of Data Machine Events ability categories.\n * Follows the same pattern as Data Machine core's AbilityCategories.\n *\n * @package DataMachineEvents\\Abilities\n * @since 0.29.0\n */\n\nnamespace DataMachineEvents\\Abilities;\n\ndefined( 'ABSPATH' ) || exit;\n\nclass AbilityCategories {\n    public static function register(): void {\n    }\n}\n";
+
+    let fp = fingerprint_from_grammar(content, &grammar, "inc/Abilities/AbilityCategories.php")
+        .expect("fingerprint should succeed");
+
+    assert_eq!(
+        fp.namespace.as_deref(),
+        Some("DataMachineEvents\\Abilities"),
+        "Namespace declared after a long docblock must be extracted. Got: {:?}",
+        fp.namespace
+    );
+}
+
+#[test]
+fn method_after_docblock_with_apostrophes_is_extracted() {
+    // Regression test for #14527 (real-world case: data-machine-events
+    // inc/Cli/Check/CheckDuplicatesCommand.php defines `__invoke` at line 79,
+    // yet the audit reported "Missing method: __invoke").
+    //
+    // WP-CLI command docblocks contain prose with apostrophes ("weren't",
+    // "doesn't"). When the docblock was misparsed as code, each apostrophe
+    // opened a phantom multi-line string that swallowed the method signature
+    // that followed — here the ENTIRE method list came back empty.
+    let Some(grammar) = php_grammar() else {
+        eprintln!("Skipping — wordpress grammar not available");
+        return;
+    };
+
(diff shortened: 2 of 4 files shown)


chubes4:
Rebased onto `origin/main` (26 commits). The previous `homeboy / Test` failure was not this change: the differential gate's inventory producer runs `cargo test --no-run --workspace`, which hit `error[E0063]: missing fields `reason`, `retry` and `state` in initializer of `ControlPlaneBlocker`` in `homeboy-agents` — a crate this PR does not touch, broken at the branch's stale base and since fixed on main (main is green at `5835105`). Candidate Test and Baseline Test both passed even then; only the workspace-wide inventory compile failed.

Worth noting the irony for triage: a PR fixing "analyzer failure reported as a code finding" was itself blocked by an analyzer-infrastructure failure surfaced as `invalid_evidence` rather than as "the workspace does not compile at this base". Same conflation, different gate — cf. #13437.

PR #14674, opened 2026-09-15
fix(daemon): survive startup when jobs outlive their generation

chubes4:
## Summary

The daemon rebuilt job-to-generation ownership on every start by replaying each
durable job's recorded lease through `record_job`, which treats an unregistered
generation as an invariant violation. Generations retire by design, so any job
store holding history from earlier daemons made that error **certain**: the
daemon aborted during startup and could never start again.

This rebuilds ownership only for jobs that still have routable work, and treats
a generation that is no longer registered as ordinary history rather than a
failure.

`record_job` stays strict for live controller admission, where the admitting
lease is the running daemon's own and an absent generation is a real defect.

## Impact

This wedges the daemon permanently, which takes every `agent-task cook` with it —
cook requires the daemon in both detached and foreground modes.

The failure also surfaced as a **misleading error**. Startup admission reported:

> daemon process 43651 did not publish its isolated startup token before timeout
> ... "The daemon startup was not accepted because the expected token and PID did not match."

Both claims were false. The token matched every attempt, and PID is deliberately
not compared — `observe_startup_lease` matches on token precisely because the
`supervise` launcher and its `serve` child have different PIDs. The real
condition was `admits_work()` staying false because the daemon had already
exited. Diagnosis cost far more than it should have; worth a follow-up so the
startup timeout reports the child's exit instead of blaming token/PID.

## Evidence

Reproduced against a real controller state directory holding 260 durable jobs
(221 succeeded, 38 failed, 1 cancelled) spread across dozens of historical
leases, with a registry containing exactly one generation — none of the jobs'
owners.

**Before** — dies immediately, every time:

```
DAEMON EXITED after 0.58s
summary: "local daemon job owner was not present in the generation registry"
```

**After** — same state directory, same binary path:

```
$ homeboy daemon start
"status": "succeeded", "pid": 97934, "lease_id": "60de9755-..."
elapsed 6s

$ homeboy daemon status
running = True   reachable = True   fresh = True
```

An `agent-task cook` then reached durable run persistence and provider execution
on that same daemon, which it could not do before.

## Simplification

The terminal-job path previously called `record_job` (insert owner, increment
`active_jobs`) and then immediately `mark_job_terminal` (insert into
`completed_jobs`, decrement). That round trip left a permanent `job_owners`
entry per historical job, growing the registry without bound, to describe work
with nothing left to route. Generation retirement already prunes exactly that
bookkeeping. Skipping terminal jobs removes the churn and the growth.

## Verification

- `cargo test -p homeboy-core --lib daemon::generation_store` — 7 passed, 0 failed,
  including the new `startup_rebuild_survives_jobs_owned_by_retired_generations`,
  which proves a retired generation does not fail recovery, that it owns nothing
  routable, that the live generation still admits, and that a job whose
  generation *is* registered is still rebuilt and routable.
- End-to-end daemon start and cook admission against the real 260-job store, above.

## Stacking

Based on `fix/cook-destination-resolution-simplify` (#14673) so this diff shows
only the daemon commit. The two changes are independent; #14673 can merge first.

## AI assistance

Claude Sonnet 4.5 via Kimaki/OpenCode diagnosed this by timing a manual daemon
start (proving it exits in 0.58s rather than being slow, which disproved an
earlier bloated-state hypothesis before anything was deleted), traced the abort
to `rebuild_generation_job_ownership`, quantified the real job store, wrote the
change and its regression test, and verified daemon start and cook admission end
to end. Chris Huber directed the work and retains merge authority.


chubes4:
Superseded by #14675. #14673 squash-merged, so this branch's base commit is no longer an ancestor of `main` and the PR showed both commits. #14675 carries the identical daemon change cherry-picked cleanly onto current `main`, and targets `main` from the start so the `pull_request: branches: [main]` CI trigger fires (this PR never ran checks because it was opened against a branch base).

AI assistance: Claude Sonnet 4.5 via Kimaki/OpenCode, under Chris Huber's direction.

PR #14689, opened 2026-09-15
fix(fanout): render cook-batch summary on successful outcomes

### LINKED ISSUE #14682: Structured output is unreliable: mixed streams, --output ignored on preview, empty summaries
## Problem

Reading any structured result from the CLI requires a custom parser. Three separate papercuts compound:

**1. Progress NDJSON and the final object share stdout.** Commands stream one JSON object per line, then print a pretty-printed result object. Neither `json.load(stdin)` nor a line-wise NDJSON reader works:

```
json.decoder.JSONDecodeError: Extra data: line 2 column 1 (char 142)
```

Every invocation needed code to locate the boundary (`txt.find('{\n')`) before parsing. That is a guess, not a contract.

**2. `--output <PATH>` is silently ignored on `--preview`.** The flag is documented and accepted, and no file appears:

```
$ homeboy agent-task fanout cook-batch ... --preview --output /tmp/fanout-preview.json
$ python3 -c "json.load(open('/tmp/fanout-preview.json'))"
FileNotFoundError: '/tmp/fanout-preview.json'
```

The flag's whole purpose is escaping problem 1, and it does not work on the command most likely to need it.

**3. `summary` is empty on success.** A successful plan returns `status: planned`, `success: true`, and `summary: ""`. The human-readable field carries nothing, so even a successful result must be extracted by walking nested `data`.

## Impact

Programmatic use of the control plane means reimplementing a parser per call site and hoping the boundary heuristic holds. `--output` is the designed escape hatch and it is unreliable, so there is no dependable machine-readable path. This is a direct incentive to bypass the CLI.

## Acceptance

- Progress events and the final result do not share one stream, or the final result is unambiguously delimited.
- `--output` writes the result file for every command that accepts it, including `--preview`.
- `summary` is populated on success as well as failure.
- A documented, stable way to obtain just the result object exists and is tested.

## AI assistance

Anthropic `claude-opus-5` via OpenCode consumed these outputs while orchestrating a batch and drafted this issue. Chris Huber directs the work.

chubes4:
## Root cause

Issue #14682 raised three claims about the structured-output control plane.
I verified each against the current code before touching anything:

1. **"Progress NDJSON and the final object share stdout."** Not reproducible.
   Every fanout/cook progress line uses `eprintln!` (stderr); the final
   pretty-printed envelope is the only thing written to stdout
   (`OutputService::emit_run` → `print_response`). Confirmed by running a real
   `agent-task fanout cook-batch --preview` invocation and inspecting stdout
   vs stderr separately — stdout contained exactly one JSON document.

2. **"`--output` is silently ignored on `--preview`."** Not reproducible for
   `fanout cook-batch`. The output-file write (`write_output_file`) is keyed
   only by `CommandOutputFileMode`, which is derived from the command/spec,
   not from `--preview`. The one place a `--preview` branch does skip a
   special-case output path is `agent-task cook` (not fanout), and that
   branch falls through to the same generic path that still writes
   `--output`. Confirmed empirically: `agent-task fanout cook-batch --repo
   fixture --preview --output <path>` wrote the file correctly (exit 2, repo
   unregistered, but the envelope was written to both stdout and the
   `--output` path).

3. **"`summary` is empty on success."** Reproduced and is the real bug. The
   top-level envelope `summary` field falls back through
   `summary_for_payload`, which tries `presentation.stdout` (the rendered
   human summary) before falling back to `data.summary`/`data.message` as
   strings. For `agent-task fanout cook-batch`,
   `render_fanout_cook_batch_summary` in
   `crates/homeboy-cli/src/commands/agent_task_summary/mod.rs` only produced
   text for `status == "blocked"` and (via
   `render_failed_fanout_child_summary`) `"failed"`/`"partial_failure"`.
   Every successful status — `"ready"` (a plan or `--preview`),
   `"completed"`, `"review_ready"`, `"running"`, etc. — fell through with no
   branch, so the renderer returned `None`, `presentation.stdout` stayed
   `None`, and `data.summary` is an object (not a string), so the remaining
   fallbacks in `summary_for_payload` also return `None`. Net effect: the
   envelope `summary` field is absent (serializes as `""` for callers that
   `unwrap_or_default()`) on every successful cook-batch call, while failures
   get a rich digest. I confirmed this directly with a unit-level probe
   (`cli_response_for_json_result_for_command` fed a real `"ready"` cook-batch
   payload) before writing the fix: `envelope.summary == None`.

## Fix

`crates/homeboy-cli/src/commands/agent_task_summary/mod.rs`: add
`render_fanout_cook_batch_progress_summary`, covering every status that
isn't `"blocked"` and isn't a failure. It renders a compact line from the
fields every cook-batch payload carries unconditionally regardless of
dry-run/materialized state: `fanout_id`, `summary.issues`,
`summary.worktrees_total`, `summary.worktrees_blocked`, `dry_run`, plus the
first `next_actions[].command` when present. This makes `presentation.stdout`
non-empty for successful outcomes, so the envelope `summary` field is
populated the same way it already is for failures.

Scope: this PR fixes the one concretely reproducible defect (#3). Claims #1
and #2 did not reproduce against current `main` and are left alone rather
than "fixed" speculatively.

## Verified

- `cargo fmt --check` — clean.
- `cargo check --all-targets -p homeboy-cli` — clean (only pre-existing,
  unrelated warnings in other crates).
- Before writing the fix, reproduced the bug at the unit level: built a real
  `"ready"` cook-batch payload via the existing
  `with_materialized_cook_batch_worktrees` test fixture, ran it through
  `cli_response_for_json_result_for_command`, and confirmed
  `envelope.summary == None`.
- Before writing the fix, reproduced the CLI-level stdout/stderr and
  `--output` claims directly against the built binary and found both already
  correct on `main` (see Root cause above).
- Added two focused unit tests next to the existing fanout summary tests
  (`ready_cook_batch_preview_summary_reports_the_plan`,
  `completed_cook_batch_run_summary_reports_the_outcome`) asserting the exact
  rendered summary text for a preview-ready payload and a materialized
  completed payload. Both assert equality against the full expected string,
  matching this test module's existing convention
  (`blocked_fanout_summary_...`, `failed_fanout_summary_...`).
- I did not run the full test/clippy suite locally per this machine's
  verification policy; GitHub Actions CI on this PR is the authoritative
  gate for `cargo test` and `cargo clippy --all-targets -- -D warnings`.

Fixes #14682

## AI assistance

Anthropic `claude-sonnet-5` via OpenCode implemented this fix. Chris Huber
directs the work.


chubes4:
Duplicate of #14688 (created by a retried automation step). Closing in favor of #14688.

Issue #14315, opened 2026-09-06
Cook status can report zero diff and success while a patch is applied and verification remains pending


chubes4:
## Summary

A local OpenCode Cook applied a substantive patch to its target worktree, but the terminal status projected contradictory state:

- `Status: succeeded`
- `Patch candidates: 0 non-empty / 0 empty`
- `Changed files: 0`
- `Diff bytes: 0`
- candidate state `verification_pending`
- a non-empty patch artifact was attached and the target worktree contained the changes

The provider took approximately 15 minutes and reserved its terminal result just before cancellation. Homeboy correctly deferred cancellation for terminal import, but status remained zero-diff after import. The Cook later entered remediation and correctly classified a required gate shared with the immutable baseline as `baseline_red`.

## Reproduction evidence

Cook: `cook-detached-44e52a81-edfb-4112-87d6-62fcebe99bbe`
Attempt: `cook-detached-44e52a81-edfb-4112-87d6-62fcebe99bbe-attempt-1-8c83a78d`
Repository tracker: https://github.com/Extra-Chill/wp-coding-agents/issues/580

After import, the patch artifact contained changes across three files and `git status` showed those files modified, while `agent-task status` still reported zero changed files/diff bytes and overall success with verification pending.

## Expected

- Terminal Cook success requires completed required verification.
- Applied non-empty patches project non-zero candidate/diff metadata.
- Status and finalization consume the same canonical candidate evidence as promotion.
- A cancellation/terminal-import race does not leave stale summary counters.

## Acceptance

A regression fixture covering provider terminal reservation followed by cancellation and import must project the canonical non-empty patch and remain non-success until required verification finishes.

---
AI assistance disclosure: GPT-5.6 Sol via OpenCode diagnosed this from Homeboy status, artifact, worktree, and gate evidence while orchestrating the affected Cook under Chris Huber direction.

chubes4:
Fresh concrete occurrence: run cook-detached-7ad137fe-ace6-44c7-942f-3cb936844062-attempt-1-07c95c9f reports top-level succeeded, candidate verification_pending, blocker IO error, while Cook terminal outcome is durable_failure. review reports apply_ready, verified:false, gates:[] and finalization not_attempted. This consumed repeated status/log/artifact reads to establish actual state. Coupled post-apply recovery is tracked in #14821. AI assistance: GPT-6 Astra via OpenCode; operator-directed audit.

chubes4:
[...]

PR #14828, opened 2026-09-21
fix(cook): preserve promotion failure causes

### LINKED ISSUE #14821: Regression: Cook loses post-apply I/O cause and continuation hangs without gate evidence
## Observed
Cook cook-detached-7ad137fe-ace6-44c7-942f-3cb936844062-attempt-1-07c95c9f produced and applied a substantive candidate, then returned only internal.io_error / IO error. Both status and diagnose drop the operation/path/OS cause. cook-continue --preflight admitted verification without provider replay, but cook-continue ran silently for 20 minutes until its caller timed out. No recorded gates or PR resulted. The candidate remains intact and verification_pending. Review calls it apply_ready while top-level status says succeeded; cancellation/retry guidance alternately calls it terminal. This drove manual artifact inspection and direct verification.

## Scope
Recover the exact underlying failure from retained evidence or reproduce its boundary; fix it upstream, not only the message. Preserve redacted operation, path, OS cause and causal phase through persistence/status/diagnose. Continue an already-applied candidate deterministically with bounded waits and heartbeat, reconcile interrupted continuation ownership and avoid replaying the provider or reapplying the patch. Distinguish provider success from overall Cook success.

## Related
Regression of closed #12154. Existing #14315 covers misleading candidate success; #14711 covers silent unbounded commands. This issue binds them to the concrete post-apply failure and must verify recovery of this candidate.

## Acceptance
Force a real post-apply filesystem/process I/O failure; exact cause survives durable status/diagnose and a corrected continuation finishes the same candidate once. Interrupt/restart the continuation and prove bounded recovery with no duplicate writer. Existing transport candidate must be recovered. Public-safe evidence only.

AI assistance: GPT-6 Astra via OpenCode observed and diagnosed the failure and filed this report under the operator’s instruction.

chubes4:
## Summary

- Preserve bounded structured promotion failures on the durable run record.
- Return the same causal diagnostic in failed promotion action results.
- Prefer the durable controller cause in Cook status and diagnose output.
- Add a synthetic post-apply recovery regression proving the candidate remains resumable.

Fixes #14821 and the misleading-success path tracked by #14315.

## Verification

- `cargo fmt --all`
- `CARGO_BUILD_JOBS=2 cargo check -p homeboy-agents`
- `CARGO_BUILD_JOBS=2 cargo test -p homeboy-agents post_apply_promotion_failure_keeps_cause_and_bounded_resume_action --lib`
- `CARGO_BUILD_JOBS=2 cargo test -p homeboy-agents agent_task_service::promotion_service::tests --lib`

The full affected library suite was bounded at ten minutes; two pre-existing unrelated tests independently failed and are documented in the session report.


chubes4:
Follow-up fixes the recovered underlying failure: deterministic-gate heartbeat persistence exceeded the control-plane event input bound. Heartbeat output now has a UTF-8-safe bounded tail before durable event serialization; the existing full gate artifact remains separate. Added a real durable-event append regression with 128 KiB output and corrected overall state projection while verification is pending. Resumed-gate and continuation/finalization regression tests passed without a second provider execution. AI assistance: GPT-5.6 Luna via OpenCode implemented/tested; GPT-6 Astra via OpenCode reviewed, preserved and published this verified follow-up under the authorized direct stabilization workflow.

Issue #14751, opened 2026-09-18
Release gate blocked: five tests depend on ambient host state, not fixtures they create


chubes4:
## Summary

The `Release` workflow has failed at the `Test` gate on every run since 2026-09-16 23:41, so nothing has shipped for over a day. Five tests fail, and all five depend on **ambient host state** rather than fixtures they establish themselves. They pass where that state happens to exist and fail where it does not, which makes them unusable as a release gate and impossible to reproduce reliably.

## Failing tests

```
homeboy-agents  agent_task_service::cook::tests::adoption_replays_provider_discovery_failure_in_the_same_recipe_attempt
homeboy-agents  agent_task_service::cook::tests::repeated_provider_discovery_failures_exhaust_the_adoption_review_allowance
homeboy-cli     commands::agent_task::fanout::tests::cook_batch_does_not_warn_for_specific_backend_names_in_core
homeboy-cli     commands::agent_task::fanout::tests::cook_batch_dry_run_returns_status_and_resume_commands
homeboy-cli     commands::agent_task::run::preview_tests::compiled_plan_preview_emits_placement_admission_schema
```

Two more in the same module fail alongside the last one locally:

```
commands::agent_task::run::preview_tests::local_preview_projects_prompt_evidence_without_execution_admission
commands::agent_task::run::preview_tests::preview_rejects_an_existing_unsafe_path_without_changing_it
```

## These are not a regression from recent merges

Measured three ways:

| Tree | `preview_tests` result |
|---|---|
| `79eec1a02` — last **green** release | **3 failed** |
| current `main` | **3 failed** — identical set |

`cook_batch_dry_run_returns_status_and_resume_commands` likewise fails at `79eec1a02`.

So the same tests fail at a commit where CI's `Test` job **passed**. Temporal correlation with recent merges is misleading; the code path was not changed by them. What changed is that the CI environment stopped satisfying a precondition these tests never establish for themselves.

## The ambient dependency, named

`compiled_plan_preview_emits_placement_admission_schema` fails with:

```
Error {
  code: ValidationMissingArgument,
  message: "Missing required argument",
  details: { "args": [
    "--repo <repo> is required because the supplied Git checkout has no configured
     repository remote mapping; provide --repo <repository-or-configured-component>"
  ]}
}
```

The test relies on the ambient Homeboy **repository/component registration** resolving the checkout it runs in. Where a matching registration exists it passes; where it does not it fails. Nothing in the test creates that registration.

[...]

Issue #14744, opened 2026-09-17
Cook cannot classify remediation when a provider times out after producing a recoverable candidate


chubes4:
## Problem

A Cook whose provider times out **after** producing a substantive candidate becomes unrecoverable through its own lifecycle, even though every downstream signal is healthy.

Observed sequence:

1. The provider produced a complete, fingerprinted candidate and then hit its wall-clock timeout.
2. The attempt terminalized as `CandidateRecoverable`, classified `timeout`, with the guidance `provider timed out after producing a recoverable candidate; promote the fingerprinted patch through controller gates`.
3. `agent-task cook-continue --preflight` reported `admitted: true` with `deterministic_checks_passed`.
4. `agent-task cook-continue` and `cook-continue --rearm` both returned:

```
status: policy_failure
stop_reason: cannot classify Cook remediation without terminal executor identity
```

The timed-out provider never emitted its declared `patch`, `agent_result` and `transcript` artifacts, so no terminal executor identity exists. Remediation classification appears to require that identity unconditionally, including on the path the diagnosis itself recommends.

`agent-task status` then offers no usable recovery. `cancel`, `resume`, `retry`, `quarantine` and `rearm` are all unavailable; only `promote` is reachable, and it is `indeterminate` pending preflight.

## Why it matters

The work is not lost, but the durable Cook cannot finish. Recovery requires stepping outside the Cook lifecycle: `agent-task promote` to place the candidate, then running the declared gate by hand, which also detaches the gate evidence from the Cook record. `finalize-pr --recover` then correctly refuses to publish, because no visible durable gate evidence is bound to the promoted candidate.

That is the right safety behavior for `finalize-pr`. The gap is upstream: a recoverable candidate should stay inside its own lifecycle.

## Expected

A timeout that yields a recoverable candidate should remain remediable through `cook-continue`, with classification degrading safely when terminal executor identity is absent, rather than refusing to classify. The recorded provider timeout and missing declared artifacts already describe the terminal state.

Whatever route is chosen, a continued Cook should be able to run its declared gates against the promoted candidate and bind that evidence to the Cook record so `finalize-pr` can publish normally.

## Acceptance

- A Cook attempt that terminalizes as `CandidateRecoverable` with zero declared executor artifacts can be classified and continued.
- Continuation runs the declared verification gates and binds their evidence to the promoted candidate.
- `finalize-pr --recover` publishes when those gates are green, and still refuses when they are not.
[...]

PR #14864, opened 2026-09-21
fix(worktree): reclaim squash and rebase merged task branches

### LINKED ISSUE #14745: delete_when_merged never fires for squash-merged branches, so worktrees accumulate until the disk fills
## Symptom

Worktrees declared `branch_cleanup_intent: delete_when_merged` + `cleanup_policy: remove_when_safe` are never reclaimed after their branch merges. `homeboy cleanup --apply` skips them with:

```
"reason": "branch is not merged into the cleanup base ref"
```

That reason is **false**. Sampled 19 active worktrees across three components and checked each branch against its GitHub PR state:

```
MERGED  15
OPEN     3
NO PR    1   (release machinery branch)
```

79% have merged branches and are still held. Examples, all merged upstream and all still registered active:

```
feat/771-stale-listing-retirement          MERGED
feat/816-js-gated-affiliate-ticket-links   MERGED
feat/818-canonical-ticket-urls             MERGED
fix/817-affiliate-urls-in-calendar-exports MERGED
refactor/822-block-name-constant           MERGED
ci/5-homeboy-gate                          MERGED
feat/7-network-endpoint                    MERGED
feat/195-ability-site-affinity             MERGED
phase-0-scaffold                           MERGED
```

## Likely cause

Every one of these merged via **squash merge**, which produces a new commit on the base branch with no ancestry link to the branch tip. A local `git branch --merged` / `git merge-base --is-ancestor` test therefore reports "not merged" and always will, no matter how long the branch has been merged.

Compounding it: the base ref may not be fetched before the comparison, so even a fast-forward merge could read as unmerged on a stale local `origin/main`.

## Impact

This is the root cause of a production incident on this host, not a tidiness issue.

Worktrees accumulate without bound — 205 registered, 82 active, ~60 of those dead. Combined with per-run temp directories that outlive them, the workspace volume hit **100% full (757M free of 196G, 38G of it stale temp)**. That surfaced as:

```
ENOSPC: no space left on device, mkdtemp '/mnt/extrachill-workspace/tmp/wp-codebox-...'
```

which failed 11 smoke tests and blocked `homeboy release wp-codebox` at `preflight.test`. The failure looked like a code regression and took a full diagnostic pass to trace back here.

## Suggested direction

Determine merge state from the remote rather than local ancestry — query the forge for the branch's PR state, or compare against a freshly fetched base and treat "no commits unique to the branch after squash" as merged. Whatever the mechanism, squash merge is the default across these repos, so ancestry-based detection cannot work.

Worth surfacing the determination in `cleanup --apply` output too: the current skip reason states a conclusion without the evidence behind it, which is why this went unnoticed for weeks.

chubes4:
Closes #14745.

## Problem

Worktree reclamation asked `git merge-base --is-ancestor`. A GitHub squash or rebase merge rewrites the landed commits, so the branch's original commit objects are never reachable from the target even though every line they touch already lives there. The predicate is therefore permanently false for the merge styles this repository actually uses, and finished worktrees are retained forever.

Measured today: after nine squash-merged pull requests, `homeboy cleanup --include task-worktrees` reported `candidate_count: 0`, `skipped_count: 112`, `reclaimed_bytes: 0`. Every one of those worktrees was clean with its work already on `main`, and all had to be removed by hand.

## Change

Ancestry stays the cheap exact test. When it fails, two content probes run:

- `git cherry` patch-id equivalence per commit, which recognizes a rebase merge.
- a synthetic commit carrying the aggregate diff against the merge base, probed the same way, which recognizes a squash where many commits collapse into one.

A genuinely unmerged branch fails both. Any probe failure falls back to the conservative existing answer rather than assuming containment.

## Verification

- `cargo test -p homeboy-core --lib worktree::tests::cleanup_reclaims`: 3 passed — squash of multiple commits, rebase, and ordinary fast-forward.
- `cargo test -p homeboy-core --lib worktree::tests::cleanup_retains`: 3 passed — genuinely unmerged commits, a dirty worktree whose branch is squash-merged, and a squash-merged worktree claimed by a live owner.
- Each reclamation test asserts first that its rewrite actually defeats plain ancestry, so it cannot silently stop proving anything.
- `cargo fmt --all --check`: passed.

The rebase fixture needed correcting during review: replaying commits onto an unchanged target reproduces identical SHAs (same tree, parent, message, author and date), so ancestry still held and the probe went untested. The target now advances first, which is what a real rebase merge does.

`worktree::tests::inventory_reports_a_path_restored_during_authority_snapshot` failed once under a full parallel run with `task-worktree registry lease deadline exhausted`, and passes in isolation. That is registry-lease contention between concurrent tests, an instance of #14751 rather than of this change.

## AI assistance

AI assistance: Yes. The candidate was produced by a Homeboy Cook run (`cook-detached-a4b16492`) using Claude Sonnet 5 via OpenCode. Its provider timed out after producing the candidate and the `cargo fmt` gate failed, so Cook could not classify remediation (#14744) and `cook-continue` refused the run as ineligible for durable retry. GPT-6 Astra via OpenCode then formatted the candidate, found and fixed the rebase fixture defect, ran the verification above, and opened this PR. Finalization occurred outside Homeboy under operator authorization; Homeboy remains the release owner.


chubes4:
The failing `homeboy / Test` check is a gate-selection defect, not a finding against this change: the changed-scope selector returns zero tests for a change confined to `crates/homeboy-core/src/**`, and zero selection is then rejected as invalid evidence. Filed as #14867 with the full evidence; reruns reproduce it deterministically, and `Lint`, `Rustfmt`, `Baseline Test`, and the candidate job are green.

The tests for this change do exist and pass when named directly: `cargo test -p homeboy-core --lib worktree::tests::cleanup_re -- --test-threads=1` reports 8 passed, covering squash, rebase, fast-forward, genuinely-unmerged, dirty, and live-claimed cases.

One correction landed after review: the squash fallback originally built a synthetic commit with `git commit-tree`, which writes a commit object into the repository being inspected. Worktree inventory and cleanup must not mutate the object database of a checkout they only report on, so the probe now compares per-path content against the target instead. That is read-only and conservative — if the target later changed one of those paths again, containment is not claimed and the worktree is retained.

AI assistance: GPT-6 Astra via OpenCode diagnosed the gate defect and made the read-only correction under operator authorization.

PR #14879, opened 2026-09-22
fix(notification-route): allow a transport-only route, consult its resolver

chubes4:
## Rule

A caller-supplied notification transport is valid on its own; the route is optional and, when absent, is derived by that transport's declared route resolver, falling back to no route when the resolver reports no match — a caller-supplied route without a transport remains invalid.

## Why route-less delivery is correct, not a loosening

`NotificationRoute` keeps its existing invariant: a *present* route is non-empty, validated, and free of control characters/credential syntax. That invariant was never the bug. The bug was that the contract layer had no way to represent "transport known, route not yet known" at all — `from_cli_or_env` forced every transport into either a full `NotificationRoute` pair or a hard error, so a transport that legitimately has no route yet (no resolver declared, or a resolver that reports no match) could never reach the caller. Every layer *below* this one already treats that as a normal, valid state:

- `homeboy-extensions/discord/scripts/notify.mjs`: `const parsedRoute = route === undefined ? undefined : parseRoute(route)`, then `route_kind: parsedRoute?.kind || 'webhook'` — with no route it delivers to the webhook's own channel. A `channel` route is explicitly refused in webhook mode, so route-less is the *only* way to target a webhook's own channel.
- `homeboy-extensions/discord/scripts/resolve-route.mjs`: emits `{status:'unmatched'}` when `KIMAKI_THREAD_ID` is absent — the normal case on a CI runner.
- `crates/contracts/homeboy-lab-contract/src/notification_route.rs` (pre-existing): `NotificationRouteResolution` already models `transport: Option<String>` independent of a route, plus a `"route_less"` classification and `missing_context: Vec<String>`.
- `crates/contracts/homeboy-extension-contract/src/notification_transport_config.rs`: `route_resolver: Option<NotificationRouteResolverConfig>` on a transport declaration exists *specifically* to derive a route the caller didn't supply.
- `crates/homeboy-core/src/notify.rs` `NotifyEvent`/`dispatch()` (pre-existing, untouched by this PR): `transport: Option<String>` and `route: Option<String>` are already independently optional, and `dispatch()` already passes `--transport` without `--route` to an installed transport when route is `None`.

So the contract-layer validation was the one place in the stack that hadn't caught up — a strictly-more-permissive fix that lets a real, already-supported state flow through, not a relaxation of any actual invariant.

## What changed

- **`homeboy-lab-contract/src/notification_route.rs`** — `from_cli_or_env` (~line 131) now returns a new `NotificationRouteContext` (`None` / `TransportOnly(String)` / `Route(NotificationRoute)`) instead of `Option<NotificationRoute>`. A route is still rejected outright without a transport (both the CLI-half-set and the env-half-set case, ~former line 147). A transport alone — CLI or env — is now `Ok(TransportOnly(..))`, still validated as a transport id. This is the only place `NotificationRoute`'s own invariants are enforced, and they are untouched.
- **`homeboy-core/src/notification_route_resolver.rs`** — `resolve_from_cli_or_env_with_evidence` now handles `TransportOnly` by calling a new `resolve_declared_transport_route`, which looks up *that one* transport's declared `route_resolver` among installed extensions and invokes it (reusing the existing `invoke()` machinery used by ambient discovery):
  - resolver **matched** → the route it returns, with `evidence.classification = "resolver"`.
  - resolver **unmatched**, or the transport has **no declared resolver at all** → `Ok(route: None)`, `evidence.classification = "route_less"`, `evidence.transport = Some(transport)` — the identity is preserved, not dropped.
  - resolver **fails to start, times out, or returns invalid output** → a real `Err`. Unlike ambient discovery (which tolerates a broken resolver among many candidates and keeps scanning), here the caller named exactly one transport, so there's no fallback candidate to silently continue past.
- **`homeboy-core/src/notify.rs`** — `NotifyEvent::with_route` now falls back to the bound `notification_route::current_resolution().transport` when the route is `None`, so a resolved-but-route-less transport selection is actually delivered through its own transport rather than silently falling back to whatever `defaults().notifications.default_transport` happens to be configured. This uses only pre-existing thread-local evidence plumbing (`with_current_resolution`, already bound around every command in `cli_runtime.rs`) — no new state.
- **`homeboy-cli/src/cli_runtime.rs`** — the ambient-discovery fallback (only reachable for `agent-task cook`/`cook-batch`) previously re-triggered whenever `notification_route.is_none()`. With this fix, transport-only-but-route-less is also `notification_route.is_none()`, so without an added guard, ambient discovery would silently override an explicit user selection with an unrelated ambiently-discovered transport. Added `&& notification_resolution.evidence.transport.is_none()` so only *truly* empty context still falls through to ambient discovery.
- **`homeboy-cli/.../local_detach.rs`, `local_detach_fanout.rs`** — `detached_route()` bypasses resolver consultation by design (documented as "must never take a cook down with it"). Updated only to compile against the new return type via `NotificationRouteContext::into_route`, preserving the exact same effective behavior (transport-only still resolves to `None` here, exactly as it did when it was an `Err` before).

No transport id or vendor name appears anywhere in the touched Rust — `resolve_declared_transport_route` takes an arbitrary `&str` transport id and looks it up generically against installed extensions' `notification_transports`.

## Tests (behavioural)

`crates/contracts/homeboy-lab-contract/src/notification_route.rs`:
- `transport_only_cli_context_is_valid`, `transport_only_environment_context_is_valid` — transport-only via CLI and via env both parse to `TransportOnly`.
- `transport_only_still_validates_the_transport_id` — a malformed transport id is still rejected even with no route.
- `route_without_transport_is_still_rejected` — route-only (CLI and env) is still a hard error.

`crates/homeboy-core/src/notification_route_resolver.rs`:
- `transport_only_consults_that_transports_own_resolver` — installed resolver returns `matched` → the derived route is used, `evidence.classification == "resolver"`.
- `transport_only_with_an_unmatched_resolver_still_delivers_route_less` — installed resolver returns `unmatched` → `Ok(route: None)`, `evidence.transport == Some("synthetic.completed")`.
- `transport_only_with_no_declared_resolver_delivers_route_less` — no resolver declared at all → also `Ok(route: None)`, transport preserved in evidence (not an error).
- `transport_only_with_a_failing_resolver_is_a_real_error` — malformed resolver output → `Err` (not silently swallowed).
- `route_without_transport_evidence_is_invalid` — explicit-CLI route-only still produces `classification == "invalid"` evidence.
- `explicit_and_environment_context_have_distinct_evidence`, `explicit_route_wins_without_invoking_a_resolver` — updated for the new return type; still assert a full explicit route never invokes a resolver.

`crates/homeboy-core/src/notify.rs`:
- `route_less_transport_selection_is_not_overridden_by_operations_default` — with an installed `chosen.transport`, a configured `default_transport = "other.default"`, and bound resolution evidence naming `chosen.transport` route-less, `dispatch()` still delivers through `chosen.transport`, not the configured default.

Ran (all green):
```
cargo test -p homeboy-lab-contract                              → 58 passed
cargo test -p homeboy-core notification_route                   → 16 passed (notification_route_resolver::)
cargo test -p homeboy-core --lib notify::                       → 16 passed
cargo test -p homeboy-cli --lib notification                    → 4 passed
cargo test -p homeboy-cli --lib local_detach                    → 70 passed
cargo test --test cli_binary notification_route_admission       → 1 passed
cargo test -p homeboy-agents --lib notif                        → 36 passed, 1 ignored (by design)
cargo test -p homeboy-lab-runner --lib notif                    → 1 passed
cargo build --workspace                                         → clean, no new warnings in touched files
```

Also manually reproduced the exact repro from the issue against a freshly built binary in an isolated `$HOME`:
```
$ HOME=/tmp/homeboy-repro-test HOMEBOY_NOTIFICATION_TRANSPORT=discord.run-completion \
    ./target/debug/homeboy deps status extrachill-network
{
  ...
  "summary": "Component not found",
  "diagnostics": { "code": "component.not_found", "details": { "id": "extrachill-network" } }
}
```
No more `"HOMEBOY_NOTIFICATION_TRANSPORT and HOMEBOY_NOTIFICATION_ROUTE must be set together"` — the command now fails for the expected, unrelated reason (no such component in this sandbox).

## Pre-existing failure comparison (like-for-like, confirmed via stash)

Per the task brief, `orchestration::tests::exact_ref_preparation_preserves_duplicate_source_version_target_artifact_paths` was flagged as a known pre-existing failure in the full `homeboy-deploy` suite. I did not touch any `homeboy-deploy` code (grepped: no `notification_route`/`notify::` references in that crate).

Compared full suite to full suite, both under default (parallel) `cargo test -p homeboy-deploy`:
- **`git stash` (clean origin/main)**: 343 passed, 1 failed — exactly `orchestration::tests::exact_ref_preparation_preserves_duplicate_source_version_target_artifact_paths`.
- **This branch**: 342–344 passed, 0–2 failed across repeated runs. The named pre-existing failure reproduces identically every time; an *additional*, non-deterministic failure appears in some runs (`preparation::tests::equal_preparation_request_builds_once_and_cleans_only_its_payload_copy` once, `preparation::tests::exact_ref_preparation_builds_detached_bytes_and_cleans_owned_resources` another time — different test each run, both about git-checkout/build staging, neither touching notification routing).
- **`cargo test -p homeboy-deploy -- --test-threads=1` on this branch**: **344 passed, 0 failed** — clean. This confirms the flakiness (both the named pre-existing failure and the extra ones) is parallel-execution resource contention (shared temp/git state under `--test-threads` default), not something introduced by this change.

## Constraints followed
- No `CHANGELOG.md` edit, no version bump.
- No changes outside `Extra-Chill/homeboy`; nothing in `homeboy-extensions` or any consumer workflow touched.
- Lines kept under 100 chars by hand (`cargo fmt`/`clippy` unavailable in this environment); the only two lines over 100 are `install_resolver(...)` test-fixture string literals matching the exact pre-existing style already used throughout this same test module (verified against `git show HEAD` — 6 pre-existing lines in this file already exceed 100 chars in the identical pattern).


chubes4:
Fixed the red Rustfmt gate in a formatting-only commit. Verification: `cargo fmt --all -- --check`, `git diff --check`, notification-route contract tests (19 passed), resolver tests (15 passed), notify tests (16 passed).

Finalization provenance: Chris Huber authorized direct landing repairs after Homeboy Cook failed before provider execution; completed in the isolated PR worktree outside Cook.

AI disclosure: GPT-6 Astra via OpenCode applied rustfmt, verified, and pushed under Chris Huber direction.

PR #14920, opened 2026-09-22
fix(admission): keep Lab execution off hot controller

### LINKED ISSUE #14700: Lab-placed cooks are gated by controller resource admission while the runner is ready
## Problem

A Cook explicitly placed on the Lab is held by **controller** resource
admission, even though the runner is connected and ready. Offloading work to the
Lab is supposed to be how a busy controller keeps making progress; instead the
busy controller blocks the offload.

## Evidence

```
$ homeboy agent-task cook ... --placement lab --detach-after-handoff
Resource policy: controller is hot. Load average is 28.7 across 18 CPU(s).
Workload `agent-task cook/...` is routed to Lab runner `homeboy-lab`.
```

Then:

```
status:            pending_resource_admission
runner_readiness:  connected_ready
retry:             { max_attempts: 20, policy: bounded_exponential }
```

The runner is `connected_ready` and idle. The work never starts because the
controller is loaded. With load at 87 across 18 CPUs, the queue simply waits.

## Why it matters

This is the single biggest reason to stop trusting Lab offload and just run
things locally or by hand. The mental model an operator wants is "the Lab is an
extension of my machine" — I hand it work and my machine's load stops being the
constraint. Today the opposite is true: the busier my machine gets, the less
able I am to hand work away.

## Expected

Controller-side admission should gate only the controller-side portion
(preflight, transport, evidence). Once a workload is routed to a ready runner,
its execution should be admitted against **that runner's** capacity, not the
controller's load average.

## Related

#13354 covers diagnose rewriting Lab runs into local retries on hot controllers.
This is the admission-side counterpart: Lab-routed work never starting because
the controller is hot.

## AI assistance

Claude Sonnet 4.5 via Kimaki/OpenCode encountered this while orchestrating real work today, reproduced it, and filed this report. Chris Huber directed the session.

chubes4:
## Summary
- admit provider execution against connected-ready Lab runner readiness, not controller load
- keep controller pressure as bounded transport/setup evidence and present placement-specific notices
- project early detached Cook admission as runner-selected instead of a controller placeholder

## Verification
- `cargo fmt --all -- --check`
- `CARGO_BUILD_JOBS=2 cargo test -p homeboy-core parsed_command_preflight::tests`
- `CARGO_BUILD_JOBS=2 cargo test -p homeboy-agents orchestration::tests::placement_`
- `CARGO_BUILD_JOBS=2 cargo test -p homeboy-cli --lib cli_runtime::tests::explicit_lab_placement_resolves_only_a_ready_policy_runner`

Closes #14700.

AI assistance disclosure: this repair was finalized outside Homeboy under explicit operator authorization using OpenCode with model `openai/gpt-5.6-luna`. AI traced the issue, proposed and implemented the owning-layer changes, added regression tests, and assisted with verification; the operator authorized and reviewed the resulting changes.


chubes4:
Reviewer follow-up pushed in commit `f0d4effd1`. The admission bypass now requires the resolver's `selected_runner_admitted` and resolved effective placement `lab`; explicit local and auto/lab-or-local local routes remain controller-pressure gated. Added stale/mismatched readiness counterexamples and preserved `lab_or_local` in the control-plane requested placement contract.

Focused core and agents suites pass. Current CI: Plan, Rustfmt, and Control Plane Contract Boundary pass; Lint and Prepare candidate Homeboy binary remain pending; Consumer contract is skipped by workflow. No CI failure observed.

Issue #14912, opened 2026-09-22
Persisted Lab staging retention field breaks new handoffs after upgrade


chubes4:
## Observed failure

On Homeboy `0.383.7+54508ca5aa6619aede4b70ba223680bf47460245`, a Cook passed preview (Lab admissible, provider readiness passed), was durably admitted, and failed before any provider ran at `lab_staging_submission`:

```text
parse runner-sessions/homeboy-lab-staging/store.json:
unknown field `preserve_workspace_on_failure`
expected ... `delete_workspace_on_failure` ...
```

Run: `agent-task-e496ad16-d04b-4c6c-88f6-69058c52c45d`
Attempt: `agent-task-e496ad16-d04b-4c6c-88f6-69058c52c45d-attempt-1-85dd2c7e`
Underlying task: Automattic/blocks-engine#2112.

The advertised controller version was current and the runner was `connected_ready`. A persisted staging record appears incompatible with the reader. Verify the exact field history and owning storage contract rather than assuming its boolean maps directly to the new deletion policy.

## Expected

- Versioned staging state survives supported upgrades through a narrow explicit migration, preserving workspace-retention intent and active records.
- Reading one historical record does not poison unrelated new handoffs.
- Required migration or actual inability to read staging state is surfaced by readiness before a multi-minute failed handoff, with one documented recovery action.
- Regressions exercise the historical serialized record and current write/read round trip, not merely a newly constructed current struct.

AI assistance: GPT-6 Astra via OpenCode observed the failure, checked open trackers, and drafted this report. Chris Huber directs the repair.

chubes4:
Taking the narrow persisted staging-schema repair as an explicitly authorized direct-runtime fallback after Cook and runner reconciliation failed. This also blocks #12581. Reproduction: `homeboy --runner homeboy-lab agent-task diagnose agent-task-382d8b84-be2c-4138-8cf7-917ff2f47fe2-attempt-1-59fc96c6`; observed unknown `preserve_workspace_on_failure` in the persisted staging store before provider execution. Work will use an isolated tracker-linked worktree and deterministic regression verification; finalization outside Cook will be recorded here. AI assistance: OpenAI gpt-6-astra via OpenCode investigated and orchestrates this repair; the implementing runtime/model and evidence will be recorded with the result.

chubes4:
[...]

PR #14931, opened 2026-09-22
fix(cli): return after durable handoff unless --wait

### LINKED ISSUE #12581: --detach-after-handoff is a wait-policy flag named as an ownership flag, and in-tree execution is a silent degradation only attached callers get
## Summary

Two findings from reading `local_detach.rs`, both of which change how #12570 should be fixed.

1. **`--detach-after-handoff` does not control detachment.** Attached local cooks are already re-executed in their own session. The flag controls whether the client *waits*, not who *owns* the process.
2. **In-tree execution is not a designed alternative.** It is one of three degradation paths, and every one of them errors loudly for detached callers while silently degrading for attached ones.

## 1. The flag is misnamed

`intercept_local_detached_cook` fires on:

```rust
if !is_unsupervised_local_cook(cli)
    || (!cli.detach_after_handoff && provider_placement != Some("local"))
{
    return Ok(None);
}
```

An **attached** local cook satisfies the second clause, so it takes the same path as a detached one: re-exec in its own session, launch token, daemon-owned controller job, start-identity pinning, supervisor projection. All of it.

The only divergence is ~160 lines later:

```rust
if cli.detach_after_handoff {
    let handoff = await_durable_handoff(&cook_id, &mut child, handoff_timeout())?;
    // ...return
}
// "Attached callers stream immediately; they do not receive a
//  handoff acknowledgement to classify."
```

So the flag selects **return-after-handoff vs stream-until-terminal**. That is a wait policy. The name says ownership.

This is why the help has to carry the warning *"Do not infer the wait policy from client interactivity"* — the documentation is compensating for the name. `--no-wait` or `--return-after-handoff` would describe what it does, with the current spelling kept as an alias.

## 2. Every fallback is loud for detached and silent for attached

Four `return Ok(None)` paths mean "run inside this client's process tree":

| line | condition | detached | attached |
|---|---|---|---|
| 190 | platform lacks session-detachment support | **errors** | silently in-tree |
| 197 | `runner_side` | **errors** | silently in-tree |
| **244** | **resident daemon build mismatch** | **errors** | **silently in-tree** |
| 176 / 205 | launch token consumed, i.e. the re-exec'd child | correct | correct |

The asymmetry is the defect. A detached caller is told its request cannot be served; an attached caller is given the hazardous shape with no signal. The line 244 comment states the intent plainly:

> An attached caller can retain foreground ownership. This keeps a newer client useful without letting an older resident daemon interpret and terminate lifecycle records it does not understand.

The goal is right. The cost is silently reverting to the exact shape that orphans work in #12570, and nothing reports it.

## Why this matters for #12570

I originally read the orphan as "attached local placement is the default and it is unsafe." That is wrong. Attached placement is *normally* safe, because it detaches too. What I actually hit was a **degradation** — and the reason it was invisible is this asymmetry.

Environment when it happened, consistent with line 244:

```
installed CLI                 homeboy 0.345.0+d478a72959b2
main                          v0.350.3
newest controller runtime     homeboy_0_345_0_d478a72959b2-e38ea999...
resident controller daemon    none running
```

Five releases of drift, no resident daemon of the current build. I am not asserting which branch fired without instrumenting it, but a build-mismatched or unreachable controller daemon is the leading candidate, and it is directly testable.

## Suggested

1. **Rename to describe wait policy.** `--no-wait` / `--return-after-handoff`, keeping `--detach-after-handoff` as a deprecated alias. Then the docs' "do not infer the wait policy" caveat can go away, because the flag will say it.
2. **Make degradation loud for attached callers.** When any of lines 190 / 197 / 244 fire, say which one and why. An attached caller does not need to be blocked, but it must not be uninformed.
3. **Name the cause in the warning.** #12577 currently emits a generic "attached local placement" line. With the cause available it can say *"running in-tree because the resident controller daemon is build-mismatched (client 0.345.0)"*, which is actionable rather than advisory.
4. Consider surfacing the degradation on the durable run record, so `agent-task status` can explain an orphaned run after the fact rather than leaving it wedged at `queued` with no clue.

## AI assistance

- **AI assistance:** Yes
- **Tool(s):** Claude via OpenCode
- **Used for:** Reading `local_detach.rs`, tracing the interception guard and fallback paths, collecting the version evidence, and drafting this report.

Refs #12570, #12577.

chubes4:
## Summary

Closes #12581 with the user-approved wait-policy inversion.

- Eligible top-level durable work returns after controller handoff by default. Global `--wait` observes terminal completion and returns its terminal status. Remove the obsolete `--detach-after-handoff` spelling rather than retaining an alias.
- Keep placement/admission independent from observation: local Cook, fallback Cook, fanout, and changed-only summary review retain their execution owner while a waiting client observes. Lab Cook uses durable admission in either mode; waiting on queued admission does not report acceptance as success.
- Normalize controller/retry and runner children to complete their assigned work, preserving forwarded argument tails. Keep internal execution-option defaults synchronous. Release preflight and deferred-workload children explicitly wait for results.
- Preserve synchronous preview/planning and read-only polling behavior; update composed routing, raw-argv provenance, generated Fuzz commands, help, and command documentation.
- Add parsing/placement, forwarded-argument, and end-to-end terminal-exit regressions. Update existing handoff tests to exercise the default and attached tests to request `--wait`.

## Verification

**The required broad gates are not all green.** Failures were investigated against immutable baseline `283c0be764c5c221d06b22307e2e0c809f278ddc`, which was fresh `origin/main` when this isolated implementation started. No unrelated baseline repairs are included.

Tests used shared Cargo resources (`CARGO_TARGET_DIR=/Users/chubes/Developer/homeboy/target`). After the initial CLI run encountered a stack overflow, subsequent runs used `RUST_MIN_STACK=16777216`. The baseline ran from a separate detached worktree with its own existing shared target directory.

### Required gates

| Command | Observed result |
| --- | --- |
| `cargo fmt --all --check` | **PASS**, including the final tree. `git diff --check` also passes. |
| `cargo test -p homeboy-cli --lib` | **Not passing.** Enlarged-stack implementation and immutable-baseline runs both reached the repository's 1,500-second suite deadline. Both also report existing failures including missing registered docs, raw-argv inventory, help/lifecycle expectations, provider/model provenance, and diagnosis fixtures. A full-run fanout timing failure passed when isolated. Changed-path regressions discovered during implementation were corrected and checked separately below. |
| `cargo test --test cook_lab_handoff_test --test reverse_cook_queue_acceptance` | **Not passing.** The last full sequential Cook handoff run had **16 passed / 1 failed**. `attached_nonlocal_cook_observes_terminal_execution` encountered warm-host admission with no runner; with explicit `--wait`, it correctly remained queued until the fixture's 180-second budget expired. The immutable baseline also fails this fixture by returning pending admission instead of the expected terminal execution. |
| `cargo test --test reverse_cook_queue_acceptance` | Run separately because Cargo stops after the failed first integration target. **Fails identically on implementation and immutable baseline:** no reverse job enqueued within 120 seconds; failure diagnostics then reject the fixture's `agent-task status --full` argument. Implementation: 133.13s; baseline: 131.32s. |

The other host-dependent fixture, `no_runner_lab_or_local_cook_reaches_local_execution`, passed in the full sequential run but failed a later focused run on a hot host (load 28.1 / 18 CPUs). The immutable baseline independently exhibits the same local-provider resource-policy refusal. Its terminal output-file assertion accommodates the supervised waiting path.

### Passing focused checks

These are focused evidence, **not substitutes claimed as a passing full library gate**. Counts overlap where filters overlap.

- **223 CLI tests passed** with sequential execution: parsing and generated/scoped help, composed Lab contracts, preview wait-policy matrix, deferred-workload orchestration, Fuzz generated commands, terminal status/watch, release orchestration, placement-independent Cook admission, and the updated raw-argv inventory check. Three already-failing baseline tests were explicitly excluded: `command_registry_docs_paths_exist_and_are_indexed`, `compiled_plan_preview_emits_placement_admission_schema`, and `local_preview_projects_prompt_evidence_without_execution_admission`.
- **79 CLI handoff/contract tests passed:** `cargo test -p homeboy-cli --lib -- local_detach command_contract::lab cook_admission_placement_is_independent_of_wait_policy --test-threads=1 --quiet`.
- **82 runner tests passed:** `cargo test -p homeboy-lab-runner --lib -- lab_args:: lab_command:: --quiet`.
- **49 watch/review tests passed:** `cargo test -p homeboy-cli --lib -- commands::agent_task::status::watch_tests commands::review::tests --test-threads=1 --quiet`.
- Recursive rig flag-stripping regression passed individually.
- New end-to-end `explicit_wait_and_runner_owned_cooks_return_terminal_failure` **passed**, both within the full integration run and in a later focused run. It proves provider execution and delayed completion, terminal exit code **1**, and no premature handoff result for both explicit local `--wait` and runner-owned execution without an observation flag.
- The full integration run also passed default verified local handoff, addressable run/watch commands, preview/help behavior, cancellation/ownership tests, and survival of an explicitly waiting client's termination.

Reproduce the 223-test focused set (with the environment above):

```sh
cargo test -p homeboy-cli --lib -- \
  cli_surface:: command_contract::lab commands::deferred_workload:: \
  commands::fuzz::stable commands::agent_task::run::preview_tests:: \
  commands::agent_task::status:: commands::release:: \
  cook_admission_placement_is_independent_of_wait_policy \
  an_acknowledged_scan_that_no_longer_exists_is_a_failure \
  --skip command_registry_docs_paths_exist_and_are_indexed \
  --skip compiled_plan_preview_emits_placement_admission_schema \
  --skip local_preview_projects_prompt_evidence_without_execution_admission \
  --test-threads=1 --quiet
```

## Authorized execution and retained evidence

- Initial Homeboy Cook: `agent-task-382d8b84-be2c-4138-8cf7-917ff2f47fe2`, failed before provider execution because of #14912 (legacy `preserve_workspace_on_failure` in the Lab staging store). Preview/providers/status/runner diagnosis and reconciliation had already been attempted; reconciliation failed.
- Chris authorized this direct isolated fallback in Discord message `1551947697081294921`; execution followed the supplied fallback authorization. No additional Cook or subagents were launched.
- Implementing OpenCode session: **`ses_f36ae09e4ffeX41KXDwbAxh6wY`**.
- Invocation: `opencode run "Implement the attached authorized task completely, verify and finalize a PR, and report evidence. Do not merge or release." --dir /Users/chubes/Developer/homeboy@fix-issue-12581-homeboy --model openai/gpt-6-astra --format json --auto --title "Authorized direct fallback: Homeboy 12581 default detach" --file /var/folders/lr/c_cmmt7s0592m4njz99v5yb40000gn/T/opencode/homeboy-12581.md`.
- Retained under the same temporary evidence directory: `homeboy-12581-direct-run.jsonl`, `homeboy-12581-cli.log`, `homeboy-12581-final-cli.log`, `homeboy-12581-baseline-cli.log`, `homeboy-12581-integration.log`, and `homeboy-12581-baseline-integration.log`. Later focused/integration command results are retained in the OpenCode session event log.
- The primary checkout and independent #14912 repair were preserved. Merge and release remain with the orchestrator.

## AI assistance

- **AI assistance:** Yes.
- **Implementing model:** OpenAI **gpt-6-astra** (`openai/gpt-6-astra`) through **OpenCode**.
- **Orchestration:** OpenAI gpt-6-astra through OpenCode; **Homeboy** was used for initial Cook admission/diagnosis and remains the orchestrator-owned shipping path.
- **Used for:** Source/tracker analysis, implementation, regression coverage, direct authorized verification, immutable-baseline comparison, and PR preparation. Chris Huber directed and authorized the task.


chubes4:
CI candidate evidence in run 35747860356 exposed a missed old-flag assertion in the replay reconstruction test. Fixed in 5580d4019: the test now parses the reconstructed CLI and asserts effective default handoff rather than requiring the retired spelling. Exact regression passes (1 passed, 0 failed); `cargo fmt --all --check` passes. The Candidate Test wrapper reporting success is not a clean test verdict: its retained result contains five failures and triggers differential baseline testing. I am waiting for that comparison and the updated-head checks before shipping. AI assistance: OpenAI gpt-6-astra via OpenCode reviewed CI logs, corrected the assertion, and ran verification.

chubes4:
Updated CI fixed the replay assertion and now reports a different candidate-only identity: `cli_runtime::tests::cli_startup_registers_both_extension_audit_providers` (aggregate literal count 0 vs 1). This unchanged test passed in the preceding candidate run and passes on the current head in isolation (1 passed, 0 failed, 0.40s). The other four failures remain inherited. Rerunning the candidate job once with dependent differential reconciliation to check this intermittent result; no gate bypass or merge performed. AI assistance: OpenAI gpt-6-astra through OpenCode inspected retained candidate/baseline artifacts and ran the exact test.

chubes4:
Merged as eb76c60e7320691853596e8c5d1f9d1121080ae8 after the final differential Test, lint, build, and formatting checks passed. Release is not published yet: `homeboy release homeboy --apply` stopped at preflight.lint because the shared Cargo filesystem was below its configured 150 GiB reserve (approximately 136 GiB available), with no tag created. The documented controller/Lab upgrade path then exhausted its 1,200-second admission wait behind runtime pin `retry-ff053291-7533-55fe-be88-3d42154ca6b5`; upgrade operation f7550515-09ab-42ed-b43f-ac35fa8c93d7. The pin-owning run still reports running/gate_failed. Retained release and upgrade evidence is linked from this OpenCode thread. No unrelated job was cancelled and no resource policy was bypassed. AI assistance: OpenAI gpt-6-astra via OpenCode performed review, merge, release attempt, and recovery diagnosis.

PR #14936, opened 2026-09-22
fix(subtree): separate split timeout from git calls

chubes4:
## Summary

- Keep cheap subtree Git calls bounded at 10 seconds.
- Give `git subtree split` a five-minute default wall-clock budget.
- Add optional `split_timeout_secs` for caller overrides.
- Add real Git fixture coverage for completion beyond 30 seconds and process-group cleanup on timeout.

## Evidence

- Legacy 30-second fixture run: failed with `split subtree timed out after 30s; terminated child process group.` after 31.26s.
- Final fixture run: passed after 33.46s.
- Stuck split fixture: passed after 2.46s and verified the delayed descendant was gone.
- Full serialized subtree test group: 7 passed, 0 failed, 57.29s.
- `cargo check -p homeboy-release`: passed.

The split uses a wall-clock budget because the existing bounded Git runner already guarantees process-group termination; the configurable budget avoids making a fixed default the long-term history-size ceiling. The related release deadline remains out of scope.

AI disclosure: Claude Code driving openai/gpt-5.6-luna, acting for Chris Huber, inspected the issue, implemented the timeout separation, added behavioral tests, and ran verification.


chubes4:
CI differential for 249e6c877 had one candidate-only failure: `split_timeout_terminates_the_delayed_process_group` reported `delayed split descendant survived cleanup`. The timeout itself fired. The assertion checked `kill(pid, 0)` immediately; in CI the orphaned descendant is reaped by an outer Linux child subreaper, so a terminated but unreaped zombie still accepts signal 0.

Fixed in 523f0a5ff: the regression waits up to 5s and treats only a non-zombie process as surviving. This does not weaken the product timeout path.

Verification: `cargo fmt --all -- --check`; `cargo test -p homeboy-core --lib git::subtree::tests -- --test-threads=1` (7 passed, including the >30s split and timeout cleanup regressions).

Finalization provenance: Chris Huber authorized continuing this landing repair directly after Homeboy Cook repeatedly failed before provider execution; the fix was made in the isolated PR worktree outside Cook.

AI disclosure: GPT-6 Astra via OpenCode diagnosed the CI artifact, implemented, tested, and pushed this fix under Chris Huber direction.

PR #14930, opened 2026-09-22
fix(lab): preserve runner provider source delivery

chubes4:
## Summary
- preserve field-level provider source provenance in Lab secret handoff metadata
- admit runner-resolvable provider sources without requiring duplicate runner secret_env refs
- prevent controller-side whole-auth-file provisioning when the sealed plan assigns the source to the runner
- cover the actual run-plan route: `opencode`, `executor.config.provider=codex`, `openai/gpt-5.6-luna`

Fixes #14929 and supports Extra-Chill/homeboy-extensions#2853.

## Verification
- `cargo test -p homeboy-lab-runner lab::secrets::tests -- --nocapture`
- `cargo test -p homeboy-lab-runner execution::tests::secret_source -- --nocapture`
- `cargo test -p homeboy-agents agent_task_provider::secrets::tests -- --nocapture`
- `cargo test -p homeboy-cli commands::agent_task::tests::contract -- --nocapture`

No secret values, auth stores, runtime credentials, or live runner configuration were modified. Tested with GPT-5.6 Luna via OpenCode under Chris Huber authorization.

PR #14971, opened 2026-09-23
fix(lab): provision credential files for the plan-selected provider route

chubes4:
Follow-up to #14930.

## Problem
#14930 uploads declared json-file credential sources to SSH runners before a Lab `agent-task run-plan` hand-off. The upload looked up sources with `provider_secret_sources_for_discovered_providers()`, which is intentionally request-independent and therefore excludes route-scoped sources (`provider_defaults` and `when`-scoped `secret_env_requirements`).

A plan that selects a route such as the OpenCode `openai-oauth` default (Extra-Chill/homeboy-extensions#2866) requires its credential names, but none of them resolved to a file source. The provision list came back empty and the upload returned `Ok` without copying anything. The runner then read its own stale auth store and failed with `authentication_rejected` / `credentials_unusable`.

## Fix
- Add `agent_task_controller_credential_sources(args)`: unconditional declared sources, overlaid with the sources of the route the command's plan selects. It reuses the existing `declared_agent_task_controller_secret_sources_with_providers` path that Lab secret metadata already uses.
- `provision_provider_file_secret_sources_for_runner` resolves sources from the plan-bearing args: `secret_preflight_args` on the Lab offload path and the request command on direct exec.

## Tests
- New `run_plan_credential_sources_include_the_selected_provider_default_file`: unscoped discovery omits the route-scoped default, and a run-plan selecting it yields the json-file source.
- `cargo test -p homeboy-lab-runner --lib -- lab::secrets`: 43 passed.
- `lab::secrets execution::tests lab::offload`: 392 passed. The 7 `execution::tests::daemon_exec` failures are pre-existing and also fail without this change.
- Not yet verified end-to-end: the Lab runner host went offline during refresh (SSH `Host is down`). A live smoke is pending the host returning.

AI assistance: Claude (Anthropic) via OpenCode diagnosed the missing route-scoped source lookup, wrote the change and test, and ran the verification above. Reviewed and directed by @chubes4.


chubes4:
Live verification on the Lab runner (controller and runner both on an integration build containing this change):

- Before: runner OpenCode auth store mtime `2026-07-24`, expired about 50 days earlier. The Lab `agent-task run-plan` smoke failed with `credentials_unusable`.
- After: the same smoke provisioned the store before hand-off. The runner store mtime is now `2026-09-23T05:03:07Z`, with the access token valid for about 210 h.
- The run itself still ended `runtime_unavailable`, for an unrelated reason: OpenAI returns `AI_APICallError: The usage limit has been reached` for this account on both the controller and the runner. The credential is accepted; the account quota is exhausted.

Separate follow-up, not in this PR: when the account is over its usage limit, `opencode run` retries silently until the readiness timeout. The OpenCode readiness probe therefore reports `transient_failure / model_execution_interrupted` instead of a quota classification with `reset_at`.

AI assistance: Claude (Anthropic) via OpenCode ran and summarized this verification.

Issue #14914, opened 2026-09-22
agent-task cook: stack overflow on submission (not --preview) — universal repro, blocks all Cook runs


chubes4:
## Summary

`homeboy agent-task cook` (the actual submission path, not `--preview`) crashes with `fatal runtime error: stack overflow, aborting` on every invocation I've tried. `--preview` with the identical arguments succeeds and returns a valid plan. This blocks all Cook submissions.

## Environment

- `homeboy self identity`: `homeboy 0.383.7+776b7093e3fa7e5cd271580e8bab2d276a5a13cb` (git commit `776b7093e3fa7e5cd271580e8bab2d276a5a13cb`), installed at `/home/opencode/.local/bin/homeboy`
- Reproduced on `0.381.0` (`4da23ce9c5c88a6a2ab614bb37515b95be19c8fc`) first; upgraded via `homeboy upgrade` to `0.383.7`, still crashes identically.
- OS: Linux, running inside a Kimaki/OpenCode coding-agent session on a shared VPS (`extrachill.com`).
- `ulimit -s` default: 8192 (8MB).

## Reproduction

Minimal case (no worktree, no repo-specific state):

```
homeboy agent-task cook \
  --repo agents-api \
  --task-url "https://github.com/Extra-Chill/agents-api/issues/552" \
  --prompt "test minimal prompt" \
  --verify "true" \
  --base main \
  --backend local-shell \
  --detach-after-handoff
```

Output:

```
thread 'main' (1089243) has overflowed its stack
fatal runtime error: stack overflow, aborting
```

Exit code 134 (SIGABRT via Rust's stack-overflow handler).

## What I ruled out

- **Not specific to one repo.** Reproduces identically on `data-machine-business` (with a real existing worktree via `--to-worktree`) and on `agents-api` (fresh, no worktree, letting Cook derive the destination).
- **Not specific to `--to-worktree`.** Crashes both with an explicit `--to-worktree <handle>` and with it omitted (native worktree derivation from `--repo` + `--task-url`).
- **Not specific to backend.** Crashes with `--backend claude-code` and `--backend local-shell` (both reported as ready/admissible in `--preview`).
- **Not specific to `--detach-after-handoff`.** Crashes with and without it.
- **Not specific to prompt content/size.** Crashes with a 9KB `@file` prompt and with a trivial 20-byte inline string.
- **Not specific to `--verify` command content.** Crashes with a real `homeboy review test <component>` gate and with a trivial `true`.
[...]

PR #14976, opened 2026-09-23
fix(agent-task): let a stalled rotation carry its patch forward

### LINKED ISSUE #14965: feat(agent-task): detect silently stalled providers and rotate, keeping worktree state
## Problem
Coding-agent providers can stall silently on usage or rate limits: the session stays "working" but produces no tokens. On 2026-09-23, three consecutive minions for #14914 stalled like this (GLM twice, Kimi once, 20 min with zero output). The operator caught each stall by reading transcripts by hand and moved the work to the next provider himself. Cook already has `--max-attempts` and `--allow-provider-rotation`, but a stalled-yet-alive provider is never classified as failed, so neither kicks in.

## Expected
Cook treats "no provider progress for N minutes" as a classified, retryable provider failure (`provider_stalled`). It:
- records it in the run's evidence;
- if rotation is allowed, moves to the next configured provider, **keeping the worktree's current state** so partial work carries over (the second and third minions tonight built on uncommitted WIP left by the stalled one);
- if not, stops with a clear terminal reason instead of hanging.

## Acceptance
- A configurable stall timeout (default e.g. 10 min without provider output/progress events).
- A test with a fake provider that emits nothing: the run is classified `provider_stalled` and rotates, or terminates.
- Evidence shows which providers were tried and why each ended.

Umbrella: #8010.

chubes4:
Closes #14965.

## The investigation, not just the fix

Before writing any code I grepped for the primitives the issue asked for, per the "check what already exists" rule. Almost everything was already built:

- **Stall classification** already exists end-to-end: `command_runner.rs`'s liveness watchdog kills a silent provider and returns `AgentTaskFailureClassification::Stalled` (`classify_stall_or_rate_limit`).
- **A configurable stall timeout** already exists: `AgentTaskProviderRotationPolicy.liveness_timeout_ms`, defaulting to `DEFAULT_PROVIDER_LIVENESS_TIMEOUT_MS` (5 min, tighter than the issue's illustrative "10 min").
- **Rotation on `Stalled`** is already in `should_rotate_provider`'s eligible-classification list, and `rotates_on_transient_and_timeout_classifications` already asserted it.
- **Per-attempt rotation evidence** (`provider_rotation.attempts[]`, with backend/model/status/classification) already records "which providers were tried and why each ended".
- **"Keeping the worktree's current state" already had a mechanism**: `AgentTaskProviderRotationEntry.adoption` + `select_candidate_adoption` + `validate_and_apply_candidate_adoption` reapplies a verified prior-attempt patch onto the next provider's fresh checkout — this is the *only* way Homeboy ever carries work across a provider swap, and it's explicit-opt-in by design ("the scheduler never adopts a candidate implicitly").

**The actual gap** was one boolean. In `engine.rs`, `candidate_ready_for_convergence` treated a `Stalled` outcome that happened to leave a dirty patch exactly like a genuine `Timeout`: it suppressed the entire rotation branch (including the already-built adoption mechanism) and converged the run straight to `candidate_recoverable` for human review. That's *precisely* the #14914 incident described in the issue — an operator had to notice the stall by hand and move the work to the next provider themselves, because the code path that would have done it automatically was unreachable for `Stalled`.

## The fix

One-line semantic change, `Stalled` excluded from the convergence gate:

```rust
let candidate_ready_for_convergence = outcome.failure_classification
    != Some(AgentTaskFailureClassification::Stalled)
    && outcome.artifacts.iter().any(is_fingerprinted_actionable_patch_artifact);
```

Effects:
- A stall with **no** partial work: unaffected — it already rotated cleanly (nothing to converge on in the first place).
- A stall that **left a dirty patch**, with an operator-configured `rotation.entries[].adoption: adopt_previous_candidate`: now reaches the rotation branch, and the next provider's checkout already carries the prior patch before it runs.
- A stall that left a dirty patch **without** that opt-in: still rotates (not blocked), but the next provider starts clean — "never adopts implicitly" holds exactly as before.
- A stall with no rotation policy at all: unaffected — still converges to `candidate_recoverable`, a clear terminal state instead of hanging.

`Timeout` and every other classification are untouched; the change is scoped to `Stalled` only.

## Why not rename `Stalled` → `provider_stalled`, or lower the default to 10 min

The issue's acceptance text says "classified `provider_stalled`" and "default e.g. 10 min", but both already exist under different, already-shipped names/values (`stalled`, 5 min). Renaming the classification would be a breaking change to already-persisted durable run records and every existing caller (dozens of sites); lowering the default would be an unrelated global behavior change affecting every Cook run, not just the stall-rotation gap this issue is about. Treating both as illustrative rather than literal and leaving them alone.

## Tests

New file `crates/homeboy-agents/src/agent_task_scheduler/tests/scheduling_tests/stall_rotation.rs`, five tests against the real scheduler loop (`AgentTaskScheduler::run`, the same path production Cook runs through) with fake providers that stall exactly as described in the issue:

- `a_provider_that_emits_nothing_is_classified_stalled_and_rotates_to_the_next_provider` — the #14914 "20 min zero output" shape; asserts classification + rotation evidence.
- `a_stall_with_an_adopting_rotation_entry_carries_the_uncommitted_edit_into_the_next_provider` — proves the fix: the second provider's workspace already contains the first provider's uncommitted edit before it runs.
- `a_stall_without_an_adopting_rotation_entry_still_rotates_but_does_not_carry_the_edit` — proves the "never adopts implicitly" invariant still holds by default.
- `a_stall_with_no_rotation_policy_terminates_as_a_recoverable_candidate` — proves the "no hang" terminal path.
- `a_stall_with_no_partial_work_and_no_rotation_policy_terminates_cleanly` — proves a stall never silently looks like success.

## Verification

```
cargo fmt --check                     # clean
cargo build --workspace               # clean
cargo test -p homeboy-agents --lib agent_task_scheduler::   # 182 passed, 1 pre-existing failure (below)
```

Targeted, single-threaded rerun of every test directly touching stall/rotation/timeout behavior (63 tests: my 5 new + all of `provider_rotation`, `timeout`, `retry_failure`, plus the two most relevant pre-existing scheduler-level stall tests) — **all 63 pass in isolation**.

### Pre-existing failures (not caused by this change)

A full `cargo test -p homeboy-agents --lib` run under this VPS's current load is flaky: it reported 48 failures with this change applied and 54 on a clean `stash`+rerun of unmodified `origin/main` moments later — a different failure *count* and largely different failure *set* from the identical clean code, which is only possible from environmental flakiness (heavy concurrent load from other sessions, near-full disk, real-subprocess timing sensitivity), not a code regression. Confirmed two ways:
- `agent_task_scheduler::tests::scheduling_tests::concurrency::concurrency_tests::retry_uses_clean_isolated_workspace_after_permission_denial` fails identically with and without my change (a `chmod 000` permission-denial test that can't work while running as root).
- `agent_task_provider::tests::scheduler_tests::stalled_provider_is_killed_and_rotates_to_configured_fallback` (real `node` subprocess + 5s wall-clock liveness timeout) failed under full-suite parallel load but passed cleanly in the isolated single-threaded rerun above, alongside every other test this change touches.

None of the ~50 flaky names in either run intersect with logic this PR modifies (`candidate_ready_for_convergence`/rotation-vs-convergence), and the full isolated pass above covers that logic directly.

## AI-assisted disclosure

Implemented by a `kimaki send` minion running outside Homeboy Cook, dispatched by the orchestrator session on extrachill.com.


chubes4:
Rebased onto current `origin/main` (`841a110e01`) — clean, no conflicts. Confirms the differential-gate failure was base skew, not caused by this change.

Re-verified on the rebased branch:
```
cargo fmt --check        # clean
cargo build --workspace  # clean
```

Targeted tests (single-threaded):
- `stall_rotation::` (my 5 new tests) — all pass.

Requested filters, with a same-command comparison against clean `origin/main` (checked out the pre-rebase parent commit in place, no new worktree):

| Filter | On this branch | On clean `origin/main` |
|---|---|---|
| `cargo test -p homeboy-cli agent_task` | 737 passed, **17** failed | 737 passed, **17** failed — identical failure list |
| `cargo test -p homeboy-agents agent_task_lifecycle` | 437 passed, **4** failed | 437 passed, **4** failed — identical failure list |

Both failure sets are byte-identical between this branch and unmodified `origin/main`: same test names, same panic messages, same counts. None of the 21 total touch `agent_task_scheduler` rotation/convergence logic — they're spread across CLI help-text auditing, promotion/resume state assertions, durable-progress cursors, and worktree-identity checks, none of which this PR modifies. They also match the categories the differential gate reported as pre-existing (`cli_surface`, `help_audit`, `loop_job`, `owned_args_guard`, `dispatchability` all live in modules this diff never touches). This is base-environment flakiness (see the original PR body for the earlier full-suite flakiness evidence), not a regression from this change.

Force-pushed the rebase with `--force-with-lease`.

chubes4:
Merged over the red differential Test gate with operator approval. None of the tests it blames are covered by this change. Run on their own, each either passes or also fails on `main`, so they're flaky. Tracked in #14984.

PR #14978, opened 2026-09-23
fix(cli): keep --detach-after-handoff as a deprecated no-op

### LINKED ISSUE #14964: fix(cli): keep removed agent-task flags as deprecated no-ops (--detach-after-handoff)
## Problem
`homeboy agent-task cook --detach-after-handoff` was a documented flag in 0.383.7. By 0.383.15 it's gone, and using it fails with clap's `unexpected argument '--detach-after-handoff'`. Detaching became the default, which is good, but existing scripts, prompts, runbooks and issue repro commands (e.g. #14914's repro) now break outright. None of the releases in between mention the removal.

## Expected
Removed or renamed CLI flags stay accepted, as a hidden no-op or alias, for at least one minor release, and print a structured deprecation warning naming the replacement (here: "detached is now the default; use global `--wait` to observe to completion"). There's also a guard (test or lint) so a public flag can't be removed without its deprecation shim.

## Acceptance
- `--detach-after-handoff` is accepted as a hidden no-op with a deprecation warning in the command-result diagnostics.
- A contract test lists retired flags and asserts they still parse.

chubes4:
Closes #14964.

## Problem

`--detach-after-handoff` was a documented `agent-task cook` flag through
0.383.7. #14931 (`fix(cli): return after durable handoff unless --wait`)
flipped the default to detach-by-default and renamed the flag to `--wait`
(inverted polarity), deleting the old spelling outright. No release note
mentioned the removal, so every script, prompt, and runbook written against
it — including #14914's own repro command — now hard-fails with clap's
`unexpected argument '--detach-after-handoff'`.

## Design

`--detach-after-handoff` is restored as a **hidden global no-op**:

- `Cli` gains `deprecated_detach_after_handoff`, a `hide = true`, `global =
  true` boolean carrying the retired long spelling. It never changes
  behavior — detaching is already the default — it only keeps the flag
  parseable.
- `warn_if_deprecated_flags_used` (in `cli_runtime.rs`) checks it on every
  command path immediately after argument parsing and, if set, prints:
  `warning: --detach-after-handoff is deprecated and is now a no-op;
  detached is now the default. Use global `--wait` to observe to
  completion.` This fires regardless of which subcommand is invoked, not
  just `agent-task cook`, since the flag was global before removal too.
- `--wait` still wins if both are passed together (tested).
- `root_global_flag_surface_is_pinned` is updated to include the restored
  spelling — that test enumerates every global flag, hidden or not, and is
  the guard that turns "delete a global flag" into a visible diff instead of
  a silent one.

### Contract test / guard (Acceptance criterion #2)

Added `crates/homeboy-cli/src/cli_surface/retired_flags.rs`: a small
registry of retired public flags (long spelling + clap `id` + sample argv
that reaches a real command), with a test
(`retired_flags_still_parse_as_hidden_no_ops`) that parses each entry's
sample argv and asserts it still succeeds. A second test
(`retired_flags_stay_hidden_from_the_built_command_tree`) asserts every
registered entry is declared `hide = true`. Future flag retirements add an
entry here instead of deleting the flag outright; nothing enforces that at
compile time (that would need a diff-based CI check, out of scope), but the
registry plus the pinned global-flag-surface test together make skipping
this step a reviewable diff.

### Investigated and out of scope

Per the task prompt, I checked `v0.383.7..origin/main` for other public
flags removed in the same range:

- `git log -S'detach-after-handoff' v0.383.7..origin/main` → one commit,
  eb76c60e7 (#14931).
- No other `#[arg(long = ...)]` removals under `commands/agent_task/args/`,
  `controller.rs`, `fanout.rs`, `review.rs`, or `status.rs` in that range —
  only additions (e.g. `fanout.rs`'s new `--idempotency-key`).
- `fuzz stable plan` also had its **own local** `--detach-after-handoff`
  flag (not the global one), renamed to `--wait` in the *same* commit
  (#14931). It's a plan-only flag that decides whether a *generated*
  `agent-task cook` command string includes `--wait` — not itself a live
  execution-policy flag, and not an `agent-task` flag. Left unshimmed here
  since the task scoped this to agent-task flags; flagging in case
  `fuzz stable plan` scripts turn out to need the same treatment.

## Tests

- `cli_surface::retired_flags::tests::retired_flags_still_parse_as_hidden_no_ops`
  (new)
- `cli_surface::retired_flags::tests::retired_flags_stay_hidden_from_the_built_command_tree`
  (new)
- `cli_surface::tests::wait_is_global_and_independent_of_placement` (extended:
  retired flag still detaches by default; `--wait` still wins when both are
  passed)
- `cli_surface::global_flag_surface_tests::root_global_flag_surface_is_pinned`
  (updated pinned list)
- `cli_binary::cook_preview_lifecycle::deprecated_detach_after_handoff_flag_still_parses_as_a_warned_no_op`
  (new, end-to-end: real binary invocation with `--detach-after-handoff`
  succeeds, prints the warning naming `--wait`, and the preview JSON is
  unaffected)

## Verification

```
cargo fmt --check                                    # clean
cargo build                                           # succeeds
cargo test -p homeboy-cli --lib cli_surface::         # 35 passed, 1 pre-existing failure (see below)
cargo test -p homeboy-cli --lib fuzz::                # 164 passed
cargo test -p homeboy-cli --lib owned_args_guard      # 6 passed, 1 pre-existing failure (see below)
cargo test --test cli_binary cook_preview_lifecycle:: # 4 passed
cargo test --test cook_lab_handoff_test               # 13 passed, 4 pre-existing failures (see below)
```

### Pre-existing failures (confirmed via `git stash` + rerun on exact
`origin/main`, unrelated to this change)

- `cli_surface::tests::command_registry_docs_paths_exist_and_are_indexed` —
  fails identically on `origin/main`: `registered command 'control-plane'
  points at missing docs path docs/commands/control-plane.md`.
- `command_capability::owned_args_guard_test::a_new_raw_argv_flag_scan_is_a_failure`
  — fails identically on `origin/main`: flags two unacknowledged raw-argv
  scans in `commands/agent_task/run.rs` (`local_preview_replay`,
  `preview_local_placement_admission`) that landed in a recent upstream
  commit, unrelated to this PR (I did not touch `run.rs`).
- `cook_lab_handoff_test::{attached_nonlocal_cook_observes_terminal_execution,
  explicit_wait_and_runner_owned_cooks_return_terminal_failure,
  foreground_local_cook_survives_client_termination_with_artifacts,
  no_runner_lab_or_local_cook_reaches_local_execution}` — all four fail
  identically on `origin/main` with `hermetic subprocess did not finish
  within its 180s budget`, `Load average is 9.0-13.2 across 8 CPU(s)`, and
  `blocked_runner_unavailable`: host resource pressure from concurrent
  builds on this shared VPS, not a regression from this change.

## AI-assisted disclosure

This PR was authored by a kimaki minion session running outside Homeboy
Cook, dispatched by Extra Chill's orchestrator agent.


chubes4:
Merged over the red differential Test gate with operator approval. None of the tests it blames are covered by this change. Run on their own, each either passes or also fails on `main`, so they're flaky. Tracked in #14984.

